PR #2, opened 2025-02-11
✨(back) create wopi application

lunika:
## Purpose

We want to implement the wopi protocol. We create a django app dedicated to manage all the wopi protocol implementation. The first thing made is to set wopi client configuration.

Documentation : 

- [Key concepts](https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/concepts)
- [Passing access token securely](https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/hostpage#passing-access-tokens-securely)
- [WOPI validator](https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/build-test-ship/validator)

## Proposal

Missing:

- [x] configure using discovery mechanism
- [x] front

## Summary by CodeRabbit

- **New Features**
  - Online document viewing/editing via WOPI (Collabora/OnlyOffice) in file preview.
  - Automatic file locking, unlocking, relock and rename during edits.

- **API**
  - New endpoint to get WOPI launch info for a file.
  - Items now include is_wopi_supported flag.

- **UX/Frontend**
  - WOPI editor iframe with loading/error states and new error preview + translations.

- **Chores**
  - Dev setup adds Collabora/OnlyOffice services, periodic WOPI configuration and CI triggers for wopi branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(back) create wopi applcation #2
+
 ### Changed
 
 - ♻️(backend) use PUT presigned-url to upload files #345
@@ -61,9 +65,6 @@ and this project adheres to
 - 🐛(front) fix workspace link react cache #310
 - 🐛(backend) allow item partial update without modifying title #316
 - 🌐(front) fix share item wording #315
-  <<<<<<< HEAD
-- # 🐛(front) fix left tree glitch #323
-  > > > > > > > c624750 (📝(changelog) update changelog)
 
 ## [v0.2.0] - 2025-08-18
 
--- a/src/frontend/apps/drive/src/features/explorer/utils/utils.ts
+++ b/src/frontend/apps/drive/src/features/explorer/utils/utils.ts
@@ -147,5 +147,6 @@ export const itemToPreviewFile = (item: Item) => {
     mimetype: item.mimetype ?? "",
     url: item.url ?? "",
     isSuspicious: item.upload_state === ItemUploadState.SUSPICIOUS,
+    is_wopi_supported: item.is_wopi_supported,
   };
 };


src/backend/drive/settings.py
```diff
@@ -591,6 +537,36 @@ def post_setup(cls):
...
+        if cls.WOPI_CLIENTS:
+            listValue = values.ListValue()
+            wopi_client_configuration = {}
+            wopi_client_errors = []
+            for wopi_client in cls.WOPI_CLIENTS:
+                wopi_client_upper = wopi_client.upper()
+                wopi_client_mimetypes = os.getenv(f"WOPI_{wopi_client_upper}_MIMETYPES")
+                wopi_client_launch_url = os.getenv(
+                    f"WOPI_{wopi_client_upper}_LAUNCH_URL"
+                )
```
qbey: Why don't you use the `value.*` from django-configurations here?

src/backend/drive/settings.py
```diff
@@ -534,6 +462,24 @@ class Base(Configuration):
...
+    # WOPI
+
+    """
+    WOOPIE_CLIENTS contains a list of client name. These client names will be used in the post_setup
+    to configure the settings related to the client.
+    Example : 
+    WOOPIE_CLIENTS = ["vendorA"]
+    Then these settings will be cheked in the post_setup:
+    WOOPIE_VENDORA_MIMETYPES=[]
+    WOOPIE_VENDORA_LAUNCH_URL=""
```
qbey: Woopsy? Whoopi Goldberg?

Also, please don't use `"""` for comments, prefix every line with `#` instead :)

src/backend/drive/settings.py
lunika: Because I just though that in the `post_setup` function it was not possible to use it. But if I can it's a good news.

.github/workflows/docker-hub.yml
```diff
@@ -6,11 +6,13 @@ on:
   push:
```
NathanVss, line 6: BUG FOUND

Uploading a plain text file, first the preview tells it is not supported, switch tabs, go back, it opens collabora. It should do it the first time.




https://github.com/user-attachments/assets/e441bf85-1c6e-44a7-8fc0-423fcdbdc6f8

NathanVss, line 6: When opening a word file I get this blank kind of screen: 
Capture d’écran 2025-09-04 à 12 21 13

It's the "doc" file here:
https://fichiers-staging.beta.numerique.gouv.fr/explorer/items/2c6693e5-9296-4dc1-b0e9-f4086ef3e5a0

NathanVss: I randomly face this error on some files, last time I got it on "doc" file here: 

https://fichiers-staging.beta.numerique.gouv.fr/explorer/items/2c6693e5-9296-4dc1-b0e9-f4086ef3e5a0

Capture d’écran 2025-09-04 à 12 23 38

NathanVss, line 6: The double closes button is misleading in UX I think, we need only one. I guess we should disable the arrow of Collabora in favor of the file previewer one?

If we click on collabora arrow we get an empty canvas


https://github.com/user-attachments/assets/ceb1810b-f4fe-42fb-a483-0a94576dbb2c

src/backend/wopi/tests/test_settings.py
```diff
@@ -0,0 +1,101 @@
...
+            "mimetypes": [
+                "application/vnd.oasis.opendocument.text",
+                "application/vnd.oasis.opendocument.spreadsheet",
+            ],
+        },
+        "vendorB": {
+            "launch_url": "https://vendorB.com/launch_url",
+            "mimetypes": ["application/vnd.oasis.opendocument.presentation"],
+        },
+    }
```
NathanVss, line 37: Why not only use `WOPI_CLIENTS_CONFIGURATION` and let the users do this configuration by its own instead of using dynamic WOPI_<vendor>_ATTRIBUTE? 

No big deal anyway

src/backend/core/api/serializers.py
```diff
@@ -132,6 +132,7 @@ class ListItemSerializer(serializers.ModelSerializer):
     url = serializers.SerializerMethodField()
     creator = UserLiteSerializer(read_only=True)
     hard_delete_at = serializers.SerializerMethodField(read_only=True)
+    is_wopi_supported = serializers.SerializerMethodField()
```
NathanVss, line 138: typo in the commit `propserty` :p

src/backend/wopi/utils/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""Utils for WOPI"""
+
+from django.conf import settings
+
+from core import models
+
+
+def is_item_wopi_supported(item):
```
NathanVss: so, maybe the test should be in this app instead? Up to you.

src/backend/core/api/serializers.py
```diff
@@ -11,6 +11,7 @@
 
 from core import models
 from core.api import utils
+from wopi import utils as wopi_utils
 
 
 class UserSerializer(serializers.ModelSerializer):
```
NathanVss, line 18: nitpick: typo in commit body ( "beem" )

src/backend/wopi/services/access.py
```diff
@@ -0,0 +1,98 @@
...
+        except (KeyError, ValueError) as error:
+            raise AccessUserItemInvalidDataError("Invalid data") from error
+
+
+class AccessUserItemService:
+    """Service managing the access token for WOPI."""
+
+    @staticmethod
+    def generate_token():
+        """Generate a rancom access token"""
```
NathanVss: typo

src/backend/wopi/services/access.py
```diff
@@ -0,0 +1,98 @@
...
+        return token_urlsafe()
+
+    def insert_new_access(self, item: Item, user: AbstractUser) -> str:
```
NathanVss: nit: `create_new_access` name is maybe more straightforward as naming convention

src/backend/wopi/tests/services/test_access.py
```diff
@@ -0,0 +1,218 @@
...
+    access_user_item_service = AccessUserItemService()
+    access_token, _ = access_user_item_service.insert_new_access(
+        access.item, access.user
+    )
+
+    access_user_item = access_user_item_service.get_access_user_item(access_token)
+
+    assert access_user_item.item == access.item
+    assert access_user_item.user == access.user
+
```
NathanVss, line 209: Maybe add a test case ensuring that subsequent tokens are different, to make sure they are unique?

src/backend/core/api/viewsets.py
```diff
@@ -34,6 +34,8 @@
 from core import enums, models
 from core.services.sdk_relay import SDKRelayManager
 from core.tasks.item import process_item_deletion
+from wopi.services import access as access_service
+from wopi.utils import get_wopi_client_config
```
NathanVss: nit: commit body typo "returing"

src/backend/core/tests/items/test_api_items_wopi.py
```diff
@@ -11,26 +10,27 @@
 pytestmark = pytest.mark.django_db
 
 
-@pytest.fixture
-def timestamp_now():
+@pytest.fixture(name="timestamp_now")
```
NathanVss, line 17: Curious: why adding name and prefixing with `fixture` those?

src/backend/wopi/tests/viewset/test_check_file_info.py
```diff
@@ -0,0 +1,146 @@
...
+        update_upload_state=models.ItemUploadStateChoices.UPLOADED,
+        link_reach=models.LinkReachChoices.RESTRICTED,
+        link_role=models.LinkRoleChoices.EDITOR,
+    )
+    user = factories.UserFactory()
+    factories.UserItemAccessFactory(
+        item=item, user=user, role=models.RoleChoices.EDITOR
+    )
+
+    upload_file = default_storage.connection.meta.client.put_object(
```
NathanVss, line 35: You could use `default_storage.save`?

src/backend/wopi/tests/viewset/test_check_file_info.py
```diff
@@ -0,0 +1,146 @@
...
+        parent=folder,
+        type=models.ItemTypeChoices.FILE,
+        filename="wopi_test.txt",
+        update_upload_state=models.ItemUploadStateChoices.UPLOADED,
+        link_reach=models.LinkReachChoices.PUBLIC,
+        link_role=models.LinkRoleChoices.READER,
+    )
+    user = AnonymousUser()
+
+    upload_file = default_storage.connection.meta.client.put_object(
```
NathanVss, line 193: Ditto

src/backend/wopi/tests/viewset/test_check_file_info.py
```diff
@@ -0,0 +1,146 @@
...
+import pytest
+from rest_framework.test import APIClient
+
+from core import factories, models
+from wopi.services.access import AccessUserItemService
+
+pytestmark = pytest.mark.django_db
+
+
+def test_check_file_info_connected_user_with_access():
```
NathanVss, line 17: Could be nice to add a test that asserts the file is readonly is it is reader ( not only anonymous )

NathanVss: Also for delete items

src/backend/wopi/viewsets.py
```diff
@@ -67,3 +71,35 @@ def retrieve(self, request, pk=None):
             "DownloadUrl": f"/media/{item.file_key}",
         }
         return Response(properties, status=200)
+
+    @action(detail=True, methods=["get"], url_path="contents")
+    def get_file_content(self, request, pk=None):
+        """
+        Implementation of the Wopi GetFile file operation
+        https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/getfile
```
NathanVss, line 113: Wrong URL: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/getfile

src/backend/drive/settings.py
```diff
@@ -626,6 +626,9 @@ class Base(Configuration):
     WOPI_ACCESS_TOKEN_TIMEOUT = values.IntegerValue(
         60 * 60 * 10, environ_name="WOPI_ACCESS_TOKEN_TIMEOUT", environ_prefix=None
     )
+    WOPI_LOCK_TIMEOUT = values.IntegerValue(
+        30 * 60, environ_name="WOPI_LOCK_TIMEOUT", environ_prefix=None
+    )
```
NathanVss, line 685: nit: commit body message "complete".

src/backend/wopi/routers.py
```diff
@@ -0,0 +1,48 @@
+"""Routers of the WOPI app."""
+
+from rest_framework.routers import DefaultRouter, DynamicRoute, Route
+
+
+class WopiRouter(DefaultRouter):
```
NathanVss, line 6: As It seems that we will not have to bother about implementing another viewset that uses this router in a generic way, why did you create a custom router?

src/backend/drive/settings.py
```diff
@@ -154,10 +154,12 @@ class Base(Configuration):
...
     )
 
-    # item images
-    ITEM_FILE_MAX_SIZE = values.PositiveIntegerValue(
-        5 * (2**30),  # 5GB
-        environ_name="ITEM_FILE_MAX_SIZE",
+    # Maximum size of the request body in memory.
+    # This is used to limit the size of the request body in memory.
+    # This also limits the size of the file that can be uploaded to the server.
+    DATA_UPLOAD_MAX_MEMORY_SIZE = values.PositiveIntegerValue(
```
NathanVss, line 170: Why did you changed the name?

src/backend/wopi/viewsets.py
```diff
@@ -117,6 +134,47 @@ def get_file_content(self, request, pk=None):
...
+
+        if not abilities["update"]:
+            return Response(status=401)
+
+        lock_value = request.META.get(HTTP_X_WOPI_LOCK)
+
+        if lock_value:
+            lock_service = LockService(item)
+            current_lock_value = lock_service.get_lock(default="")
+            if current_lock_value != lock_value:
```
NathanVss, line 161: You could use `is_lock_valid`?

src/backend/wopi/viewsets.py
```diff
@@ -117,6 +134,47 @@ def get_file_content(self, request, pk=None):
...
+            if body_size > 0:
+                return Response(status=409, headers={X_WOPI_LOCK: ""})
+
+        try:
+            file = ContentFile(request.body)
+        except RequestDataTooBig:
+            return Response(status=413)
+
+        s3_client = default_storage.connection.meta.client
+        file = s3_client.put_object(
```
NathanVss: It does not seem that here we update the `size` attribute of the file in DB?

src/backend/wopi/tests/viewset/test_put_file_content.py
```diff
@@ -0,0 +1,260 @@
...
+        headers={
+            "X-WOPI-Override": "PUT",
+        },
+    )
+    assert response.status_code == 409
+    assert response.headers.get("X-WOPI-Lock") == ""
+
+
+def test_put_file_content_with_no_lock_header_and_body_size_0():
+    """User can put file content when not providing a lock header and the body size is 0."""
```
NathanVss, line 376: Why? 🤔

NathanVss: It means that we reset erase a file without having a lock?

src/backend/wopi/tests/viewset/test_put_file_content.py
```diff
@@ -0,0 +1,260 @@
...
+
+import pytest
+from rest_framework.test import APIClient
+
+from core import factories, models
+from wopi.services.access import AccessUserItemService
+from wopi.services.lock import LockService
+
+pytestmark = pytest.mark.django_db
+
```
NathanVss, line 13: Edge case: If the user loose its ability to write to a file because it has been removed from a workspace and was editing a file, it should not be able to save

src/backend/wopi/tests/viewset/test_put_file_content.py
```diff
@@ -0,0 +1,260 @@
...
+
```
NathanVss, line 14: Another: what if the user edits a file that has been moved to trash or hard deleted since?

src/backend/wopi/viewsets.py
```diff
@@ -310,4 +314,77 @@ def _rename_file(self, request, pk=None):
...
+            Item.objects.filter(path__descendants=".".join(parent_path))
+            .filter(path__depth=item.depth)
+            .filter(filename=new_filename_with_extension)
+            .exclude(id=item.id)
+        )
+
+        if queryset.exists():
+            return Response(
+                status=400,
+                headers={X_WOPI_INVALIDFILENAMERROR: "Filename already exists"},
```
NathanVss, line 356: Should we send error with the current language of the user? I'd bet this error will be display inside collabora to the user

src/backend/drive/settings.py
```diff
@@ -619,11 +619,44 @@ class Base(Configuration):
     # Then these settings will be cheked in the post_setup:
     # WOPI_VENDORA_MIMETYPES="application/vnd.oasis.opendocument.text"
     # WOPI_VENDORA_LAUNCH_URL="https://vendorA.com/launch_url"
+    # WOPI_VENDORA_DISCOVERY_URL = https://vendorA.com/hosting/discovery
     # If they are missing, a ValueError will be raised.
     WOPI_CLIENTS = values.ListValue(
         [], environ_name="WOPI_CLIENTS", environ_prefix=None
     )
     WOPI_CLIENTS_CONFIGURATION = {}
+    WOPI_EXCLUDED_MIMETYPES = values.ListValue(
```
NathanVss, line 638: Why not use the reverse approach? Defining `WOPI_MIMETYPES` only? So that way we have a fine control over what's will be editable via WOPI. It also ensures that we prevent any potential security issues that would be based on mimetype from discovery urls

src/backend/wopi/tasks/configure_wopi.py
```diff
@@ -0,0 +1,95 @@
...
+                wopi_configuration["mimetypes"][mimetype] = action.get("urlsrc")
+
+            else:
+                extension = action.get("ext")
+
+                if extension in settings.WOPI_EXCLUDED_EXTENSIONS:
+                    continue
+
+                wopi_configuration["extensions"][extension] = action.get("urlsrc")
+
```
NathanVss, line 93: It seems that with this approach we are not able to easily use multiple wopi vendors and, for instance use vendorA for .EXT1 and vendorB for .EXT2 if both handles both file types. 

It could happen in a scenarios were even if both vendors support editing of those two extension we rather prefer for UX reasons send the user to a specific vendor.

NathanVss: With the current approach the second WOPI_CLIENTS will override the mimetype supported by the previous one. 

We should maybe rely on a setting that is a map 

```
[mime]: "vendorA"
```

And then pick the matching URL inside this discovery method.

WYDT?

src/helm/drive/templates/backend_celery_beat.yaml
```diff
@@ -0,0 +1,136 @@
+{{- $envVars := include "drive.common.env" (list . .Values.backend) -}}
```
NathanVss, line 1: nit: commit body typo "scheduled"

env.d/development/common
```diff
@@ -51,4 +51,9 @@ FRONTEND_THEME=default
 FRONTEND_MORE_LINK=https://docs.numerique.gouv.fr/docs/fa0aba15-e119-4185-b466-a4b37ad95950/
 FRONTEND_FEEDBACK_BUTTON_SHOW=True
 FRONTEND_FEEDBACK_BUTTON_IDLE=False
-FRONTEND_FEEDBACK_ITEMS={"form":{"url": "https://tally.so/r/w2YaWL"}, "tchap":{"url": "https://tchap.gouv.fr/#/room/#fichiers:agent.dinum.tchap.gouv.fr"}, "visio":{"url": "https://cal.com/robin-lecomte-eyul9w/fichiers"}}
\ No newline at end of file
+FRONTEND_FEEDBACK_ITEMS={"form":{"url": "https://tally.so/r/w2YaWL"}, "tchap":{"url": "https://tchap.gouv.fr/#/room/#fichiers:agent.dinum.tchap.gouv.fr"}, "visio":{"url": "https://cal.com/robin-lecomte-eyul9w/fichiers"}}
+
+# WOPI
+WOPI_CLIENTS="collabora"
```
NathanVss: Should be a `ListValue` so `["collabora]`?

src/frontend/apps/drive/src/features/ui/preview/wopi/WopiEditor.scss
```diff
@@ -0,0 +1,5 @@
+.wopi-editor-iframe {
+  width: 100%;
+  height: calc(100vh - 52px);
```
NathanVss, line 3: Could be cool to defined the `52px` as a var the matches the height of the header

NathanVss: ( and use it for the header too, obviously )

src/frontend/apps/drive/src/features/ui/preview/wopi/WopiEditor.tsx
```diff
@@ -0,0 +1,78 @@
...
+    queryFn: () => getDriver().getWopiInfo(item.id),
+  });
+
+  const wopiUrl = useMemo(() => {
+    if (!wopiInfo) {
+      return undefined;
+    }
+
+    const baseUrl = wopiInfo.launch_url;
+    const separator = baseUrl.includes("?") ? "&" : "?";
```
NathanVss: Maybe make use `new URL(launch_url)`  as customize the `searchParams` instead? 

https://developer.mozilla.org/fr/docs/Web/API/URL/URL

src/backend/wopi/tests/test_utils.py
```diff
@@ -92,59 +277,243 @@ def test_get_wopi_client_config():
...
+        value
+        for value in models.ItemUploadStateChoices.values
+        if value
+        not in [
+            models.ItemUploadStateChoices.READY,
+            models.ItemUploadStateChoices.SUSPICIOUS,
+        ]
+    ],
+)
+def test_get_wopi_client_config_not_ready_but_creator(upload_state):
```
NathanVss, line 337: maybe I missed it but I don't find a test that ensure an owner can edit its analyzing file?

lunika: Because an owner can't. We differentiate creator and owner. So a creator can. An owner can't.

env.d/development/common
lunika: A list value is represented with a comma separated string when used with an environment variable.

src/backend/wopi/tests/test_settings.py
lunika: I change how it is configured by using the discovery process

src/backend/wopi/utils/__init__.py
lunika: Not sure to understand why. Everything related to the wopi feature is in the wopi app.

src/backend/core/tests/items/test_api_items_wopi.py
lunika: Because pylint is raising an error if a function name and a function arg as the same name.

src/backend/wopi/tests/viewset/test_check_file_info.py
lunika: I'm not using it because I need to know the version_id of the created file in the bucket.

src/backend/wopi/viewsets.py
lunika: Hmmm not sure to understand it is wrong. What is the good one ?

src/backend/wopi/routers.py
lunika: Just to manage the trailing slash an other way django is managing it.

src/backend/drive/settings.py
lunika: I think it is explain in the comment.
This setting is internal to django and I reuse it.

src/backend/wopi/tests/viewset/test_put_file_content.py
lunika: Added

src/backend/wopi/tests/viewset/test_put_file_content.py
lunika: Added

src/backend/wopi/tests/viewset/test_put_file_content.py
lunika: > When a host receives a PutFile request on a file that's not locked, the host checks the current size of the file. If it's 0 bytes, the PutFile request should be considered valid and should proceed. If it's any value other than 0 bytes, or missing altogether, the host should respond with a [409 Conflict](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.10). For more information, see [Creating new files using Microsoft 365 for the web](https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/scenarios/createnew).

Because... :man_shrugging:

src/backend/wopi/viewsets.py
lunika: I need the `current_lock_value` and don't want to query twice redis to check first if it's valid and then get the value if not.

src/backend/wopi/viewsets.py
lunika: Good catch

src/backend/wopi/viewsets.py
lunika: For now we don't manage translation in the backend application. I think it's out of the scope of this PR

src/backend/drive/settings.py
lunika: Because I think you have no idea of the list of supported mimetype and how many mimtype different exists for the same file type ^^

src/backend/wopi/tasks/configure_wopi.py
lunika: I made it this way because if both can manage the same, you should configure your tools to decide which one to use.
But, maybe, in a near future we would like a different approach.
For now I would like to keep this one

src/backend/wopi/routers.py
NathanVss: What was causing issue with the way django is managing it?

src/backend/drive/settings.py
NathanVss: I know how much mimetypes could be supported :p But let's imagine a scenario where the WOPI editor gets compromised when editing a specific mimetype added during a nasty version, Drive would automatically allow the edition of this file and then the exploit of the security breach. If we are ok with this risk, which is edgy I admit, let's go.

src/backend/wopi/tasks/configure_wopi.py
NathanVss: Let's go ;)

PR #22, opened 2025-03-12
Create default workspace for newly created user

### LINKED ISSUE #13: Create a default Personal workspace for each user
## Feature Request

Create a default "Personal" workspace for each new user, it must be empty. 

It will be used as a base for the frontend to allow the user to create folders in it, upload its files, etc.

⚠ It must be impossible to delete.

It must be tagged so we can detect if it is a personal folder from the source code, like "personal: true" or some sort of things, up to you.

Image

lunika:
## Purpose

When a new user is created, we want to create a default workspace for him.
This default workspace must not be removed, a new property `main_workspace` has been added on the Item model, if set to true, it is not possible to (soft) delete it.

## Proposal

- [x] add main_workspace property on Item model
- [x] create a default workspace on user creation

Fixes #13

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -22,7 +22,7 @@ from django.db import models, transaction
 from django.db.models.expressions import RawSQL
 from django.template.loader import render_to_string
 from django.utils import timezone
-from django.utils.functional import cached_property, lazy
+from django.utils.functional import cached_property
 from django.utils.translation import get_language, override
 from django.utils.translation import gettext_lazy as _
 
@@ -207,7 +207,7 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
 
     language = models.CharField(
         max_length=10,
-        choices=lazy(lambda: settings.LANGUAGES, tuple)(),
+        choices=settings.LANGUAGES,
         default=settings.LANGUAGE_CODE,
         verbose_name=_("language"),
         help_text=_("The language in which the user wants to see the interface."),
--- a/src/backend/core/migrations/0001_initial.py
+++ b/src/backend/core/migrations/0001_initial.py
@@ -35,7 +35,7 @@ class Migration(migrations.Migration):
                 ('short_name', models.CharField(blank=True, max_length=20, null=True, verbose_name='short name')),
                 ('email', models.EmailField(blank=True, max_length=254, null=True, verbose_name='identity email address')),
                 ('admin_email', models.EmailField(blank=True, max_length=254, null=True, unique=True, verbose_name='admin email address')),
-                ('language', models.CharField(choices="(('en-us', 'English'), ('fr-fr', 'French'), ('de-de', 'German'))", default='en-us', help_text='The language in which the user wants to see the interface.', max_length=10, verbose_name='language')),
+                ('language', models.CharField(choices=settings.LANGUAGES, default='en-us', help_text='The language in which the user wants to see the interface.', max_length=10, verbose_name='language')),
                 ('timezone', timezone_field.fields.TimeZoneField(choices_display='WITH_GMT_OFFSET', default='UTC', help_text='The timezone in which the user wants to see times.', use_pytz=False)),
                 ('is_device', models.BooleanField(default=False, help_text='Whether the user is a device or a real user.', verbose_name='device')),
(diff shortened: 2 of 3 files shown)


src/backend/core/migrations/0002_item_main_workspace_alter_user_language.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.AddField(
+            model_name='item',
+            name='main_workspace',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AlterField(
+            model_name='user',
+            name='language',
+            field=models.CharField(choices="(('en-us', 'English'), ('fr-fr', 'French'), ('de-de', 'German'))", default='en-us', help_text='The language in which the user wants to see the interface.', max_length=10, verbose_name='language'),
+        ),
```
qbey: I've removed the lazy in `people` project to get rid of these noop migrations : https://github.com/suitenumerique/people/commit/469014ac415b25be0ceed08b31a87d2d40d743cd

The `LANGUAGES` settings is not supposed to change during runtime...

src/backend/core/models.py
```diff
@@ -478,6 +498,7 @@ class Item(TreeModel, BaseModel):
     )
     numchild = models.PositiveIntegerField(default=0)
     numchild_folder = models.PositiveIntegerField(default=0)
+    main_workspace = models.BooleanField(default=False)
```
qbey, line 502: Don't you want an index on this? Maybe it's too soon

src/backend/core/models.py
```diff
@@ -255,10 +255,26 @@ def save(self, *args, **kwargs):
...
         is_adding = self._state.adding
+
         super().save(*args, **kwargs)
 
         if is_adding:
             self._convert_valid_invitations()
+            self._create_workspace()
+
+    def _create_workspace(self):
+        """Create a workspace for the user."""
```
qbey, line 266: I don't have full context about the workspace, but isn't it a bit confusing to call a folder a workspace?

src/backend/core/models.py
```diff
@@ -255,10 +255,26 @@ def save(self, *args, **kwargs):
...
+        obj = Item.objects.create_child(
+            creator=self,
+            type=ItemTypeChoices.FOLDER,
+            title=_("Workspace"),
+            main_workspace=True,
```
qbey, line 271: Should we have a constraint to allow only `main_workspace` to be set on folder?

src/backend/core/models.py
```diff
@@ -303,6 +319,10 @@ def teams(self):
         """
         return []
 
+    def get_main_workspace(self):
+        """Get the main workspace for the user."""
+        return Item.objects.get(creator=self, main_workspace=True)
```
qbey, line 324: Do we need a constaint on this item unicity?

src/backend/core/migrations/0002_item_main_workspace_alter_user_language.py
lunika: Thanks

src/backend/core/models.py
lunika: It's the main folder containing all the files and folders of the user. We call it a workspace this is why I reuse the same word here. If you think it is confusing I can rename it in main_folder ?

src/backend/core/models.py
lunika: Yes probably. But we have already lot of constraint and I looks like harder and harder to add others. You should probably take a look at existing constraint to see if there is a more efficient way to do the same thing ?

src/backend/core/models.py
lunika: I don't think so. I can have many workspaces I think

src/backend/core/models.py
```diff
@@ -255,10 +255,26 @@ def save(self, *args, **kwargs):
...
         """
         is_adding = self._state.adding
+
         super().save(*args, **kwargs)
 
         if is_adding:
             self._convert_valid_invitations()
+            self._create_workspace()
+
+    def _create_workspace(self):
```
NathanVss, line 265: WDYT about add a guard to prevent creating > 1 workspace for each user?

lunika: I don't know if we should have a strong constraint on this.

PR #38, opened 2025-03-18
Manage file mimetype

lunika:
## Purpose

- [x] save file item mimetype in upload_ended endpoint
- [x]  expose mimetype property in Item serializers

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -140,6 +140,7 @@ class ListItemSerializer(serializers.ModelSerializer):
             "type",
             "upload_state",
             "url",
+            "mimetype",
         ]
         read_only_fields = [
             "id",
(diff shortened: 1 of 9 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -576,8 +578,14 @@ def upload_ended(self, request, *args, **kwargs):
                 {"item": "This action is only available for items in PENDING state."}
             )
 
+        mime_detector = magic.Magic(mime=True)
+        file = default_storage.open(item.file_key)
+        mimetype = mime_detector.from_buffer(file.read(2048))
```
NathanVss, line 583: Isn't a call like the following enough?

```
mime_type = magic.from_file(absolute_file_path, mime=True)
```

lunika: We don't have the absolute file path, it's in an object storage and it avoids to download all the file buyt just the few kilobytes we need.

PR #60, opened 2025-03-24
✨(front) handle item title overflow in explorer

### LINKED ISSUE #58: Handle filename overflow in explorer ( with "..." and tooltip on hover )
Image

NathanVss:
When there is a an overflow, we display a tooltip on hover.

Fixes #58

https://github.com/user-attachments/assets/c24e09cf-14b1-45e8-b172-350112adf91b

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/ExplorerGrid.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/ExplorerGrid.tsx
@@ -7,7 +7,7 @@ import {
   getCoreRowModel,
   useReactTable,
 } from "@tanstack/react-table";
-import { useCallback, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { NavigationEventType, useExplorer } from "./ExplorerContext";
 import clsx from "clsx";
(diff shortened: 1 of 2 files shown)


src/frontend/apps/drive/src/features/explorer/components/ExplorerGrid.tsx
```diff
@@ -303,3 +298,46 @@ const ItemActions = () => {
...
+    return () => {
+      window.removeEventListener("resize", checkOverflow);
+    };
+  }, [item.title]);
+
+  const renderTitle = () => {
+    // We need to have the element holding the ref nested because the Tooltip component
+    // seems to make the top-most children ref null.
+    return (
+      <div style={{ display: "flex", overflow: "hidden" }}>
```
PanchoutNathan, line 326: inline style ?

NathanVss: Can you elaborate?

PanchoutNathan: why not use a className like other ?

NathanVss: Because it is a hackish technique as described in the comment, it makes no semantic sense. But if you want I can make the change

PR #68, opened 2025-03-25
✨(front) add import folders (hybrid)

NathanVss:
- [x] Import folder via dropdown
- [x] Fix nav while uploading
- [x] Support hybrid upload
- [x] Improve error handling

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/ExplorerTree.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/ExplorerTree.tsx
@@ -4,7 +4,7 @@ import { NavigationEventType, useExplorer } from "./ExplorerContext";
 import { Item } from "@/features/drivers/types";
 import { ExplorerCreateFolderModal } from "./modals/ExplorerCreateFolderModal";
 import { useState } from "react";
-import { DropdownMenu } from "@gouvfr-lasuite/ui-kit";
+import { DropdownMenu, useDropdownMenu } from "@gouvfr-lasuite/ui-kit";
 import uploadFileSvg from "@/assets/icons/upload_file.svg";
 import uploadFolderSvg from "@/assets/icons/upload_folder.svg";
 
@@ -12,7 +12,7 @@ export const ExplorerTree = () => {
   const { t } = useTranslation();
 
   // itemId is the id of the current item
-  const { item, tree, onNavigate, dropZone } = useExplorer();
+  const { item, tree, onNavigate } = useExplorer();
 
   const createFolderModal = useModal();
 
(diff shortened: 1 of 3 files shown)


src/frontend/apps/drive/src/features/explorer/components/ExplorerContext.tsx
```diff
@@ -74,6 +78,8 @@ export const ExplorerProvider = ({
       : [];
   };
 
+  const { dropZone } = useUploadZone({ item: item! });
```
PanchoutNathan, line 81: I find it more logical to initialize the hooks at the top of the file.

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -0,0 +1,323 @@
...
+      return driver.createFile(...payload);
+    },
+    onSuccess: (data, variables) => {
+      if (variables.parentId) {
+        queryClient.invalidateQueries({
+          queryKey: ["items", variables.parentId],
+        });
+      }
+    },
+  });
```
PanchoutNathan: I think it's a good practice to externalize it into a hook so that if someone else needs this feature, they don't duplicate the code. We'll have the reflex to check if the mutation hook already exists.

src/frontend/apps/drive/src/features/explorer/components/ExplorerTree.tsx
```diff
@@ -3,12 +3,16 @@ import { useTranslation } from "react-i18next";
...
+import { DropdownMenu } from "@gouvfr-lasuite/ui-kit";
+import uploadFileSvg from "@/assets/icons/upload_file.svg";
+import uploadFolderSvg from "@/assets/icons/upload_folder.svg";
 
 export const ExplorerTree = () => {
   const { t } = useTranslation();
 
   // itemId is the id of the current item
-  const { item, tree, onNavigate } = useExplorer();
+  const { item, tree, onNavigate, dropZone } = useExplorer();
```
PanchoutNathan: is unused

src/frontend/apps/drive/src/features/explorer/components/ExplorerTree.tsx
```diff
@@ -40,6 +44,8 @@ export const ExplorerTree = () => {
     );
   };
 
+  const [isImportDropdownOpen, setIsImportDropdownOpen] = useState(false);
```
PanchoutNathan: we have a useDropdownMenu hook from the design kit, it's the same thing but if we later add features to this hook, it will already be taken into account

NathanVss: We can't use spread operator on the DropdownMenu with the hook, it's sad that the API does not match

PR #79, opened 2025-03-31
Manage item title unicity constraint

lunika:
## Purpose

We managed the title unicity constraint on the item creation. We have to deal with reusing a title even if already used in a deleted items and also when renaming an existing item.

## Proposal

- [x] allow to create an item with same title of a deleted one
- [x] Check title unicity when renaming an existing one

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -417,6 +417,20 @@ class ItemQuerySet(TreeQuerySet):
         return self.filter(models.Q(link_reach=LinkReachChoices.PUBLIC))
 
 
+def _is_item_title_existing(queryset, title):
+    """Check if the title is unique in the same path."""
+    return (
+        queryset.filter(title=title)
+        .filter(
+            models.Q(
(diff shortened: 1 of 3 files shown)


src/backend/core/models.py
```diff
@@ -440,7 +440,17 @@ def create_child(self, parent=None, **kwargs):
             if parent.type != ItemTypeChoices.FOLDER:
                 raise ValidationError({"type": _("Only folders can have children.")})
 
-            if self.children(parent.path).filter(title=kwargs.get("title")).exists():
+            if (
```
NathanVss: WDYT about creating a dedicated function that performs this check, as the logic is a bit duplicated with the ItemSerializer?

`checkTitleExists(parent, title)` checks if a child already have `title`?

sampaccoud: there is also a problem that the method is called "create_child" but does a check?

lunika: The same way the save method is doing a full_clean I guess :slightly_smiling_face:

PR #80, opened 2025-03-31
🐛(back) manage filename with special characters

### LINKED ISSUE #69: File item url are not working properly
## Bug Report

**Problematic behavior**

On click some file redirect to this kind of errors:

(image)

Image

(archive)

Image

(video)

Image

But some other work properly:

Image

Image

I could not find any pattern to identify to source of the problem

lunika:
## Purpose

We want to save the original filename so it can contains white spaces
and special characters in it. When we receive the origin headers, this
one is url encoded. We need to unquote it in order to compute the S3
signature.

## Proposal

- [x] ⚡️(back) use redis as session backend in developement
- [x] 🐛(back) manage filename with special characters'

Fixes #69

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -32,7 +32,7 @@ ITEM_FOLDER = "item"
 UUID_REGEX = (
     r"[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
 )
-FILE_EXT_REGEX = r"\.[a-zA-Z]{3,4}"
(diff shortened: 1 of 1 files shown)


src/backend/drive/settings.py
```diff
@@ -628,6 +628,28 @@ class Development(Base):
     SESSION_COOKIE_NAME = "drive_sessionid"
 
     USE_SWAGGER = True
+    SESSION_CACHE_ALIAS = "session"
```
sampaccoud, line 631: shoud we port this to Docs as we are now starting to use cache extensively?

lunika: It's already in docs

NathanVss:
I am still facing those errors:

Capture d’écran 2025-03-31 à 14 45 26

PR #102, opened 2025-04-04
🐛(back) enable pg_trgm extension

NathanVss:
Without this extension we couldn't filter using %> operator.


lunika:
Already present in the main branch

PR #133, opened 2025-04-18
➕(back) add django-debug-toolbar

NathanVss:
I find it hard to track performance and make sure the new feature does not causes slow queries without this kind of tools. Let's add it.

Capture d’écran 2025-04-18 à 11 17 32


NathanVss:
Oh shit, I broke the tests, I need to fix them

lunika (approved):

PR #139, opened 2025-04-22
✨(back) adapt abilities to project

NathanVss:
## Goals

- [x] Creating a child folder does not make you owner of it. Why? If your role is downgraded on the workspace from editor to a read-only, we don't want you to still be able to act as the owner of the sub folders you created. We want you to loose you write permission on this sub folder, even if you created it. So we want the abilities to only rely on your top-most roles.
- [x] We don't want to allow moving workspaces, in any cases.
- [x] If you are >= EDITOR you can write on the entire sub folders structure, including adding files and removing them.
- [x] Only OWNER can destroy a workspace
- [x] Only OWNER and ADMIN can update ( title, description ) of a workspace

## Tests
- [x] Can't move workspace
- [x] Can't destroy workspace if editor
- [x] Can't edit anymore sub folder you created when role is downgraded


lunika:
@NathanVss I would like to discuss about modification made in this PR. I don't really agree with some abilities changed.

PR #142, opened 2025-04-22
Add codes to validation errors

sampaccoud:
## Purpose

Add standardized error codes for 40x responses

## Proposal

Proper handling of 40x error responses by the frontend requires machine codes.

DRF only returns a human readable message by default by we can rely on the standardized_errors package mentioned in DRF documentation.

### DIFF
--- a/src/backend/core/api/__init__.py
+++ b/src/backend/core/api/__init__.py
@@ -1,12 +1,13 @@
 """Drive core API endpoints"""
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError as DjangoValidationError
 
+from drf_standardized_errors.handler import exception_handler as drf_exception_handler
 from rest_framework import exceptions as drf_exceptions
-from rest_framework import views as drf_views
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
+from rest_framework.serializers import as_serializer_error
 
 
 def exception_handler(exc, context):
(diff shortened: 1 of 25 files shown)


sampaccoud:
> A huge leap forward for the frontend! 🙏
> 
> One point I wonder about, for non validation errors, why the format is like this:
> 
> ```
> {
>     "errors": [
>         {
>             "attr": None,
>             "code": "permission_denied",
>            "detail": "You do not have permission to perform this action.",
>         },
>     ],
>     "type": "client_error",
> }  
> ```
> 
> And not like the following ( knowing that in these cases there cannot be multiple errors, because the backend returns the error as soon as the first one is raised )
> 
> ```
> {
>     "code": "permission_denied",
>     "detail": "You do not have permission to perform this action.",
> }
> ```

I would guess that they want a "standardized" way of displaying errors that works for all? That's in the name of the lib :man_shrugging:  I propose not to tweak it unless you see a drawback with this approach?

src/backend/core/api/__init__.py
```diff
@@ -16,18 +17,10 @@ def exception_handler(exc, context):
...
-        elif hasattr(exc, "messages"):
-            detail = exc.messages
-
-        exc = drf_exceptions.ValidationError(detail=detail)
-
-    return drf_views.exception_handler(exc, context)
+    if isinstance(exc, DjangoValidationError):
+        exc = drf_exceptions.ValidationError(as_serializer_error(exc))
+
+    return drf_exception_handler(exc, context)
```
lunika, line 23: Did you try to remove the usage of our own exception handler and use directly the one from `drf_standardized_errors` on the `REST_FRAMEWORK` settings ?

sampaccoud: I just tried and it doesn't catch django core ValidationError:

```python
  File "/app/core/models.py", line 930, in move
    raise ValidationError(
django.core.exceptions.ValidationError: {'target': ['Only folders can be targeted when moving an item']}
INFO     request.summary:middleware.py:82
```

lunika: Too bad

lunika (approved):

sampaccoud:
@lunika I rebased and came accross hard deletion. I added some missing tests and fixed more validation errors. Please check the fixup for this.

Then I added a check on the item's restore method as explained in new commit message. Something is weird though: filtering on path__ancestors=self.path will return the self item although the ltree document says it should not :thinking:

lunika (approved):

PR #152, opened 2025-04-25
WIP 🚀(scalingo) add Scalingo deployment scripts and fix frontend build

sylvinus:
We're starting a beta deployment for [Mon Espace Collectivité](https://monespacecollectivite.incubateur.anct.gouv.fr/).

This is the branch we'll deploy to Scalingo.

Some patches could be merged to `main` later!


src/backend/drive/settings.py
```diff
@@ -93,6 +95,10 @@ class Base(Configuration):
     }
     DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
+    DATA_DIR = values.Value(
+        os.path.join("/", "data"), environ_name="DJANGO_DATA_DIR", environ_prefix=None
```
```suggestion
        os.path.join("/", "data")
```

There is no need to use these arguments if you want to use `DJANGO_DATA_DIR`

lunika: I also thing `DATA_DIR` should remain a global constant. You should move it after `BASE_DIR` and do this:

```
DATA_DIR = os.getenv("DATA_DIR", os.path.join("/", "data"))
```

src/backend/drive/settings.py
```diff
@@ -422,7 +428,9 @@ class Base(Configuration):
     THUMBNAIL_ALIASES = {}
 
     # Celery
-    CELERY_BROKER_URL = values.Value("redis://redis:6379/0")
+    CELERY_BROKER_URL = values.Value(
+        "redis://redis:6379/0", environ_name="REDIS_URL", environ_prefix=None
```
lunika: IMO this is a bad idea. We should not reuse the redis url. You may want to use a different redis or a different database in the redis instance.

src/backend/drive/settings.py
lunika: And the main reason to keep it at the top it's because this is  not a django settings. We reuse it in the settings to compute other settings.

src/nginx/servers.conf.erb
```diff
@@ -0,0 +1,86 @@
...
+        auth_request_set $authContentSha256 $upstream_http_x_amz_content_sha256;
+
+        # Pass specific headers from the auth response
+        proxy_set_header Authorization $authHeader;
+        proxy_set_header X-Amz-Date $authDate;
+        proxy_set_header X-Amz-Content-SHA256 $authContentSha256;
+
+        # Get resource from Object Storage
+        proxy_pass <%= ENV["AWS_S3_BUCKET_INTERNAL_URL"] %>;
+        proxy_set_header Host <%= ENV["AWS_S3_BUCKET_INTERNAL_HOST"] %>;
```
lunika, line 52:
```suggestion
        proxy_set_header Host <%= ENV["AWS_S3_BUCKET_INTERNAL_HOST"] %>;
        
        add_header Content-Security-Policy "default-src 'none'" always;
```

Issue #150, opened 2025-04-24
Can't select range + ctrl key


PanchoutNathan:
## Bug Report

**Problematic behavior**
If I press ctrl and select items in range mode, it deselects the ones I have already selected while I press ctrl

https://github.com/user-attachments/assets/64dbdfb6-f6f4-4596-a03d-b2b0bd2e426f

PR #163, opened 2025-04-28
🐛(frontend) many fix before prod

PanchoutNathan:
## Purpose

- The item in which we drop an item remains in its surrounded version and with its border after having dropped
- We get the workspace menu when clicking on "..." of a folder
- translations for explorer actions
- Can't select range + ctrl key #150

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/ExplorerDndProvider.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/ExplorerDndProvider.tsx
@@ -18,7 +18,7 @@ import { ExplorerDragOverlay } from "./tree/ExploreDragOverlay";
 import { TreeViewNodeTypeEnum, useTreeContext } from "@gouvfr-lasuite/ui-kit";
 import { addItemsMovedToast } from "./toasts/addItemsMovedToast";
 import { useModal } from "@openfun/cunningham-react";
-import { useState } from "react";
+import { createContext, useContext, useState } from "react";
 import {
   ConfirmationMoveState,
   ExplorerTreeMoveConfirmationModal,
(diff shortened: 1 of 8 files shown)


src/frontend/apps/drive/src/features/explorer/components/modals/workspaces/ExplorerCreateWorkspaceModal.tsx
```diff
@@ -6,7 +6,7 @@ import {
 } from "@openfun/cunningham-react";
 import { useTranslation } from "react-i18next";
 import { SubmitHandler, useForm } from "react-hook-form";
-import { WorkspaceForm } from "./WorkspaceForm";
+import { ItemFolderForm } from "./ItemFolderForm";
```
NathanVss: I think we misunderstood what we meant. I don't think folders should be edited in the same modals as workspaces, folders are not supposed to have descriptions. imo we should just add the "rename" option to mimic what we already have in the grid. It is not homogeneous that way nope ?

And even nothing LGTM, just "Informations " + "Delete" is enough imo for the MVP. 

Additionally the workspace modals is doomed to evolve a lot to include all workspaces related things.

Capture d’écran 2025-04-29 à 14 58 42
Capture d’écran 2025-04-29 à 15 02 59

NathanVss: Menu of a folder in the grid:

Capture d’écran 2025-04-29 à 15 02 36

src/frontend/apps/drive/src/features/explorer/components/ExplorerDndProvider.tsx
```diff
@@ -18,7 +18,7 @@ import { ExplorerDragOverlay } from "./tree/ExploreDragOverlay";
 import { TreeViewNodeTypeEnum, useTreeContext } from "@gouvfr-lasuite/ui-kit";
 import { addItemsMovedToast } from "./toasts/addItemsMovedToast";
 import { useModal } from "@openfun/cunningham-react";
-import { useState } from "react";
+import { createContext, useContext, useState } from "react";
```
NathanVss, line 21: Maybe I miss something but I don't quite see how this commit fixes the actual bug?

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTree.tsx
```diff
@@ -150,6 +149,16 @@ export const ExplorerTree = () => {
     }
   }, [initialOpenState, setTreeIsInitialized, treeIsInitialized]);
 
+  useEffect(() => {
```
NathanVss, line 153: plz add comment explaining quickly what the useEffect does

NathanVss (approved):

src/frontend/apps/drive/src/features/explorer/components/ExplorerDndProvider.tsx
PanchoutNathan: Before we managed the overed elements, in the EXplorer Grid, outside, so when we made the request for move, we did not have access to this state.

src/frontend/apps/drive/src/features/explorer/components/modals/workspaces/ExplorerCreateWorkspaceModal.tsx
PanchoutNathan: roll back

PR #174, opened 2025-05-05
🐛(front) prevent updates before tree initialization

PanchoutNathan:
## Purpose

Added a check in the ExplorerTree component to ensure that tree updates translations  only occur after the tree has been initialized.


NathanVss (approved):
gogogo

PR #177, opened 2025-05-09
Fix superuser sub in README

NielsCodes:
## Purpose

Make docs match the user created by `make superuser`


NathanVss:
Thank you :) !!

PR #179, opened 2025-05-12
✨(global) add posthog whitelisting feature

NathanVss:
In order to be able to allow specific users during the alpha we need to add its support to the app.

## Posthog

Capture d’écran 2025-05-12 à 16 17 07

## Error message

Capture d’écran 2025-05-12 à 16 17 39

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -6,11 +6,13 @@ from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
 from django.utils.translation import gettext_lazy as _
 
+import posthog
 import requests
 from mozilla_django_oidc.auth import (
     OIDCAuthenticationBackend as MozillaOIDCAuthenticationBackend,
 )
 
+from core.authentication.exceptions import EmailNotAlphaAuthorized
 from core.models import DuplicateEmailError, User
 
 logger = logging.getLogger(__name__)
@@ -87,6 +89,10 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
 
         sub = user_info["sub"]
         email = user_info.get("email")
+
+        if settings.FEATURES_ALPHA:
+            if not posthog.feature_enabled("alpha", email):
+                raise EmailNotAlphaAuthorized()
 
         # Get user's full name from OIDC fields defined in settings
         full_name = self.compute_full_name(user_info)
--- a/src/backend/core/authentication/exceptions.py
+++ b/src/backend/core/authentication/exceptions.py
@@ -1,0 +1,5 @@
+"""Exceptions for the authentication module."""
+
+
+class EmailNotAlphaAuthorized(Exception):
+    """Exception raised when an email is not whitelisted."""
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -16,6 +16,7 @@ from socket import gethostbyname, gethostname
 
 from django.utils.translation import gettext_lazy as _
 
+import posthog
 import sentry_sdk
 from configurations import Configuration, values
 from sentry_sdk.integrations.django import DjangoIntegration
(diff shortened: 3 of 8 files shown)


src/backend/core/authentication/exceptions.py
```diff
@@ -0,0 +1,5 @@
+"""Exceptions for the authentication module."""
+
+
+class EmailNotWhitelisted(Exception):
```
lunika:
```suggestion
class EmailNotAuthorized(Exception):
```

src/backend/drive/settings.py
```diff
@@ -616,6 +624,11 @@ def post_setup(cls):
                 "OIDC_ALLOW_DUPLICATE_EMAILS cannot be set to True simultaneously. "
             )
 
+        if cls.POSTHOG_KEY is not None:
+            posthog.api_key = cls.POSTHOG_KEY
+            posthog.host = cls.POSTHOG_HOST
+            posthog.capture("test-id", "test-event")
```
lunika: To remove ?

src/backend/core/authentication/backends.py
```diff
@@ -88,6 +90,11 @@ def get_or_create_user(self, access_token, id_token, payload):
         sub = user_info["sub"]
         email = user_info.get("email")
 
+        if settings.FEATURES_ALPHA:
+            alpha_enabled = posthog.feature_enabled("alpha", email)
+            if not alpha_enabled:
```
lunika:
```suggestion
            if not posthog.feature_enabled("alpha", email):
```

src/backend/core/authentication/backends.py
```diff
@@ -88,6 +90,11 @@ def get_or_create_user(self, access_token, id_token, payload):
...
+                raise EmailNotWhitelisted("Email not authorized for alpha")
```
lunika:
```suggestion
                raise EmailNotWhitelisted()
```

Once the exception, I think the message can be also removed.

lunika:
And at least a test is missing for the backend application

src/frontend/apps/drive/src/features/i18n/translations.json
```diff
@@ -257,6 +262,11 @@
         "title": "Stockage et partage faciles",
         "subtitle": "Stockez et partagez vos fichiers simplement dans un espace cloud collaboratif synchronisé."
       },
+      "authentication": {
+        "error": {
+          "alpha": "Cette application est en version alpha et n'est pas disponible à tous."
```
lunika:
```suggestion
          "alpha": "Cette application est en version alpha et son accès est restreint."
```

WDYT ?

src/backend/core/authentication/backends.py
NathanVss: It is on purpose that we don't want to make a Posthog feature flag call if `FEATURES_ALPHA` is not set. ( It is optional )

src/backend/drive/settings.py
NathanVss: oh yep

src/backend/core/authentication/exceptions.py
NathanVss: The specific name is on purpose to allow to make a specific redirection in the backend ( `?auth_error=alpha` to keep track of the real reason )

src/frontend/apps/drive/src/features/i18n/translations.json
NathanVss: LGTM, no gender is better

src/backend/core/authentication/exceptions.py
lunika: You can rename the exception while redirecting with this specific error. I don't see why you can't rename it.

src/backend/core/authentication/backends.py
lunika: With my suggestion you will still fetch posthog only if the settings `FEATURES_ALPHA` is set to True.

src/backend/drive/settings.py
```diff
@@ -115,6 +116,16 @@ class Base(Configuration):
         },
     }
 
+    FEATURES_ALPHA = values.Value(
```
```suggestion
    FEATURES_ALPHA = values.BooleanValue(
```

src/backend/core/authentication/exceptions.py
lunika: or you can use `EmailNotAllowed`.

See https://www.ncsc.gov.uk/blog-post/terminology-its-not-black-and-white

src/backend/core/authentication/backends.py
NathanVss: Read to quick, k

src/backend/core/authentication/exceptions.py
NathanVss: Right, I'm renaming.

lunika (approved):

PR #190, opened 2025-05-19
🐛(frontend) multiple fix before alpha 

PanchoutNathan:
## Purpose

For this PR, I tested the app and debugged the bugs found as I went along.

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/ExplorerBreadcrumbs.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/ExplorerBreadcrumbs.tsx
@@ -56,10 +56,9 @@ export const ExplorerBreadcrumbs = () => {
     });
   };
 
-  const breadcrumbsItems = useMemo(
-    () => getBreadcrumbsItems(),
-    [item, treeIsInitialized]
-  );
+  const breadcrumbsItems = useMemo(() => {
+    return getBreadcrumbsItems();
+  }, [item, treeIsInitialized, treeContext?.treeData.nodes]);
 
   if (!item || !treeIsInitialized) {
     return null;


src/frontend/apps/drive/src/features/explorer/components/modals/share/WorkspaceShareModal.tsx
```diff
@@ -60,11 +62,27 @@ export const WorkspaceShareModal = ({
...
+  /**
+   * Keep previous search result when changing the query value to avoid flickering
+   * Because when it's the first time we have this value, the queryKey changes and by default react-query returns undefined
+   */
   const { data: users, isLoading: isLoadingUsers } = useUsers(
     { q: queryValue },
-    { enabled: !!queryValue && queryValue !== "" }
+    {
+      enabled: queryValue !== undefined && queryValue !== "",
+      initialData: previousSearchResult.current,
```
NathanVss: It seems placeholderData ( previously keepPreviousData ) is made for this use case. WDYT about using this build-in mecanism, maybe it will reduce to complexity added ? https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5/#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function

src/frontend/apps/drive/src/features/explorer/components/ExplorerFilters.tsx
```diff
@@ -41,6 +31,23 @@ export const ExplorerFilters = () => {
           </div>
         ),
         value: "file",
+        showSeparator: true,
+      },
+      {
+        label: t("explorer.filters.type.options.reset"),
```
NathanVss, line 37: would be great to have smth like this built in the ui-kit

src/frontend/apps/drive/src/features/i18n/translations.json
```diff
@@ -308,7 +311,9 @@
           "last_update": "Dernière modification",
           "empty": {
             "caption": "Déposez vos fichiers ici",
-            "cta": "Ou cliquez sur \"Créer\""
+            "caption_no_create": "Vous n'avez pas les droits pour ajouter des fichiers",
```
NathanVss, line 314: WDYT: "Aucun documents ici" ?

NathanVss: As we already inform in the CTA that readers cannot add files, at least we need to tell the user that there is no files

NathanVss (approved):

src/frontend/apps/drive/src/features/explorer/components/ExplorerFilters.tsx
PanchoutNathan: yes

PR #191, opened 2025-05-19
✨(cors) add a management command to set CORS on the bucket

sylvinus:
This is useful when creating new instances on external S3 storage providers such as Scaleway.
https://www.scaleway.com/en/docs/object-storage/api-cli/setting-cors-rules/#deleting-the-cors-configuration-of-a-bucket

@sylvaingi did the same manually before I think.

Beware, the S3 key must have the rights to set the CORS config. If not you can switch on your personal key.

### DIFF
--- a/src/backend/core/management/commands/bucket_cors.py
+++ b/src/backend/core/management/commands/bucket_cors.py
@@ -1,0 +1,76 @@
+"""Management command to get or set the CORS configuration of the storage bucket"""
+
+import json
+
+from django.conf import settings
+from django.core.files.storage import default_storage
+from django.core.management.base import BaseCommand
+
+from botocore.exceptions import ClientError
+
+
+class Command(BaseCommand):
+    """
+    Management command to get or set the CORS configuration of the storage bucket
+    """
+
+    help = "Get or set the CORS configuration of the storage bucket based on the Django settings"
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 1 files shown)


src/backend/core/management/commands/bucket_cors.py
```diff
@@ -0,0 +1,76 @@
...
+                    ]
+                }
+
+                s3_client.put_bucket_cors(
+                    Bucket=default_storage.bucket_name, CORSConfiguration=cors_config
+                )
+                self.stdout.write(
+                    self.style.SUCCESS("CORS configuration successfully updated.")
+                )
+            except (ClientError, json.JSONDecodeError, FileNotFoundError) as e:
```
lunika: From where come from the `FileNotFoundError` and `json.JSONDecodeError` ? Probably you should put the try/except just around the the call to `put_bucket_cors` otherwise we will don't know from where the error come from

NathanVss: Or maybe no try/except so we can get full awareness about the stack trace in case of host-specific errors and stuff. ( I'm thinking of Outscale for instance which has some custom implem )

sylvinus:
FYI, just used this successfully for a second time

NathanVss:
@lunika do you think we can merge that?

lunika:
Yes definitely

PR #196, opened 2025-05-21
🔧(backend) support _FILE for secret environment variables

soyouzpanda:
Allow configuration variables that handles secrets, like `DJANGO_SECRET_KEY` to be able to read from a file which is given through an environment file.

For example, if `DJANGO_SECRET_KEY_FILE` is set to `/var/lib/meet/django-secret-key`, the value of `DJANGO_SECRET_KEY` will be the content of `/var/lib/meet/django-secret-key`.


NathanVss:
Thoughts @lunika ?

lunika:
It has already been merged on docs, it can be merged here. Without the modification on the `pyproject.toml` file

PR #197, opened 2025-05-21
🐛(back) fix mail language and url

NathanVss:
The link in the link was not correct, also added french translation for the sent mails. The language used was not the language of the user, I replaced with the most up to date approach used on Docs.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1062,13 +1062,12 @@ class ItemAccessViewSet(
     def perform_create(self, serializer):
         """Add a new access to the item and send an email to the new added user."""
         access = serializer.save()
-        language = self.request.headers.get("Content-Language", "en-us")
 
         access.item.send_invitation_email(
             access.user.email,
             access.role,
             self.request.user,
-            language,
+            self.request.user.language or settings.LANGUAGE_CODE,
         )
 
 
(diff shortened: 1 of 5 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -1062,13 +1062,12 @@ class ItemAccessViewSet(
...
         """Add a new access to the item and send an email to the new added user."""
         access = serializer.save()
-        language = self.request.headers.get("Content-Language", "en-us")
 
         access.item.send_invitation_email(
             access.user.email,
             access.role,
             self.request.user,
-            language,
+            self.request.user.language or settings.LANGUAGE_CODE,
```
lunika, line 1070:
```suggestion
            self.request.user.language,
```

It is not enough ? The default language is already `settings.LANGUAGE_CODE` and with the permission you must be connected.
But maybe I'm missing something

src/backend/core/tests/items/test_api_item_accesses_create.py
```diff
@@ -115,6 +116,8 @@ def test_api_item_accesses_create_authenticated_administrator(via, mock_user_tea
     An email should be sent to the accesses to notify them of the adding.
     """
     user = factories.UserFactory()
+    user.language = settings.LANGUAGE_CODE
+    user.save()
```
lunika:
```suggestion
    user = factories.UserFactory(language=settings.LANGUAGE_CODE)
```

src/backend/core/api/viewsets.py
NathanVss: I took the exact code from Docs, I bet there was a reason, just in case for maximum ceinture-bretelles

PR #200, opened 2025-05-22
🔧(backend) allow `DATA_DIR` to be configurable

soyouzpanda:
`DATA_DIR` was always set to `/data`


lunika:
It is already managed in PR #152

Issue #189, opened 2025-05-18
drive-frontend-dev-1 container exits: missing next binary during dev startup


RooTooRD:
## Bug Report

**Problematic behavior**
When running the project using the provided Makefile with `make run-with-frontend`, the `drive-frontend-dev-1` container exits unexpectedly.

The logs show the following error:

```
$ next dev
/bin/sh: next: not found
error Command failed with exit code 127.
```

**Expected behavior/code**
The `drive-frontend-dev-1` container should stay running and serve the frontend development server at `http://localhost:3000`.

**Steps to Reproduce**
1. Clone the repository:
   ```sh
   git clone https://github.com/suitenumerique/drive.git
   cd drive
   ```
2. Bootstrap the project:
   ```sh
   make bootstrap
   ```
3. Run the frontend container:
   ```sh
   make run-with-frontend
   ```
4. Observe that `drive-frontend-dev-1` exits with an error.

**Environment**
- Drive version: `main` branch (commit: `846e3f8`)
- Host OS: Ubuntu (WSL on Windows)

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

wwagops:
Hi 👋 I’ve just opened **PR <#206>** that fixes this issue.

**Root cause**  
The `frontend-dev` service bind-mounted the whole `./src/frontend` directory into `/home/frontend/`.  
That mount masked the `node_modules` folder created during the Docker build stage, so the binary `node_modules/.bin/next` was missing and the container exited with:

/bin/sh: next: not found

**Fix**  
The PR narrows the bind mount to `./src/frontend/apps/drive` (and optional local packages).  
This leaves `node_modules` intact inside the container while still allowing hot-reload on the app code.

Feel free to review and let me know if any adjustments are needed. 🚀

RooTooRD:
hi
I’ve tested your solution, and it successfully resolves the problem. Thanks for the fix!

PR #206, opened 2025-06-04
:bug:(frontend) fix bind-mount hiding node_modules

wwagops:
Narrow the bind-mount to ./src/frontend/apps/drive so that the node_modules folder created during the image build remains available inside the container. Solves: /bin/sh: next: not found when running make run-with-frontend.

Closes: #189

---

## Purpose

Running the project with `make run-with-frontend` made the container
`drive-frontend-dev-1` exit immediately:

$ next dev
/bin/sh: next: not found

The full `./src/frontend → /home/frontend/` bind-mount hid the
`node_modules` directory produced during the Docker build stage, removing
`node_modules/.bin/next` from the container’s `PATH`.

---

## Proposal

Bind-mount **only** the sources that need hot reload (`apps/drive`) and
stop masking the dependencies installed in the image.

```yaml
# docker-compose.yml — service frontend-dev
volumes:
-  - ./src/frontend/:/home/frontend/
+  - ./src/frontend/apps/drive:/home/frontend/apps/drive

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to
 
 ## [Unreleased]
 
+# - 🔧(frontend) fix bind-mount that hid node_modules in frontend-dev container #XYZ
+
 ## Added
 
 ## Changed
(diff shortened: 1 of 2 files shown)


lunika:
We made modification around this topic in PR #298.

PR #248, opened 2025-07-02
Feat/custom feedback

NathanVss:


src/backend/drive/settings.py
```diff
@@ -418,6 +418,28 @@ class Base(Configuration):
...
     )
+    FRONTEND_MORE_LINK = values.Value(
+        "https://docs.numerique.gouv.fr/docs/fa0aba15-e119-4185-b466-a4b37ad95950/",
+        environ_name="FRONTEND_MORE_LINK",
+        environ_prefix=None,
+    )
+    FRONTEND_FEEDBACK_BUTTON_SHOW = values.BooleanValue(
+        default=False, environ_name="FRONTEND_FEEDBACK_BUTTON_SHOW", environ_prefix=None
+    )
+    # For instance, you might want to bind this button to a external library like
```
```suggestion
    # For instance, you might want to bind this button to an external library like 
```

src/backend/drive/settings.py
```diff
@@ -418,6 +418,28 @@ class Base(Configuration):
     FRONTEND_THEME = values.Value(
         None, environ_name="FRONTEND_THEME", environ_prefix=None
     )
+    FRONTEND_MORE_LINK = values.Value(
+        "https://docs.numerique.gouv.fr/docs/fa0aba15-e119-4185-b466-a4b37ad95950/",
```
lunika: I see no reason to put this default, should be `None` IMO

src/backend/drive/settings.py
```diff
@@ -418,6 +418,28 @@ class Base(Configuration):
     FRONTEND_THEME = values.Value(
         None, environ_name="FRONTEND_THEME", environ_prefix=None
     )
+    FRONTEND_MORE_LINK = values.Value(
```
lunika, line 421: You can probably group all these settings in one ? And build dynamically the feedback buttons based on this settings ? It's a little bit weird to ship code with `tchap` as name, it's specific to our use case

NathanVss: It would imply some more customization, also in translations which imply more complex logic via settings which we don't have at the moment. 

But ultimately I agree with you

src/frontend/apps/drive/src/features/feedback/Feedback.tsx
```diff
@@ -8,42 +8,69 @@ import {
...
         color="tertiary"
         icon={<Icon name="info" />}
         className="c__feedback__button"
-        onClick={modal.open}
+        onClick={() => {
+          if (config?.FRONTEND_FEEDBACK_BUTTON_IDLE) {
+            return;
+          }
+          modal.open();
+        }}
```
PanchoutNathan: create method !

lunika:
IMO we have to go in the direction of having a configuration file. This is what is made on docs and allow us to customize lot of think in the interface

PR #253, opened 2025-07-11
✨(front) add theming via /config route

NathanVss:
## Purpose

Previously the theming was not customisable. Now we can use the backend `FRONTEND_THEME` var to inject the current theme. 

Also did a bit of refactor to introduce a ConfigProvider that displays a loader when config is not loaded ( we don't want to display the interface before the actuel `FRONTEND_THEME` is loaded right ? ) and then provide the config via a context globally.  ( Moved to old usages of useApiConfig to useConfig so ) 

I made sure to have the CunninghamProvider above ConfigProvider because the config provider needs to, at least, display a loader during loader, and then after set the actual theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to
 - ✨(front) add move modal #213
 - ✨(front) update the homepage to alpha #234
 - ✨(global) add customization to feedbacks
+- ✨(front) make frontend themable
 
 ## Changed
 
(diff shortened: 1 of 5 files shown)


sylvinus (approved):
LGTM

@NathanVss you prefer to have FRONTEND_HIDE_GAUFRE and potentially many more, instead of a `FRONTEND_ADD_CSS=".gaufre {display:none}"` ?

NathanVss:
> LGTM
> 
> @NathanVss you prefer to have FRONTEND_HIDE_GAUFRE and potentially many more, instead of a `FRONTEND_ADD_CSS=".gaufre {display:none}"` ?

Using plain CSS for this component means it still loads 🤔

PR #256, opened 2025-07-17
🐛(global) fix wrong Content-Type on specific s3 implementations

NathanVss:
## Issue

When requesting file from the Scaleway object storage the Content-Type in the response header contain value like `multipart/form-data; boundary=----WebKitFormBoundaryP....` instead of expected values containing mime type like `application/pdf`. 

This was causing issue when previewing pdf, it was triggering a download instead of the browser native preview. 

This was only happening on Scaleway object storage, it is totally fine with MinIO which seems to guess itself the Content-Type based on Claude conversation. 

## Proposal

In its example here ( https://www.scaleway.com/en/docs/object-storage/api-cli/post-object/#examples ) we need to include Content-Type in the form data fields BEFORE ( ⚠️⚠️⚠️⚠️ ) the `file` field ( otherwise it doesn't work ) and also include a blank condition `["starts-with", "$Content-Type", ""]` when generating the policy otherwise it fails. 

This approach works both with MinIO and Scaleway object storage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,5 +24,6 @@ and this project adheres to
 
 - 🐛(i18n) fix language detection and rendering
 - 🌐(front) add english translation for rename modal
+- 🐛(global) fix wrong Content-Type on specific s3 implementations
 
 ## Deleted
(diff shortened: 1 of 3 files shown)


PanchoutNathan (approved):

sylvinus (approved):
Some dark magic in there

PR #262, opened 2025-07-18
🐛(front) fix rename modal

PanchoutNathan:
## Purpose

Before the release, we noticed a few minor issues:

- The rename window wouldn't close.
- The extension was present in the rename window.

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGridActionsCell.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGridActionsCell.tsx
@@ -141,14 +141,14 @@ export const EmbeddedExplorerGridActionsCell = (
         {renameModal.isOpen && (
           <ExplorerRenameItemModal {...renameModal} item={item} key={item.id} />
         )}
-        {canShareWorkspace && (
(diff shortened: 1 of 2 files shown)


src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGridActionsCell.tsx
```diff
@@ -141,14 +141,14 @@ export const EmbeddedExplorerGridActionsCell = (
...
-        {canShareWorkspace && (
+        {canShareWorkspace && shareWorkspaceModal.isOpen && (
           <WorkspaceShareModal
             {...shareWorkspaceModal}
             item={item}
             key={item.id}
           />
         )}
-        {canShareFile && (
+        {canShareFile && shareFileModal.isOpen && (
```
NathanVss, line 151: Weird that is doesn't close by itself as the closed state should be passed to the modal component, nope? 

I see that this is how we did before (https://github.com/suitenumerique/drive/blob/44d66f6941f09cbbda4a5cd4ea1c866103d76201/src/frontend/apps/drive/src/features/explorer/components/grid/ExplorerGridActionsCell.tsx#L129-L134) the refacto but anyway weird 🤔 do you know why ?

NathanVss (approved):

PR #264, opened 2025-07-18
🔧(cron): add pgdump cron on scalingo deployment

zeylos:
add scalingo_pgdump script to dump the database and upload it to scaleway s3.
add cron.json to schedule pgdump everyday at noon UTC.

### DIFF
--- a/bin/scalingo_pgdump.sh
+++ b/bin/scalingo_pgdump.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+mkdir -p /tmp/pgdump
+cd /tmp/pgdump
+
+RESTIC_VERSION=0.18.0
+curl -fsSL --remote-name-all "https://github.com/restic/restic/releases/download/v${RESTIC_VERSION}/restic_${RESTIC_VERSION}_linux_amd64.bz2" \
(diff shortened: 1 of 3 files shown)


bin/scalingo_pgdump.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Actually dump and upload to scaleway s3
+FILENAME="${APP}_pgdump.sql"
+
+pg_dump --clean --if-exists --format c --dbname "${SCALINGO_POSTGRESQL_URL}" --no-owner --no-privileges --no-comments --exclude-schema 'information_schema' --exclude-schema '^pg_*' --file "${FILENAME}"
+
+export AWS_ACCESS_KEY_ID=${BACKUP_PGSQL_S3_KEY}
+export AWS_SECRET_ACCESS_KEY=${BACKUP_PGSQL_S3_SECRET}
+export RESTIC_PASSWORD=${BACKUP_PGSQL_ENCRYPTION_PASS}
+export RESTIC_REPOSITORY=s3:https://s3.fr-par.scw.cloud/${BACKUP_PGSQL_S3_BUCKET}/${APP}
```
sylvinus, line 31: Ideally the whole RESTIC_REPOSITORY should be an env var too, so that we don't hardcode Scaleway anywhere?

PR #266, opened 2025-07-21
✨(front) public workspaces !

NathanVss:
Allow sharing public workspaces from the share modal.

### Make a workspace public
Capture d’écran 2025-07-21 à 18 00 20

### Accessing a public workspace

Capture d’écran 2025-07-21 à 17 59 18

### Added 401, 403 proper pages
 
Capture d’écran 2025-07-24 à 15 16 06
Capture d’écran 2025-07-24 à 15 15 39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,18 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(front) Add public workspaces
+- ✨(front) Add 401, 403 pages
+- ✨(front) Add redirect after login logic
+
 ## [v0.1.1] - 2025-07-30
 
 ### Fixed
 
 - 🐛(backend) stop decreasing twice the numchild in deletion process #284
-- ⚡️(backend) optimize trashbin endpoint #276 
+- ⚡️(backend) optimize trashbin endpoint #276
 - ♻️(backend) modify sdk-relay to use DRF viewset and serializers #269
 
 ## [v0.1.0] - 2025-07-25
@@ -39,7 +45,6 @@ and this project adheres to
 - 🐛(i18n) fix language detection and rendering
 - 🌐(front) add english translation for rename modal
 - 🐛(global) fix wrong Content-Type on specific s3 implementations
-
 
 [unreleased]: https://github.com/suitenumerique/drive/compare/v0.1.1...main
 [0.1.1]: https://github.com/suitenumerique/drive/releases/v0.1.1


src/frontend/apps/drive/package.json
```diff
@@ -13,7 +13,7 @@
     "node": ">=22.0.0 <23.0.0"
   },
   "dependencies": {
-    "@gouvfr-lasuite/ui-kit": "0.9.0",
+    "@gouvfr-lasuite/ui-kit": "0.10.0",
```
PanchoutNathan: You can use the 0.11.0 now

src/frontend/apps/drive/src/features/api/fetchApi.ts
```diff
@@ -42,8 +54,23 @@ export const fetchAPI = async (
...
+
+  if (response.status === 403) {
+    if (options?.redirectOn40x ?? true) {
+      // We don't want to save the attempted url when having a 403 error because
+      // it would be a redirect loop and it means we know that the user is not
+      // allowed to access the page.
+      redirect("/403", false);
+      // So that the app can handle the error and not show a toast by verifying the error code.
+      throw new APIError(response.status);
+    }
```
PanchoutNathan:
```suggestion 
 const redirectOn40x = options?.redirectOn40x ?? true;
  if (response.status === 401 && redirectOn40x) {
      redirect("/401");
      // So that the app can handle the error and not show a toast by verifying the error code.
      throw new APIError(response.status);
  }

  if (response.status === 403 && redirectOn40x) {
      // We don't want to save the attempted url when having a 403 error because
      // it would be a redirect loop and it means we know that the user is not
      // allowed to access the page.
      redirect("/403", false);
      // So that the app can handle the error and not show a toast by verifying the error code.
      throw new APIError(response.status);
  }
```

src/frontend/apps/drive/src/features/ui/components/generic-disclaimer/GenericDisclaimer.scss
```diff
@@ -0,0 +1,25 @@
+.drive__generic-disclaimer {
```
PanchoutNathan, line 1: I don't know where it comes from, but we have a scroller in the center, on mobile and desktop. There's no need for it.

src/frontend/apps/drive/src/pages/_app.tsx
```diff
@@ -57,7 +65,9 @@ const onError = (error: Error, query: unknown) => {
 
 const queryClient = new QueryClient({
   mutationCache: new MutationCache({
-    onError: (error, variables, context, mutation) => onError(error, mutation),
+    onError: (error, variables, context, mutation) => {
+      onError(error, mutation);
+    },
```
PanchoutNathan, line 70: why ?

src/frontend/apps/drive/src/features/auth/Auth.tsx
```diff
@@ -5,7 +5,7 @@ import { User } from "@/features/auth/types";
 import { baseApiUrl } from "../api/utils";
```
PanchoutNathan, line 5: God thx !

PanchoutNathan (approved):

src/frontend/apps/drive/src/features/ui/components/generic-disclaimer/GenericDisclaimer.scss
NathanVss: Not sure to understand what you mean? This page is meant to be a generic page to use to display in full page any information, such as 401, 403 errors

src/frontend/apps/drive/src/pages/_app.tsx
NathanVss: Auto formatting

PR #269, opened 2025-07-22
♻️(backend) modify sdk-relay to use DRF viewset and serializers

lunika:
## Purpose

No data validation was made on the post endpoint, that can lead to unwanted behavior when constructed the redis key and we ensure the event is a valid json.
Also, we use a VewSet instead of the APIView, allowing us to use a router to define the urls.

## Proposal

- [x] ♻️(backend) modify sdk-relay to use DRF viewset and serializers

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1250,6 +1250,14 @@ class SDKRelayEventViewset(drf.viewsets.ViewSet):
 
     throttle_scope = "sdk_event_relay"
 
+    def get_permissions(self):
+        """
+        Return the list of permissions that this view requires.
+        """
+        if self.action == "create":
+            return [permissions.IsAuthenticated()]
+        return super().get_permissions()
+
     def handle_cors(self, request, response):
         """Handle CORS preflight requests."""
         # Same approach as here:
--- a/src/backend/core/tests/test_api_sdk_relay.py
+++ b/src/backend/core/tests/test_api_sdk_relay.py
@@ -5,6 +5,8 @@ Test SDK relay API endpoints.
 import pytest
 from rest_framework.test import APIClient
 
+from core import factories
+
 pytestmark = pytest.mark.django_db
 
 
@@ -18,6 +20,17 @@ def test_api_sdk_relay_get_event_anonymous():
     assert response.json() == {}
 
 
+def test_api_sdk_relay_register_event_anonymous():
+    """Anonymous users should not be allowed to register an event."""
+    client = APIClient()
+    response = client.post(
+        "/api/v1.0/sdk-relay/events/",
+        {"token": "1Az6SO4CE7JAl9hE96dXl7145nghwZNP", "event": {"type": "test"}},
+        format="json",
+    )
+    assert response.status_code == 401
+
+
 def test_api_sdk_relay_register_event():
     """Anonymous users should be allowed to register an event."""
     client = APIClient()
@@ -28,6 +41,8 @@ def test_api_sdk_relay_register_event():
     assert "Access-Control-Allow-Origin" not in response
     assert response.json() == {}
 
+    user = factories.UserFactory()
+    client.force_authenticate(user)
     response = client.post(
         "/api/v1.0/sdk-relay/events/",
         {"token": "1Az6SO4CE7JAl9hE96dXl7145nghwZNP", "event": {"type": "test"}},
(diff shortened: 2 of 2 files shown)


src/backend/core/api/serializers.py
```diff
@@ -532,3 +533,26 @@ class MoveItemSerializer(serializers.Serializer):
...
+    token = serializers.RegexField(regex=r"^[0-9a-zA-Z]{32}$", required=True)
+    event = serializers.JSONField(required=True)
+
+    def validate_event(self, value):
+        """Validate that the event JSON data doesn't exceed maximum length."""
+
+        # Convert the JSON data to a string to check its length
+        json_string = json.dumps(value)
+        # One selected file could use up to 1000 characters, this limit is set to
+        # avoid DDOS attacks to fullfill the redis.
```
qbey, line 550: I don't remember, is there a rate limit on this endpoint?

src/backend/core/api/viewsets.py
```diff
@@ -1242,33 +1216,59 @@ def get(self, request):
...
+        if self.action == "create":
+            return [permissions.IsAuthenticated()]
+        return super().get_permissions()
+
+    def handle_cors(self, request, response):
+        """Handle CORS preflight requests."""
+        # Same approach as here:
+        # https://github.com/adamchainz/django-cors-headers/blob/b04460f37cbf458984bb377d8e6afb56776c3465/src/corsheaders/middleware.py#L96
+        origin = request.headers.get("origin")
+        if origin and origin in settings.SDK_CORS_ALLOWED_ORIGINS:
```
qbey, line 1266: ❤️

src/backend/core/api/viewsets.py
```diff
@@ -1242,33 +1216,59 @@ def get(self, request):
...
+            serializer.validated_data.get("event"),
+        )
         return drf.response.Response(status=status.HTTP_201_CREATED)
+
+    def options(self, request, *args, **kwargs):
+        """
+        OPTIONS /api/v1.0/sdk-relay/events/<token>/
+        Handle CORS preflight requests.
+        """
+        response = drf.response.Response(status=status.HTTP_200_OK)
```
qbey, line 1299: Why don't you get the response from the `super().options(...)` ? It's not mandatory but a comment could just say way (I guess it's better to have an empty OPTIONS like you did)

src/backend/core/api/viewsets.py
```diff
@@ -1242,33 +1216,59 @@ def get(self, request):
...
+            return [permissions.IsAuthenticated()]
+        return super().get_permissions()
+
+    def handle_cors(self, request, response):
+        """Handle CORS preflight requests."""
+        # Same approach as here:
+        # https://github.com/adamchainz/django-cors-headers/blob/b04460f37cbf458984bb377d8e6afb56776c3465/src/corsheaders/middleware.py#L96
+        origin = request.headers.get("origin")
+        if origin and origin in settings.SDK_CORS_ALLOWED_ORIGINS:
+            response["Access-Control-Allow-Origin"] = origin
```
qbey: Don't you want to reuse the constants like `ACCESS_CONTROL_ALLOW_ORIGIN` and so?

src/backend/core/api/viewsets.py
```diff
@@ -1242,33 +1216,59 @@ def get(self, request):
...
+        )
         return drf.response.Response(status=status.HTTP_201_CREATED)
+
+    def options(self, request, *args, **kwargs):
+        """
+        OPTIONS /api/v1.0/sdk-relay/events/<token>/
+        Handle CORS preflight requests.
+        """
+        response = drf.response.Response(status=status.HTTP_200_OK)
+        self.handle_cors(request, response)
```
qbey, line 1300: Do you think the OPTIONS should also adds the specific headers like https://github.com/adamchainz/django-cors-headers/blob/main/src/corsheaders/middleware.py#L125-L129 ^^'

src/backend/core/api/serializers.py
lunika: Yes, there is a throttle https://github.com/suitenumerique/drive/issues/294 ^^

src/backend/core/api/viewsets.py
lunika: I take a look, the answer given by the parent function is useless IMO

src/backend/core/api/viewsets.py
lunika: Yes good idea

src/backend/core/api/viewsets.py
lunika: I want to answer no. While we don't need it we don't add it.

Issue #274, opened 2025-07-24
Slow /trashbin request


lunika:
## Bug Report

**Problematic behavior**

The `/api/v1.0/trashbin` is really slow, we need to investigate why.

lunika:
The generated sql query is this one:

```sql
SELECT "drive_item"."path",
       "drive_item"."id",
       "drive_item"."created_at",
       "drive_item"."updated_at",
       "drive_item"."title",
       "drive_item"."link_reach",
       "drive_item"."link_role",
       "drive_item"."creator_id",
       "drive_item"."deleted_at",
       "drive_item"."ancestors_deleted_at",
       "drive_item"."hard_deleted_at",
       "drive_item"."filename",
       "drive_item"."type",
       "drive_item"."upload_state",
       "drive_item"."numchild",
       "drive_item"."numchild_folder",
       "drive_item"."mimetype",
       "drive_item"."main_workspace",
       "drive_item"."size",
       "drive_item"."description",
       ARRAY((SELECT U0."role" AS "role" FROM "drive_item_access" U0 INNER JOIN "drive_item" U2 ON (U0."item_id" = U2."id") WHERE (U0."user_id" = '2e7a6c4e3c314f34acba0bf406b1163f'::uuid AND U2."path" @> ("drive_item"."path")) ORDER BY U0."created_at" DESC)) AS "user_roles",
       "drive_user"."password",
       "drive_user"."last_login",
       "drive_user"."is_superuser",
       "drive_user"."id",
       "drive_user"."created_at",
       "drive_user"."updated_at",
       "drive_user"."sub",
       "drive_user"."full_name",
       "drive_user"."short_name",
       "drive_user"."email",
       "drive_user"."admin_email",
       "drive_user"."language",
       "drive_user"."timezone",
       "drive_user"."is_device",
       "drive_user"."is_staff",
       "drive_user"."is_active"
  FROM "drive_item"
  LEFT OUTER JOIN "drive_user"
    ON ("drive_item"."creator_id" = "drive_user"."id")
[...]

PR #276, opened 2025-07-24
⚡️(backend) optimize trashbin endpoint

lunika:
## Purpose

The trashbin endpoint can be slow sometimes. We identified a bottleneck in the sql query made to fetch the delete items that can be managed by the current user.
All the details can be found in the issue #274

## Proposal

- [x] ⚡️(backend) optimize trashbin endpoint

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -479,6 +479,27 @@ class ItemViewSet(
             user_roles=db.Value([], output_field=output_field),
         )
 
+    def annotate_has_owner_role(self, queryset):
+        """
+        Optimized annotation to check if user has owner role on item or ancestors.
+        This is more efficient than the full user_roles annotation when we only
+        need to check for owner access.
+        """
+        user = self.request.user
+
+        if user.is_authenticated:
+            user_teams = user.teams if hasattr(user, "teams") else []
+
+            owner_access_exists = models.ItemAccess.objects.filter(
+                db.Q(user=user) | db.Q(team__in=user_teams),
+                role=models.RoleChoices.OWNER,
+                item__path__ancestors=db.OuterRef("path"),
+            )
+
+            return queryset.annotate(has_owner_role=db.Exists(owner_access_exists))
+
+        return queryset.annotate(has_owner_role=db.Value(False))
+
     def get_queryset(self):
         """Get queryset performing all annotation and filtering on the item tree structure."""
         user = self.request.user
@@ -696,17 +717,40 @@ class ItemViewSet(
 
         The selected items are those deleted within the cutoff period defined in the
         settings (see TRASHBIN_CUTOFF_DAYS), before they are considered permanently deleted.
+
+        Optimized version that uses EXISTS instead of expensive subqueries to check
+        owner access on items or their ancestors.
         """
-        queryset = self.queryset.select_related("creator").filter(
-            deleted_at__isnull=False,
-            deleted_at__gte=models.get_trashbin_cutoff(),
+        user = request.user
+
+        # Build the EXISTS subquery to check if user has owner access
+        # to the item or any of its ancestors
+        owner_access_exists = models.ItemAccess.objects.filter(
+            db.Q(user=user) | db.Q(team__in=user.teams),
+            role=models.RoleChoices.OWNER,
+            item__path__ancestors=db.OuterRef("path"),
         )
-        queryset = self.annotate_user_roles(queryset)
-        queryset = queryset.filter(user_roles__contains=[models.RoleChoices.OWNER])
+
+        # Filter trashbin items to only those where user has owner access
+        # Before we were filtering on the user_roles annotation, but it was too slow
+        # Here the optimization is to filter on the owner_access_exists subquery
+        # which is much faster.
+        queryset = (
+            self.queryset.select_related("creator")
+            .filter(
+                deleted_at__gte=models.get_trashbin_cutoff(),
+            )
+            .filter(db.Exists(owner_access_exists))
+        )
+
+        # Apply filtering similar to children method
         filterset = ItemFilter(request.GET, queryset=queryset)
         if not filterset.is_valid():
             raise drf.exceptions.ValidationError(filterset.errors)
         queryset = filterset.qs
+
+        # Only annotate with user roles for the filtered set if needed by serializer
+        queryset = self.annotate_user_roles(queryset)
 
         return self.get_response_for_queryset(queryset)
 
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -345,12 +345,12 @@ class Base(Configuration):
             "rest_framework.parsers.JSONParser",
             "nested_multipart_parser.drf.DrfNestedParser",
         ],
-        "DEFAULT_RENDERER_CLASSES": [
-            # 🔒️ Disable BrowsableAPIRenderer which provides forms allowing a user to
-            # see all the data in the database (ie a serializer with a ForeignKey field
-            # will generate a form with a field with all possible values of the FK).
-            "rest_framework.renderers.JSONRenderer",
-        ],
+        # "DEFAULT_RENDERER_CLASSES": [
+        #     # 🔒️ Disable BrowsableAPIRenderer which provides forms allowing a user to
+        #     # see all the data in the database (ie a serializer with a ForeignKey field
+        #     # will generate a form with a field with all possible values of the FK).
+        #     "rest_framework.renderers.JSONRenderer",
+        # ],
         "EXCEPTION_HANDLER": "core.api.exception_handler",
         "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
         "PAGE_SIZE": 20,
--- a/src/backend/core/api/permissions.py
+++ b/src/backend/core/api/permissions.py
@@ -100,7 +100,10 @@ class AccessPermission(permissions.BasePermission):
     """Permission class for access objects."""
 
     def has_permission(self, request, view):
-        return request.user.is_authenticated or view.action != "create"
+        return request.user.is_authenticated or view.action not in [
+            "create",
+            "trashbin",
+        ]
 
     def has_object_permission(self, request, view, obj):
         """Check permission for a given object."""
--- a/src/backend/core/tests/items/test_api_items_trashbin.py
+++ b/src/backend/core/tests/items/test_api_items_trashbin.py
@@ -29,13 +29,7 @@ def test_api_items_trashbin_anonymous(reach, role):
 
     response = APIClient().get("/api/v1.0/items/trashbin/")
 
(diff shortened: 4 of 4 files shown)


src/backend/drive/settings.py
```diff
@@ -345,12 +345,12 @@ class Base(Configuration):
...
-            # see all the data in the database (ie a serializer with a ForeignKey field
-            # will generate a form with a field with all possible values of the FK).
-            "rest_framework.renderers.JSONRenderer",
-        ],
+        # "DEFAULT_RENDERER_CLASSES": [
+        #     # 🔒️ Disable BrowsableAPIRenderer which provides forms allowing a user to
+        #     # see all the data in the database (ie a serializer with a ForeignKey field
+        #     # will generate a form with a field with all possible values of the FK).
+        #     "rest_framework.renderers.JSONRenderer",
+        # ],
```
lunika: Undo

src/backend/core/api/viewsets.py
```diff
@@ -696,18 +717,41 @@ def trashbin(self, request, *args, **kwargs):
...
+            item__path__ancestors=db.OuterRef("path"),
         )
-        queryset = self.annotate_user_roles(queryset)
-        queryset = queryset.filter(user_roles__contains=[models.RoleChoices.OWNER])
+
+        # Filter trashbin items to only those where user has owner access
+        # Before we were filtering on the user_roles annotation, but it was too slow
+        # Here the optimization is to filter on the owner_access_exists subquery
+        # which is much faster.
+        queryset = (
```
NathanVss, line 717: Running an query plan analysis shows this:


```

QUERY PLAN
--
Limit  (cost=26.87..43.23 rows=1 width=535) (actual time=0.191..0.299 rows=5 loops=1)
->  Result  (cost=26.87..43.23 rows=1 width=535) (actual time=0.190..0.297 rows=5 loops=1)
->  Sort  (cost=26.87..26.87 rows=1 width=503) (actual time=0.081..0.082 rows=5 loops=1)
Sort Key: drive_item.created_at
Sort Method: quicksort  Memory: 27kB
->  Nested Loop Left Join  (cost=0.29..26.86 rows=1 width=503) (actual time=0.030..0.058 rows=5 loops=1)
Join Filter: (drive_item.creator_id = drive_user.id)
Rows Removed by Join Filter: 5
->  Nested Loop Semi Join  (cost=0.29..25.79 rows=1 width=326) (actual time=0.026..0.049 rows=5 loops=1)
Join Filter: (u2.path @> drive_item.path)
->  Seq Scan on drive_item  (cost=0.00..9.22 rows=5 width=326) (actual time=0.008..0.030 rows=5 loops=1)
Filter: ((hard_deleted_at IS NULL) AND (deleted_at >= '2025-06-24 14:45:24.615336+00'::timestamp with time zone))
Rows Removed by Filter: 93
->  Materialize  (cost=0.29..16.49 rows=1 width=139) (actual time=0.003..0.003 rows=1 loops=5)
->  Nested Loop  (cost=0.29..16.49 rows=1 width=139) (actual time=0.014..0.014 rows=1 loops=1)
->  Index Scan using unique_item_user on drive_item_access u0  (cost=0.14..8.16 rows=1 width=16) (actual time=0.006..0.006 rows=1 loops=1)
Index Cond: (user_id = 'a1803096-ceb2-41ec-a4c7-6678ed5a12bb'::uuid)
Filter: ((role)::text = 'owner'::text)
->  Index Scan using drive_item_pkey on drive_item u2  (cost=0.14..8.16 rows=1 width=155) (actual time=0.007..0.007 rows=1 loops=1)
Index Cond: (id = u0.item_id)
->  Seq Scan on drive_user  (cost=0.00..1.03 rows=3 width=177) (actual time=0.001..0.001 rows=2 loops=5)
SubPlan 1
->  Sort  (cost=16.34..16.34 rows=1 width=66) (actual time=0.041..0.041 rows=1 loops=5)
Sort Key: u0_1.created_at DESC
Sort Method: quicksort  Memory: 25kB
->  Nested Loop  (cost=0.28..16.33 rows=1 width=66) (actual time=0.019..0.039 rows=1 loops=5)
Join Filter: (u2_1.id = u0_1.item_id)
Rows Removed by Join Filter: 15
->  Index Scan using unique_item_user on drive_item_access u0_1  (cost=0.14..8.16 rows=1 width=82) (actual time=0.001..0.001 rows=5 loops=5)
Index Cond: (user_id = 'a1803096-ceb2-41ec-a4c7-6678ed5a12bb'::uuid)
->  Index Scan using drive_item_path_a306e2_gist on drive_item u2_1  (cost=0.14..8.15 rows=1 width=16) (actual time=0.005..0.007 rows=3 loops=25)
Index Cond: (path @> drive_item.path)

```

There few key warning point that we might think of at scale:
- There is a Seq Scan on drive_user 
**This would induce high CPU and slower and slower response time with users amount increasing**
-> Adding an index on `creator_id` would help

- Seq Scan on drive_item on Filter: ((hard_deleted_at IS NULL) AND (deleted_at >= :date:))
**At scale the DB will have to seq without index remove all the items removed by other users to serve this query**
-> Add index on those fields would help too

- Also this logic is using a sub query, usually those do not scale very well although we have some time ahead of before we get hit by such perf issue, but if we have thousands of users, with hundred of thousands items in the trash the query will have to perform hundred of thousands subqueries and we might suffer at that moment. But solving this would mean introducing a cache mecanism that we don't have at the moment and it would maybe be more profitable to the product to just increase the DB CPU amount instead of introducing huge complexity in the codebase to maintain a item to user user_roles cache.

WDYT?

src/backend/core/api/viewsets.py
```diff
@@ -696,18 +717,41 @@ def trashbin(self, request, *args, **kwargs):
...
-            deleted_at__gte=models.get_trashbin_cutoff(),
+        user = request.user
+
+        # Build the EXISTS subquery to check if user has owner access
+        # to the item or any of its ancestors
+        owner_access_exists = models.ItemAccess.objects.filter(
+            db.Q(user=user) | db.Q(team__in=user.teams),
+            role=models.RoleChoices.OWNER,
+            item__path__ancestors=db.OuterRef("path"),
         )
```
NathanVss, line 711: Do you think there is a way to avoid code duplication with `annotate_has_owner_role`?

src/backend/core/api/viewsets.py
qbey: My two cents:

 - I'm not sure on what amount of data you made your test, but the query planner might have chosen not to use the `creator_id` index (foreign keys are indexed) because is was not helpful. Also the query planner will take samples to determine what is the best plan, did you try to force a recompute before the explain?

  - Regarding `hard_deleted_at` I think at some point we should accept to remove the data for real :D
  
Otherwise, I would keep it as this for now, also creating new index is not always the best move and does not prevent seq scan use, depending of the similarity of the data you have in the DB, so I would wait and see :)

qbey (approved):
LGTM

src/backend/core/api/viewsets.py
PanchoutNathan: It might be worth doing a test by adding the indexes and redoing the analysis? It doesn't take much time and the gain can be great, WDYT ?

src/backend/core/api/viewsets.py
```diff
@@ -479,6 +479,27 @@ def annotate_user_roles(self, queryset):
...
+            owner_access_exists = models.ItemAccess.objects.filter(
+                db.Q(user=user) | db.Q(team__in=user_teams),
+                role=models.RoleChoices.OWNER,
+                item__path__ancestors=db.OuterRef("path"),
+            )
+
+            return queryset.annotate(has_owner_role=db.Exists(owner_access_exists))
+
+        return queryset.annotate(has_owner_role=db.Value(False))
+
```
lunika: To be removed

src/backend/core/api/viewsets.py
```diff
@@ -696,18 +717,41 @@ def trashbin(self, request, *args, **kwargs):
...
+        # Build the EXISTS subquery to check if user has owner access
+        # to the item or any of its ancestors
+        owner_access_exists = models.ItemAccess.objects.filter(
+            db.Q(user=user) | db.Q(team__in=user.teams),
+            role=models.RoleChoices.OWNER,
+            item__path__ancestors=db.OuterRef("path"),
         )
-        queryset = self.annotate_user_roles(queryset)
-        queryset = queryset.filter(user_roles__contains=[models.RoleChoices.OWNER])
```
sampaccoud, line 705: My experience in docs is that it is much faster (and simpler?) to do it in 2 queries:
```python
owned_documents_ids = models.ItemAccess.objects.filter(
    db.Q(user=user) | db.Q(team__in=user.teams),
    role=models.RoleChoices.OWNER,
).values_list("document_id", flat=True)

queryset = (
    self.queryset.select_related("creator")
    .filter(
        deleted_at__gte=models.get_trashbin_cutoff(),
    )
    .filter(ancestors__id__in=owned_document_ids)
)
```
You just need to be sure that the list of owned documents ids fits in RAM but in this case definitely even if it is a few millions... and you'll just keep it a few nanoseconds.

src/backend/core/api/viewsets.py
```diff
@@ -479,6 +479,27 @@ def annotate_user_roles(self, queryset):
             user_roles=db.Value([], output_field=output_field),
         )
 
+    def annotate_has_owner_role(self, queryset):
```
sampaccoud: Can be removed now? I agree that it should be in the trashbin method if it is not factorized at least 3 times.

lunika: Yes I forget to remove it, this method is unused

src/backend/core/api/viewsets.py
lunika: I dig a bit with your solution, It might works but with some addition on the `django_ltree` library.
First, the `owned_documents_ids` returns a queryset and this queryset is not evaluated, so it will be a subquery, this not what we want.
Second we don't want `ids` but `path` to look on all the descendants.
Then, the ltree extension has a contains operator but it does not compare between ltree type but a ltree with a lquery (see https://www.postgresql.org/docs/16/ltree.html#LTREE-DEFINITIONS). 
This type, `lquery` does not exists in the django_ltree library, if I create it I can work with and all the tests are green.

```python
        owned_documents_paths = models.ItemAccess.objects.select_related("item").filter(
            db.Q(user=user) | db.Q(team__in=user.teams),
            role=models.RoleChoices.OWNER,
        ).values_list("item__path", flat=True)
        lquery = []
        for path in owned_documents_paths:
            lquery.append(f"{path}.*")
        output_field = ArrayField(base_field=models.LqueryField())
        queryset = (
            self.queryset.select_related("creator")
            .filter(
                deleted_at__gte=models.get_trashbin_cutoff(),
            )
            .filter(path__contains=db.Value(lquery, output_field=output_field),)
        )
```

The final code look like this.
I think my initial version is easier to understand, WDYT ?

PR #280, opened 2025-07-25
 🔖(minor) bump release version 0.1.0

lunika:
### Added
    
- 🚀 Drive, A collaborative file sharing and document management platform
- ✨(front) add move modal #213
- ✨(front) update the homepage to alpha #234
- ✨(global) add customization to feedbacks
- ✨(front) add PDF, Audio, Video, Image viewers
- ✨(front) make frontend themable
- ✨(global) Add File Picker SDK
- 🔧(cron) add pgdump cron on scalingo deployment #264

### Changed

- 🐛(email) add missing email logo

### Fixed

- 🐛(i18n) fix language detection and rendering
- 🌐(front) add english translation for rename modal
- 🐛(global) fix wrong Content-Type on specific s3 implementations

PR #284, opened 2025-07-25
🐛(backend) stop decreasing twice the numchild in deletion process

lunika:
## Purpose
In the deletion process we maintain the numchild property by decreasing
when the item is soft deleted. But the delete method is also decreasing
it. We must only rely on the soft delete process and ensure that an item
can not be deleted if it was not solft_deleted first.

## Proposal

- [x] 🐛(backend) stop decreasing twice the numchild in deletion process
- [x] ✨(backend) management command to sync numchild property


lunika:
~~Changelog will be updated once #280 merged~~ :heavy_check_mark:

PR #293, opened 2025-07-28
🧑‍💻(keycloak) do not require ssl for dev server

jbpenrath:
## Purpose

Set `sslRequired` to `none` for all realms in the keycloak dev service. By default this settings is set to `external` and only local IPs are granted to access to the server without SSL. But sometimes, with docker (desktop?) it appears the network configuration may keycloak to consider the accessing ip as not local.

We also took opportunity of this change to use the latest keycloak image version

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -190,7 +190,7 @@ services:
       - env.d/development/kc_postgresql
 
   keycloak:
-    image: quay.io/keycloak/keycloak:20.0.1
+    image: quay.io/keycloak/keycloak:26.3.2
     volumes:
       - ./docker/auth/realm.json:/opt/keycloak/data/import/realm.json
     command:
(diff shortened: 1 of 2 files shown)


compose.yaml
```diff
@@ -190,18 +190,17 @@ services:
...
       - ./docker/auth/realm.json:/opt/keycloak/data/import/realm.json
     command:
       - start-dev
       - --features=preview
       - --import-realm
       - --proxy=edge
-      - --hostname-url=http://localhost:8083
-      - --hostname-admin-url=http://localhost:8083/
+      - --hostname=http://localhost:8083
+      - --hostname-admin=http://localhost:8083/
```
lunika:
```suggestion
      - --http-enabled=true
```
`--hostname-admin` is not needed anymore

lunika (approved):

Issue #289, opened 2025-07-28
Version history


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

For now, a file can be uploaded once and then it is not possible to replace it or override it with a new upload.
Also, when the WOPI implementation (#2) will be merged, we will have an history for a same file.

We want to be able to manage history for a given file.

**Describe the solution you'd like**

Manage history of a given file, if they are modified by an external tool (WOPI) or if an upload replace an existing file.

robin-lecomte:
See https://github.com/suitenumerique/drive/issues/638

PanchoutNathan:
close a duplicate #638

PR #296, opened 2025-07-29
♻️(back) manage title uniqueness by generating new one if existing

lunika:
## Purpose

For now it wasn't possible to create twice an item with the same title in the same path and an exception was raised. We change it by incrementing a version number in the item title. It is still not possible to have twice the same title but the item is created with a different one.

## Proposal

- [x] ♻️(back) manage title uniqueness by generating new one if existing

## Note

This is a first iteration that does not need design. An other iteration will be made with #289

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -30,6 +30,8 @@ from django_ltree.managers import TreeManager, TreeQuerySet
 from django_ltree.models import TreeModel
 from timezone_field import TimeZoneField
 
+from core.utils.item_title import manage_unique_title as manage_unique_title_utils
+
 logger = getLogger(__name__)
 
 
@@ -418,20 +420,6 @@ class ItemQuerySet(TreeQuerySet):
         return self.filter(models.Q(link_reach=LinkReachChoices.PUBLIC))
 
 
-def _is_item_title_existing(queryset, title):
-    """Check if the title is unique in the same path."""
-    return (
-        queryset.filter(title=title)
-        .filter(
-            models.Q(
-                models.Q(deleted_at__isnull=True)
-                | models.Q(ancestors_deleted_at__isnull=True)
-            )
-        )
-        .exists()
-    )
-
-
 class ItemManager(TreeManager):
     """Custom manager for Item model overriding create_child method."""
 
@@ -461,19 +449,9 @@ class ItemManager(TreeManager):
                         )
                     }
                 )
-
-            if _is_item_title_existing(
-                self.children(parent.path),
-                kwargs.get("title"),
-            ):
-                raise ValidationError(
-                    {
-                        "title": ValidationError(
-                            _("title already exists in this folder."),
-                            code="item_create_child_title_already_exists",
-                        )
-                    }
-                )
+            kwargs["title"] = manage_unique_title_utils(
+                self.children(parent.path), kwargs.get("title")
+            )
 
         if not kwargs.get("id"):
             kwargs["id"] = str(uuid.uuid4())
@@ -645,9 +623,9 @@ class Item(TreeModel, BaseModel):
         """Generate a unique cache key for each item."""
         return f"item_{self.id!s}_nb_accesses"
 
-    def is_item_title_existing(self, title):
+    def manage_unique_title(self, title):
         """Check if the title is unique in the same path."""
-        return _is_item_title_existing(
+        return manage_unique_title_utils(
             self.siblings(),
             title,
         )
--- a/src/backend/core/utils/item_title.py
+++ b/src/backend/core/utils/item_title.py
@@ -1,0 +1,76 @@
+"""Module to manage item title uniqueness."""
+
+from os.path import splitext
+
+from django.db import models
+
+
+def _get_non_deleted_filter():
+    """Return a Q object for filtering non-deleted items."""
+    return models.Q(
+        models.Q(deleted_at__isnull=True) | models.Q(ancestors_deleted_at__isnull=True)
+    )
+
+
+def _is_item_title_existing(queryset, title):
+    """Check if the title is unique in the same path."""
+    return queryset.filter(title=title).filter(_get_non_deleted_filter()).exists()
+
+
+def _extract_number_from_title(title):
+    """Extract the numeric suffix from a title with the given base."""
+    base, _ = splitext(title)
+    try:
+        return int(base.split("_")[-1])
+    except (ValueError, IndexError):
+        return 0
+
+
+def _build_numbered_title_regex(base_title, ext):
+    """Build a regex pattern to match numbered versions of a title."""
+    escaped_base = base_title.replace("(", r"\(").replace(")", r"\)")
+    pattern = rf"^{escaped_base}_\d+"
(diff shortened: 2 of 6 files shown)


src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,76 @@
...
+
+from os.path import splitext
+
+from django.db import models
+
+
+def _get_non_deleted_filter():
+    """Return a Q object for filtering non-deleted items."""
+    return models.Q(
+        models.Q(deleted_at__isnull=True) | models.Q(ancestors_deleted_at__isnull=True)
```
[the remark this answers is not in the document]
lunika: This is wrong, the `OR` operator should be used.

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+def _is_item_title_existing(queryset, title):
+    """Check if the title is unique in the same path."""
+    return queryset.filter(title=title).filter(_get_non_deleted_filter()).exists()
+
+
+def _extract_number_from_title(title):
+    """Extract the numeric suffix from a title with the given base."""
+    base, _ = splitext(title)
+    try:
+        return int(base.split("_")[-1])
```
qbey:
```suggestion
        return int(base.rsplit("_", 1)[-1])
```

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+def _build_numbered_title_regex(base_title, ext):
+    """Build a regex pattern to match numbered versions of a title."""
+    escaped_base = re.escape(base_title)
+    pattern = rf"^{escaped_base}_\d+"
+    if ext:
+        escaped_ext = re.escape(ext)
+        pattern += rf"{escaped_ext}$"
+    else:
+        pattern += r"$"
+    return pattern
```
qbey:
```suggestion
def _build_numbered_title_regex(base_title, ext):
    """Build a regex pattern to match numbered versions of a title."""
    escaped_base = re.escape(base_title)
    escaped_ext = re.escape(ext) if ext else ""
    return rf"^{escaped_base}_\d+{escaped_ext}$"
```

src/backend/core/models.py
```diff
@@ -645,9 +623,9 @@ def get_nb_accesses_cache_key(self):
         """Generate a unique cache key for each item."""
         return f"item_{self.id!s}_nb_accesses"
 
-    def is_item_title_existing(self, title):
+    def manage_unique_title(self, title):
         """Check if the title is unique in the same path."""
-        return _is_item_title_existing(
+        return manage_unique_title_utils(
```
qbey: You may update the docstring 👼

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+def _get_non_deleted_filter():
+    """Return a Q object for filtering non-deleted items."""
+    return models.Q(
+        models.Q(deleted_at__isnull=True) | models.Q(ancestors_deleted_at__isnull=True)
+    )
+
+
+def _is_item_title_existing(queryset, title):
+    """Check if the title is unique in the same path."""
+    return queryset.filter(title=title).filter(_get_non_deleted_filter()).exists()
```
qbey: Are you sure you want to make a function for this? This seems used only once

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
+"""Module to manage item title uniqueness."""
+
+import re
+from os.path import splitext
+
+from django.db import models
+
+
+def _get_non_deleted_filter():
```
qbey: What happens when you "undelete" something then?

Ex:
 - file.txt is deleted
 - I add a new file.txt : filename is preserved
 - I undelete the first file (I don't know if possible yet)

src/backend/core/utils/item_title.py
qbey: I would write a custom queryset for the model for that : https://docs.djangoproject.com/en/5.2/topics/db/managers/#from-queryset

``` python
class CustomQuerySet(models.QuerySet):
    def filter_non_deleted(self, **kwargs):
        """Filter the non delete items"""
        return self.filter(
            models.Q(models.Q(deleted_at__isnull=True)
            | models.Q(ancestors_deleted_at__isnull=True),
            **kwargs,
        )

class MyModel(models.Model):
    objects = Manager.from_queryset(CustomQuerySet)()
```

So you can call `queryset.filter_non_deleted(title=title).exists()`

Be careful, the model, might already have a custom queryset and manager (for the tree) so you might inherit from them.

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    if not existing_titles:
+        return 1
+
+    # Extract numbers and find the maximum
+    max_number = 0
+    for title in existing_titles:
+        number = _extract_number_from_title(title)
+        max_number = max(max_number, number)
```
qbey:
```suggestion
    max_number = max(_extract_number_from_title(title) for title in existing_titles)
```

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+def manage_unique_title(queryset, title):
+    """Manage the unique title in the same path."""
+    # Return original title if it doesn't exist
+    if not _is_item_title_existing(queryset, title):
+        return title
+
+    # Handle duplicate by adding numeric suffix
+    base_title, ext = splitext(title)
+    next_number = _get_next_available_number(queryset, base_title, ext)
+    return f"{base_title}_{next_number}{ext}"
```
qbey, line 48: Question:
Do we want to format the next number? like it could have leading 0 (like "01", "02", ...) so at least the 100 first duplicated files are sortable by name (in the frontend I mean, this would not change anything on the backend side).

src/backend/core/utils/item_title.py
```diff
@@ -0,0 +1,77 @@
...
+        escaped_ext = re.escape(ext)
+        pattern += rf"{escaped_ext}$"
+    else:
+        pattern += r"$"
+    return pattern
+
+
+def _get_next_available_number(queryset, base_title, ext):
+    """Get the next available number for a duplicate title."""
+    title_regex = _build_numbered_title_regex(base_title, ext)
```
qbey: `_build_numbered_title_regex` is called only once. I feel like you want to split into a lot of small functions, but I find it hard to read. Or at least you must unit test every function ;)

src/backend/core/utils/item_title.py
lunika: Yes it's possible, and you will have two files with the same name. And we decided that it's ok to have this behavior.

src/backend/core/utils/item_title.py
lunika: Testing private function is a bad pattern you know :-)

src/backend/core/utils/item_title.py
lunika: Yes, it's a good idea

PR #299, opened 2025-07-31
Allow theme customization using a configuration file

### LINKED ISSUE #273: allow theme customnization using a configuration file
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

We need to configure the front application. As the front application is a static artifact, we can't introduce variables. An existing config endpoint is used for some information like configuring sentry but the list of settings used can grow really fast. We would like instead a configuration file in JSON dedicated to the frontend theme.

**Describe the solution you'd like**

The first thing we need is to build dynamically the links in the footer. 
We can introduce this config file based on this PR made on docs: https://github.com/suitenumerique/drive/pull/256

lunika:
## Purpose

We want to customize the theme by using a configuration file. This
configuration file path can be defined using the settings
THEME_CUSTOMIZATION_FILE_PATH. If this file does not exists or is an
invalid json, an empty json object will be added in the config endpoint.

In the front application, we can use this new theme_customization present in the config endpoint to first configure the footer.

**:warning: :warning:  The PR https://github.com/suitenumerique/ui-kit/pull/99 must be merged first and a new release made in order to end this PR.
Also, the cunningham-react package must be updated to at least version 3.2.1**

## Proposal

- [x] ✨(back) allow theme customnization using a configuration file
- [x] ✨(helm) allow to load custom theme file in a configMap
- [x] ✨(front) use theme_customization to configure the footer

Fix #273


src/backend/drive/configuration/theme/default.json
```diff
@@ -0,0 +1,74 @@
...
+            "label": "DINUM",
+            "href": "https://www.numerique.gouv.fr/dinum/"
+          },
+          {
+            "label": "ANCT",
+            "href": "https://anct.gouv.fr/"
+          }
+        ],
+        "bottomInformation": {
+          "label": "Unless otherwise stated, all content on this site is under",
```
NathanVss: Having translation in such a file along with the existing translations file bother me a bit, although I don't have any solutions to bring as in this case it provides a runtime customization. What's your feeling on this?

src/frontend/apps/drive/src/features/drivers/types.ts
```diff
@@ -124,6 +124,40 @@ export type User = {
...
+
+export interface ContentType {
+  logo?: Logo;
+  externalLinks?: Link[];
+  legalLinks?: Link[];
+  bottomInformation?: BottomInformation;
+}
+
+export interface FooterType {
+  default: ContentType;
```
NathanVss: Once the PR is merge on ui-kit we could use the `FooterProps` type here ( maybe using `Partial<FooterProps>` ) instead of duplicating the types

src/frontend/apps/drive/src/pages/index.tsx
```diff
@@ -27,9 +27,12 @@ import { Button } from "@openfun/cunningham-react";
 import { Feedback } from "@/features/feedback/Feedback";
 import { useConfig } from "@/features/config/ConfigProvider";
 export default function HomePage() {
-  const { t } = useTranslation();
+  const { i18n, t } = useTranslation();
   const { user } = useAuth();
   const { config } = useConfig();
+  const footerJson = config?.theme_customization?.footer;
+  const resolvedLanguage = i18n.resolvedLanguage;
```
NathanVss: We will prefer `language` as `resolvedLanguage` could be undefined and I think refers to the default browser language, but here `language` is set to the user language if defined

PR #307, opened 2025-08-18
🔖(minor) bump release version 0.2.0

NathanVss:
Added

- ✨(front) Add public workspaces
- ✨(front) Add 401, 403 pages
- ✨(front) Add redirect after login logic

Changed

- ♻️(back) manage title uniqueness by generating new one if existing #296
- 🧑‍💻(docker) handle frontend development images with docker compose #298
- 🔧(project) change env.d system by using local files #298
- Bump ui-kit to remove the usage of base64 font in CSS #302

Fixed

- 🐛(front) set the correct move icon
- 🐛(nginx) add trash route
- 💬(front) update feedback texts

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "main",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "private": true,
   "workspaces": [
     "apps/*",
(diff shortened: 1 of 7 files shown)


src/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "main",
-  "version": "0.1.1",
+  "version": "0.2.0",
```
lunika, line 3: In the `src/frontend` directory you have other packages to update.

NathanVss: Those have their own release lifecycle, to avoid version gaps without changes.

PR #309, opened 2025-08-19
🔧(nginx) fix trash route

NathanVss:
The previous fix was missing to explorer part.


lunika (commented):
Can you add a Fix in the changelog ?

NathanVss:
> Can you add a Fix in the changelog ?

Here we go ;)

lunika (approved):

Issue #279, opened 2025-07-25
Search an item by their title


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

A first step to introduce search in the Drive project is to search for an item with its title

**Describe the solution you'd like**

Implement filter in backend to search items by their title (case insensitive, etc) and make the front design associated

lunika:
Managed in #312 and #326

PR #312, opened 2025-08-19
✨(backend) search endpoint for ItemViewSet

lunika:
## Purpose

We need a new endpoint for the search feature. Using existing filters on the list or children endpoint does not match what we want. We want to search for all accessible items for the current user.

## Proposal

- [x] ✨(backend) search endpoint for ItemViewSet

Part of #279

### DIFF
--- a/src/backend/core/tests/items/test_api_items_search.py
+++ b/src/backend/core/tests/items/test_api_items_search.py
@@ -1,0 +1,556 @@
+"""Tests for the Item viewset search method."""
+
+import random
+
+import pytest
+from rest_framework.test import APIClient
+
+from core import factories, models
+
+pytestmark = pytest.mark.django_db
+
+
+def test_api_items_search_anonymous():
+    """Anonymous users should not be able to search for items."""
+    client = APIClient()
+    response = client.get("/api/v1.0/items/search/")
+    assert response.status_code == 401
+
+
+def test_api_items_search_authenticated_without_filters():
+    """
+    Authenticated users should be able to search for items.
+    Without filters, all the items the user has access to should be returned.
+    """
+    user = factories.UserFactory()
+    client = APIClient()
+    client.force_login(user)
+
+    top_parent = factories.ItemFactory(
+        title="Item 1", type=models.ItemTypeChoices.FOLDER
+    )
+    top_parent_access = factories.UserItemAccessFactory(item=top_parent, user=user)
+    parent = factories.ItemFactory(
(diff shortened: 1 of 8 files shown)


src/backend/core/tests/items/test_api_items_search.py
```diff
@@ -0,0 +1,260 @@
...
+
+import random
+
+import pytest
+from faker import Faker
+from rest_framework.test import APIClient
+
+from core import factories, models
+
+fake = Faker()
```
NathanVss, line 11: Not used?

NathanVss: Not used?

src/backend/core/tests/items/test_api_items_search.py
```diff
@@ -0,0 +1,490 @@
...
+    top_parent = factories.ItemFactory(
+        title="Item 1", users=[user], type=models.ItemTypeChoices.FOLDER
+    )
+    non_accessible_top_parent = factories.ItemFactory(
+        title="Item 3", type=models.ItemTypeChoices.FOLDER
+    )
+    for title in titles:
+        parent = (
+            factories.ItemFactory(type=models.ItemTypeChoices.FOLDER, parent=top_parent)
+            if random.choice([True, False])
```
NathanVss, line 380: Randomness in test can cause unconsistent behavior and unpredictable subsequent exection, maybe can another test case to always test both situations?

src/backend/core/tests/items/test_api_items_search.py
```diff
@@ -0,0 +1,490 @@
+"""Tests for the Item viewset search method."""
```
NathanVss, line 1: Pagination does not seem to be tested, nor invalid parameters ( `type=nothing` for instance ), WDYT?

src/backend/core/tests/items/test_api_items_search.py
lunika: This is what I want. If it's inconsistent, then we have a problem.

src/backend/core/tests/items/test_api_items_search.py
lunika: Pagination is not tested because it is tested in DRF, I don't have to test it.
Ok for adding an invalid paramters.

src/backend/core/tests/items/test_api_items_search.py
lunika: Yes, removed. Thanks

PR #317, opened 2025-08-20
✅(front) add e2e tests

NathanVss:
## How-to

Just run

```
make run-tests-e2e
```

It will:
- Start the backend with the e2e db
- Clear the DB
- Apply migrations
- Run the e2e tests on it

You can also pass parameters like so:

```
make run-tests-e2e -- --project chromium --headed
```

After the run the backend will still run on the e2e DB without clearing, this is on purpose in case you need to interact with the state of the app after the tests have run or failed, in order to update them.

## Todo

- [x] Setup playwright
- [x] Make authent
- [x] Write first test
- [x] Setup CI
- [x] DB e2e en local

### DIFF
--- a/.github/workflows/drive-frontend.yml
+++ b/.github/workflows/drive-frontend.yml
@@ -34,3 +34,55 @@ jobs:
 
       - name: Check linting
         run: cd src/frontend/ && yarn lint
+
+  test-e2e:
+    runs-on: ubuntu-latest
+    needs: install-front
+    strategy:
+      matrix:
+        browser:
+          - chromium
+          - webkit
+          - firefox
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+
+      - name: Restore the frontend cache
+        uses: actions/cache@v4
+        with:
+          path: "src/frontend/**/node_modules"
+          key: front-node_modules-${{ hashFiles('src/frontend/**/yarn.lock') }}
+          fail-on-cache-miss: true
+
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps ${{matrix.browser}}
+
+      - name: Start Docker services
+        run: |
+          make bootstrap-e2e
+
+      - name: Start frontend
+        run: |
+          cd src/frontend && yarn dev &
+
+      - name: Wait for Keycloak to be ready
+        run: |
+          timeout 30 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8083/)" != "302" ]]; do echo "Waiting for Keycloak..." && sleep 2; done' && echo "Keycloak is ready!"
+
+      - name: Run e2e tests
+        run: |
+          cd src/frontend/apps/e2e
+          yarn test --project=${{ matrix.browser }}
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: report-${{ matrix.browser }}
+          path: src/frontend/apps/e2e/report/
+          retention-days: 7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to
 
 - ✨(back) allow theme customization using a configuration file #299
 - ✨(front) use theme_customization to configure the footer #299
+- ✅(front) add e2e testing #317
 
 ### Fixed
 
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ COMPOSE_RUN_CROWDIN = $(COMPOSE_RUN) crowdin crowdin
 # -- Backend
 MANAGE              = $(COMPOSE_RUN_APP) python manage.py
 MAIL_YARN           = $(COMPOSE_RUN) -w /app/src/mail node yarn
+PSQL_E2E = ./bin/postgres_e2e
 
 # -- Frontend
 PATH_FRONT          = ./src/frontend
@@ -105,6 +106,7 @@ build-frontend: ## build the frontend container
 .PHONY: build-frontend-development
 
 down: ## stop and remove containers, networks, images, and volumes
+	rm -rf data/postgresql.*
 	@$(COMPOSE) down
 .PHONY: down
 
@@ -116,6 +118,33 @@ run-backend: ## start the backend container
 	@$(COMPOSE) up --force-recreate -d celery-dev
 	@$(COMPOSE) up --force-recreate -d nginx
 .PHONY: run-backend
+
+bootstrap-e2e: ## bootstrap the backend container for e2e tests, without frontend
+bootstrap-e2e: \
+	data/media \
+	data/static \
+	create-env-local-files \
+	build-backend \
+	back-i18n-compile \
+	run-backend-e2e
+.PHONY: bootstrap-e2e
+
+clear-db-e2e: ## quickly clears the database for e2e tests, used in the e2e tests
+	$(PSQL_E2E) -c "$$(cat bin/clear_db_e2e.sql)"
+.PHONY: clear-db-e2e
+
+run-backend-e2e: ## start the backend container for e2e tests, always remove the postgresql.e2e volume first
+	@$(MAKE) stop
+	rm -rf data/postgresql.e2e
+	@ENV_OVERRIDE=e2e $(MAKE) run-backend
+	@ENV_OVERRIDE=e2e $(MAKE) migrate
+.PHONY: run-backend-e2e
+
+run-tests-e2e: ## run the e2e tests, example: make run-tests-e2e -- --project chromium --headed
+	@$(MAKE) run-backend-e2e	
+	@args="$(filter-out $@,$(MAKECMDGOALS))" && \
+	cd src/frontend/apps/e2e && yarn test $${args:-${1}}
+.PHONY: run-tests-e2e
 
 run: ## start the development server and frontend development
 run: 
--- a/compose.yaml
+++ b/compose.yaml
@@ -12,7 +12,9 @@ services:
       retries: 300
     env_file:
       - env.d/development/postgresql
-      - env.d/development/postgresql.local
+      - env.d/development/postgresql.${ENV_OVERRIDE:-local}
+    volumes:
+      - ./data/postgresql.${ENV_OVERRIDE:-local}:/var/lib/postgresql/data
 
   redis:
     image: redis:5
@@ -71,7 +73,7 @@ services:
       - env.d/development/common
       - env.d/development/common.local
       - env.d/development/postgresql
-      - env.d/development/postgresql.local
+      - env.d/development/postgresql.${ENV_OVERRIDE:-local}
     ports:
       - "8071:8000"
     volumes:
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -60,6 +60,7 @@ class Base(Configuration):
     """
 
     DEBUG = False
+    LOAD_E2E_URLS = False
     USE_SWAGGER = False
 
     API_VERSION = "v1.0"
@@ -737,6 +738,7 @@ class Development(Base):
     CORS_ALLOW_ALL_ORIGINS = True
     CSRF_TRUSTED_ORIGINS = ["http://localhost:8072", "http://localhost:3000"]
     DEBUG = True
+    LOAD_E2E_URLS = True
 
     SESSION_COOKIE_NAME = "drive_sessionid"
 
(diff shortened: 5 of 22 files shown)


compose.yaml
```diff
@@ -12,7 +12,9 @@ services:
       retries: 300
     env_file:
       - env.d/development/postgresql
-      - env.d/development/postgresql.local
+      - env.d/development/postgresql.${ENV_OVERRIDE:-local}
+    volumes:
+      - ./data/postgresql.${ENV_OVERRIDE:-local}:/var/lib/postgresql/data
```
lunika, line 17: Why do you need this ?

Makefile
```diff
@@ -117,6 +117,19 @@ run-backend: ## start the backend container
 	@$(COMPOSE) up --force-recreate -d nginx
 .PHONY: run-backend
 
+run-backend-e2e: ## start the backend container for e2e tests, always remove the postgresql.e2e volume first
+	rm -rf data/postgresql.e2e
+	@ENV_OVERRIDE=e2e $(MAKE) run-backend
+	@ENV_OVERRIDE=e2e $(MAKE) migrate
+.PHONY: run-backend-e2e
+
+run-tests-e2e: ## run the e2e tests
```
lunika: You should add information on how to use it. Can be a complete example.

.github/workflows/drive-frontend.yml
```diff
@@ -34,3 +34,62 @@ jobs:
...
+
+      - name: Start Docker services
+        run: |
+          make data/media
+          make data/static
+          make create-env-local-files
+          make build-backend
+          make migrate
+          make back-i18n-compile
+          make run-backend
```
lunika: Create a `bootstrap-e2e` make task ?

compose.yaml
NathanVss: Because `POSTGRES_DB` is used only the first time postgresql container is started

.github/workflows/drive-frontend.yml
NathanVss: Yep

compose.yaml
NathanVss: That means if we change POSTGRES_DB on subsequent run it would still use the previous one. By switching volume it really switch the DB by preserving both without data loss

lunika: And if the variables change, it is not recreated ?

NathanVss: Nope:

```
POSTGRES_DB
This optional environment variable can be used to define a different name for the default database that is created when the image is first started. If it is not specified, then the value of POSTGRES_USER will be used.
```

From https://hub.docker.com/_/postgres#postgres_db

So switching volume is the cleanest solution I found out

Makefile
```diff
@@ -117,6 +117,30 @@ run-backend: ## start the backend container
...
+	data/static \
+	create-env-local-files \
+	build-backend \
+	migrate \
+	back-i18n-compile \
+	run-backend
+.PHONY: bootstrap-e2e
+
+run-backend-e2e: ## start the backend container for e2e tests, always remove the postgresql.e2e volume first
+	rm -rf data/postgresql.e2e
```
lunika, line 138: Can you also add in the `down` task `rm -rf data/postgesql.*` ?

src/frontend/apps/e2e/playwright.config.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  maxFailures: process.env.CI ? 3 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 3 : undefined,
```
PanchoutNathan: I'm for doing it sequentially.

```suggestion
  workers: process.env.CI ? 1 : undefined,
```

src/frontend/apps/e2e/playwright.config.ts
```diff
@@ -0,0 +1,84 @@
...
+      use: {
+        ...devices["Desktop Firefox"],
+        locale: "en-US",
+        timezoneId: "Europe/Paris",
+        launchOptions: {
+          firefoxUserPrefs: {
+            "dom.events.asyncClipboard.readText": true,
+            "dom.events.testing.asyncClipboard": true,
+          },
+        },
```
PanchoutNathan: In my opinion it's overkill to do all 3, especially at first.

Makefile
PanchoutNathan: And I honestly think that this should not be done there, but rather in the playwright config file.

src/frontend/apps/e2e/__tests__/app-drive/utils-common.ts
```diff
@@ -0,0 +1,57 @@
+import { expect, Page } from "@playwright/test";
+import { exec } from "child_process";
+
+const ROOT_PATH = "../../../..";
+const CLEAR_DB_TARGET = "clear-db-e2e";
```
PanchoutNathan, line 5: perhaps a more explicit name, because it is a command.

.github/workflows/drive-frontend.yml
```diff
@@ -34,3 +34,55 @@ jobs:
...
+
+      - name: Restore the frontend cache
+        uses: actions/cache@v4
+        with:
+          path: "src/frontend/**/node_modules"
+          key: front-node_modules-${{ hashFiles('src/frontend/**/yarn.lock') }}
+          fail-on-cache-miss: true
+
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
```
lunika: You don't have to specify the browser to install ? Refering to the documentation all default browsers will be installed and you will use only one. Also if you inspect the logs of this step, you will see tht multiple browsers are installed.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to
 
 - ✨(back) allow theme customnization using a configuration file #299
 - ✨(front) use theme_customization to configure the footer #299
+- ✅(front) add e2e testing #317
```
lunika, line 17: Once rebased, move it to the unreleased section. A release has been made.

Makefile
```diff
@@ -48,6 +48,7 @@ COMPOSE_RUN_CROWDIN = $(COMPOSE_RUN) crowdin crowdin
 # -- Backend
 MANAGE              = $(COMPOSE_RUN_APP) python manage.py
 MAIL_YARN           = $(COMPOSE_RUN) -w /app/src/mail node yarn
+PSQL = $(COMPOSE) exec -T postgresql psql -U $(shell grep POSTGRES_USER env.d/development/postgresql.e2e | cut -d'=' -f2) -d $(shell grep POSTGRES_DB env.d/development/postgresql.e2e | cut -d'=' -f2) -v ON_ERROR_STOP=1
```
lunika: This is the kind of command that should be in a shell script rather than in a make variable. Like we have `bin/pytest` you can create a bash script in `bin/postgres`

src/backend/demo/urls.py
```diff
@@ -0,0 +1,26 @@
...
+user_auth_router = DefaultRouter()
+user_auth_router.register(
+    "user-auth",
+    viewsets.UserAuthViewSet,
+    basename="user-auth",
+)
+
+urlpatterns = [
+    path(
+        f"api/{settings.API_VERSION}/demo/",
```
lunika: IMO this should not be in a `demo` app but in a `e2e` app. I'm probably responsible for the confusion when I mentioned the demo app.

src/backend/core/tests/test_api_demo.py
```diff
@@ -0,0 +1,74 @@
+"""
+Test demo API endpoints.
```
lunika: This tests should be in the `demo` app, not in the core one.

src/backend/demo/viewsets.py
```diff
@@ -0,0 +1,39 @@
...
+
+    permission_classes = [AllowAny]
+    authentication_classes = []
+
+    def create(self, request):
+        """
+        POST /api/v1.0/demo/user-auth/
+        Create a user with the given email if it doesn't exist and log them in
+        """
+        email = request.data.get("email")
```
lunika: I think you should use at least a validator, or better, a serializer.

src/backend/drive/settings.py
```diff
@@ -308,6 +309,7 @@ class Base(Configuration):
     # Django applications from the highest priority to the lowest
     INSTALLED_APPS = [
         "core",
+        "demo",
```
lunika: Should be added only in the `Test` and `Development` config. Not in the base one.

src/backend/drive/settings.py
```diff
@@ -737,6 +739,7 @@ class Development(Base):
     CORS_ALLOW_ALL_ORIGINS = True
     CSRF_TRUSTED_ORIGINS = ["http://localhost:8072", "http://localhost:3000"]
     DEBUG = True
+    LOAD_DEMO_URLS = True
```
lunika: This settings is useless IMO, you should load the urls only if `DEBUG` is True.

NathanVss: it makes the tests fails if we set DEBUG true in them

.github/workflows/drive-frontend.yml
NathanVss: It takes the same amount of time (weird ok)

src/backend/drive/settings.py
lunika: Not sure to understand. We want to add the urls only if DEBUG is True. So there is no need to change this settings from the tests.

NathanVss: -> We discuss during a call about this, we decided to keep `LOAD_DEMO_URLS` because `DEBUG` loads debug toolbars that make the tests fail.

bin/postgres
```diff
@@ -0,0 +1,12 @@
...
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+# Get database credentials from environment file
+ENV_FILE="${REPO_DIR}/env.d/development/postgresql.e2e"
+POSTGRES_USER=$(grep POSTGRES_USER "$ENV_FILE" | cut -d'=' -f2)
+POSTGRES_DB=$(grep POSTGRES_DB "$ENV_FILE" | cut -d'=' -f2)
+
+# Execute PostgreSQL command (run as postgres user, not host user)
+_docker_compose exec -T postgresql psql -U "$POSTGRES_USER" -d "$POSTGRES_DB" -v ON_ERROR_STOP=1 "$@"
```
lunika: Why not using `ENV_OVERRIDE=e2e` here ?

NathanVss: We could but I would not change anything as far as I know?

lunika: You are taking the variables value from the e2e environment variable file, so you will connect to the e2e database but if it's not the good volume mounted, then you can have trouble I think.

NathanVss: Yep so we can leave this like that

lunika: Ok so i'm misunderstading what you are doing. For me you are mixing both `local` and `e2e` volumes and this will lead to error in the database.

lunika: First run 

```
$ make clear-db-e2e 
./bin/postgres -c "$(cat bin/clear_db_e2e.sql)"
🐳(compose) project, file: '/home/manu/dev/dinum/drive/compose.yaml'
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "drive_e2e" does not exist
make: *** [Makefile:134: clear-db-e2e] Error 2
```

After a `make bootsrap`

```
$ make clear-db-e2e 
./bin/postgres -c "$(cat bin/clear_db_e2e.sql)"
🐳(compose) project, file: '/home/manu/dev/dinum/drive/compose.yaml'
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "drive_e2e" does not exist
make: *** [Makefile:134: clear-db-e2e] Error 2
```

The only case this will work it is when you have run `make run-e2e-backend` first. So something is missing to ensure this command is ran in a good context.

lunika: also, you should rename the file in `postgres-e2e`, you can use only in this context.

Makefile
```diff
@@ -117,6 +119,34 @@ run-backend: ## start the backend container
 	@$(COMPOSE) up --force-recreate -d nginx
 .PHONY: run-backend
 
+bootstrap-e2e: ## bootstrap the backend container for e2e tests, without frontend
+bootstrap-e2e: \
+	data/media \
+	data/static \
+	create-env-local-files \
+	build-backend \
+	migrate \
```
lunika: Useless, this will migrate the `local` database, not the e2e. The e2e migration is made in the `run-backend-e2e`

bin/postgres
NathanVss: It is only intended to be run from e2e tests that's why.

Renaming the file to `postgres-e2e` would be enough for you?

lunika (approved):

PR #319, opened 2025-08-21
⚡️(ci) use setup-python cache option

lunika:
## Purpose

The setup-python action is able to cache the dependencies and reuse this
cache while the pyproject file has not changed. It is easy to setup,
juste the package manager used has to be declared in the cache settings.

## Proposal

- [x] ⚡️(ci) use setup-python cache option


NathanVss (approved):
This rocks :)

PR #321, opened 2025-08-25
🔧(back) customize cache config

lunika:
## Purpose

We want to split the cache config between the app cache and the session cache. In the app cache, the default one, we allow to configure a prefix. By default this prefix is a fixed string so the cache will be never revoked because it is changing but it allow every instance to implement its own strategy like prefixing the keyx cache with a timestamp.
To not impact session, the session cache is splitted in the settings.

## Proposal

- [x] 🔧(back) customize cache config


NathanVss (approved):
I was thinking about doing it too, great you did it

PR #328, opened 2025-09-01
🐛(front) fix the content when opening the right panel

PanchoutNathan:
## Purpose

when we opened the right panel, we did not arrive at the right item

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
@@ -174,14 +174,13 @@ export const GlobalExplorerProvider = ({
   }, [rightPanelForcedItem]);
 
   /**
-   * When the right panel is open, we need to force the new item to be displayed in the right panel.
+   * We need to force the current folder to be displayed in the right panel.
    */
   useEffect(() => {
-    if (item && rightPanelOpen) {
-      setRightPanelForcedItem(item);
+    if (rightPanelOpen) {
       setSelectedItems([]);
     }
-  }, [item, rightPanelOpen]);
+  }, [rightPanelOpen]);
 
   const { dropZone } = useUploadZone({ item: item! });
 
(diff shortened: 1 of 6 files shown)


src/frontend/apps/e2e/__tests__/app-drive/item/right-content-info.spec.ts
```diff
@@ -0,0 +1,20 @@
+import test, { expect } from "@playwright/test";
+import { clearDb, login } from "../utils-common";
+import { clickOnRowItemActions, getRowItem } from "../utils-embeded-grid";
+
+test("Check that the right content is displayed correctly", async ({
```
NathanVss, line 5: would be great to test with a file also

src/frontend/apps/e2e/__tests__/app-drive/utils-embeded-grid.ts
```diff
@@ -0,0 +1,35 @@
+import { expect, Page } from "@playwright/test";
+
```
NathanVss: typo in filename

src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
```diff
@@ -172,11 +172,10 @@ export const GlobalExplorerProvider = ({
    * When the right panel is open, we need to force the new item to be displayed in the right panel.
```
NathanVss: "we need to force the **current folder** to be displayed in the right panel." ?

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
```diff
@@ -235,6 +235,7 @@ export const EmbeddedExplorerGrid = (props: EmbeddedExplorerGridProps) => {
                 const isOvered = !!overedItemIds[row.original.id];
                 return (
                   <tr
+                    role="row"
```
NathanVss: Claude tells: 

```
Redundant declaration: Adding role="row" to a <tr> element is redundant because:
<tr> elements automatically have the "row" role
Screen readers and other assistive technologies already understand that <tr> elements represent table rows
The table structure is already properly semantic with <table>, <thead>, <tbody>, and <tr> elements
```

Why did you put the role ?

NathanVss (approved):

PR #329, opened 2025-09-02
🐛(renovate) fix renovate configuration

lunika:
## Purpose

- schedule at 7am the monday
- avoid wiping manual fixes unless needed
- only create during schedule

## Proposal

- [x] 🐛(renovate) fix renovate configuration


AntoLC (approved):
Schedule does not seem follow on Docs.

It could be updated as well as suggested by @lebaudantoine https://github.com/numerique-gouv/renovate-configuration/blob/main/default.json

lunika:
Managed in https://github.com/numerique-gouv/renovate-configuration/pull/1

PR #334, opened 2025-09-05
♻️(tilt) use helm dev-backend chart

lunika:
## Purpose

Remove usage of bitname charts and use our own dev-backend charts instead.

## Proposal

- [x] ♻️(tilt) use hem dev-backend chart


rouja (approved):
LGTM

PR #336, opened 2025-09-09
🐛(back) encode white spaces in item url

### LINKED ISSUE #283: Url containing space are not properly handled
## Bug Report

**Problematic behavior**

We we copy a url of a file the white spaces are not replaced with "%20" codes.

https://fichiers.numerique.gouv.fr/media/item/7fd535d6-45f7-423e-8a2d-60ef3aazmkla/mon fichier final.pdf

**Expected behavior/code**
We should export url with "%20" codes.

NathanVss:
The white space were not encoded, so when being posted in chats the white space was breaking urls. Fixes #283


src/backend/core/api/serializers.py
```diff
@@ -221,7 +222,8 @@ def get_url(self, item):
         ):
             return None
 
-        return f"{settings.MEDIA_BASE_URL}{settings.MEDIA_URL}{item.file_key}"
+        url = f"{settings.MEDIA_BASE_URL}{settings.MEDIA_URL}{quote(item.file_key)}"
+        return url
```
```suggestion
        return f"{settings.MEDIA_BASE_URL}{settings.MEDIA_URL}{quote(item.file_key)}"
```

src/backend/core/tests/items/test_api_items_retrieve.py
```diff
@@ -1148,6 +1148,62 @@ def test_api_items_retrieve_file_with_url_property(upload_state):
     }
 
 
+def test_api_items_retrieve_file_with_url_property_with_spaces():
+    """
+    The `url` property should not be none if the item is not pending.
```
lunika: This description is not the good one.

lunika (approved):

PR #343, opened 2025-09-12
:sparkles:(frontend) Make the footer visibility configurable

Alevale:
## Purpose

This allows users to configure if the footer should be shown or not.

## Proposal

Through the `FRONTEND_SHOW_FOOTER` variable you're able to show or not the la suite footer in the main page


lunika:
We made a modification in #299 allowing to configure the footer. You can change all the links and labels. Is it still needed for you to hide the footer with this configuration feature ?

NathanVss (commented):
Sorry for the latency, could you instead rely on this file ? https://github.com/suitenumerique/drive/blob/main/src/backend/drive/configuration/theme/default.json

using `footer.show` as a bool

Thank you for your PR 🙏

PR #345, opened 2025-09-17
♻️(backend) use PUT presigned-url to upload files

lunika:
## Purpose

The file upload to the object storage backend was made using a presigned
  post. But not many object storage solution implement this method. They
  all used presigned url combined with the put_object method. We choose to
  change to this method to be more widely compatible.

## Proposal

Description...

- [x] ♻️(backend) use PUT presigned-url to upload files
- [x] ♻️(helm) adapt tilt deployment to correctly signe presigned urls
- [x] 💩(front) use PUT method to upload files

### DIFF
--- a/src/backend/core/api/utils.py
+++ b/src/backend/core/api/utils.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from django.conf import settings
 from django.core.files.storage import default_storage
 
+import boto3
 import botocore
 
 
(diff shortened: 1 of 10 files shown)


src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -253,29 +253,44 @@ export class StandardDriver extends Driver {
...
+    //   );
+    // }
+
+    // await uploadFile(urlObject.toString(), formData, (progress) => {
+    //   progressHandler?.(progress);
+    // });
+
+    // await fetchAPI(`items/${item.id}/upload-ended/`, {
+    //   method: "POST",
+    // });
```
sampaccoud: clean-up?

src/backend/core/api/utils.py
```diff
@@ -101,17 +102,24 @@ def generate_upload_policy(item):
     # Generate a unique key for the item
     key = f"{item.key_base}/{item.filename}"
 
+    if settings.AWS_S3_DOMAIN_REPLACE:
```
sampaccoud, line 112: I think you need to amend the settings documentation page. I will be interested by the explanation about this setting :stuck_out_tongue:

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
lunika: Done

src/backend/core/api/utils.py
lunika: Done

lunika:
@coderabbitai full review

src/backend/core/api/utils.py
qbey: I don't like the setting's name ^^' why not simply "AWS_S3_FRONTEND_DOMAIN" or "AWS_S3_DOMAIN_FOR_FRONTEND"... 

Anyway, thanks for the really helpful comment :)

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -253,27 +253,7 @@ export class StandardDriver extends Driver {
...
-    Object.entries(fields).forEach(([key, value]) => {
-      formData.append(key, value);
-    });
-    // IMPORTANT !!! The Content-Type must be BEFORE file, otherwise it will be ignored by
-    // Scaleway object storage. Order matters !!!
-    formData.append("Content-Type", file.type);
-    formData.append("file", file);
-
-    let urlObject = new URL(url);
-    if (process.env["NEXT_PUBLIC_S3_DOMAIN_REPLACE"]) {
```
qbey, line 269: You should also clean `NEXT_PUBLIC_S3_DOMAIN_REPLACE` from env files, Dockerfile, etc. while it seems not used anymore?

Makefile
```diff
@@ -107,8 +107,8 @@ build-frontend: ## build the frontend container
 .PHONY: build-frontend-development
 
 down: ## stop and remove containers, networks, images, and volumes
-	rm -rf data/postgresql.*
 	@$(COMPOSE) down
+	rm -rf data/postgresql.*
```
NathanVss, line 111: Maybe shouldn't be in a commit about frontend

NathanVss (approved):

PR #349, opened 2025-09-23
✨(backend) force content-disposition header during upload

lunika:
## Purpose

For the files without the preview feature, we want to force the
content-disposition to attachment. We guess the mimetype based on the
file extension and if it is not in the previewable list, we set the
ContentDisposition to the upload policy. When we will fetch the url,
this header will be added.

## Proposal

- [x] ✨(backend) force content-disposition header
- [x] ✨(frontend) adapt upload with policy given by the backend

### DIFF
--- a/src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
+++ b/src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
@@ -253,7 +253,7 @@ export class StandardDriver extends Driver {
       throw new Error("No policy found");
     }
 
-    await uploadFile(item.policy, file, (progress) => {
+    await uploadFile(item.policy.url, item.policy.headers, file, (progress) => {
       progressHandler?.(progress);
     });
 
@@ -301,19 +301,24 @@ const jsonToItem = (data: any): Item => {
 
 /**
  * Upload a file, using XHR so we can report on progress through a handler.
- * @param url The URL to POST the file to.
- * @param formData The multi-part request form data body to send (includes the file).
+ * @param url The url to use to upload the file.
+ * @param headers The headers to use to upload the file.
+ * @param file The file to upload.
  * @param progressHandler A handler that receives progress updates as a single integer `0 <= x <= 100`.
  */
 export const uploadFile = (
   url: string,
+  headers: Record<string, string>,
   file: File,
   progressHandler: (progress: number) => void
 ) =>
   new Promise((resolve, reject) => {
+
     const xhr = new XMLHttpRequest();
     xhr.open("PUT", url);
-    xhr.setRequestHeader("X-amz-acl", "private");
+    Object.entries(headers).forEach(([key, value]) => {
+      xhr.setRequestHeader(key, value);
+    });
     xhr.setRequestHeader("Content-Type", file.type);
 
     xhr.addEventListener("error", reject);
--- a/src/frontend/apps/drive/src/features/drivers/types.ts
+++ b/src/frontend/apps/drive/src/features/drivers/types.ts
@@ -70,7 +70,10 @@ export type Item = {
     update: boolean;
     upload_ended: boolean;
   };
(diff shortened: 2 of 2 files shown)


src/backend/core/api/utils.py
```diff
@@ -126,8 +144,28 @@ def generate_upload_policy(item):
...
+            previewable_mime_type
+        ):
+            is_previewable = True
+            break
+
+        if mime_type == previewable_mime_type:
+            is_previewable = True
+            break
+
+    return is_previewable
```
qbey: What happened here? XD

At least:

```suggestion
def _is_previewable(mime_type):
    """Check if a mime type is previewable."""
    for allowed in settings.ITEM_PREVIEWABLE_MIME_TYPES:
        if allowed.endswith("/"):
            if mime_type.startswith(allowed):
                return True
        elif mime_type == allowed:
            return True
    return False
```

ooooor, consider using a regex? ^^ Faster, less readable....

```
re.compile(
        "^(?:"
        + "|".join(
            f"{re.escape(mt)}.*" if mt.endswith("/") else re.escape(mt)
            for mt in previewable_types
        )
        + ")$"
    )
```
(generated via AI, like your code)

src/backend/drive/settings.py
```diff
@@ -171,75 +171,16 @@ class Base(Configuration):
...
-        "application/x-csh",
-        "application/x-ksh",
-        "application/x-ms-regedit",
-        "application/x-msdownload",
-        "application/xml",
-        "image/svg+xml",
-    ]
+    ITEM_PREVIEWABLE_MIME_TYPES = values.ListValue(
+        [
+            "image/",
```
qbey, line 176: For "beauty"'s sake I would have liked this to be `image/*` but don't bother making me happy.

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -301,19 +301,23 @@ const jsonToItem = (data: any): Item => {
 
 /**
  * Upload a file, using XHR so we can report on progress through a handler.
- * @param url The URL to POST the file to.
- * @param formData The multi-part request form data body to send (includes the file).
+ * @param policy The policy (url and headers) to use to upload the file.
```
qbey: Based on the following lines, the parameter is still named `url` :) Isn't it a bit misleading to group two parameters under the same `@param` definition?

src/backend/core/api/utils.py
```diff
@@ -97,10 +98,27 @@ def generate_s3_authorization_headers(key):
...
+        "ACL": "private",
+    }
+
+    headers = {
+        "X-amz-acl": "private",
+    }
+
+    if not _is_previewable(guessed_mime_type):
+        params["ContentDisposition"] = f'attachment; filename="{item.filename:s}"'
+        headers["Content-Disposition"] = f'attachment; filename="{item.filename:s}"'
```
qbey, line 121: I'm not sure this is tested, maybe I missed something

src/backend/core/api/utils.py
lunika: > What happened here? XD

I tested how coderrabit can be dumb.

src/backend/core/api/utils.py
lunika: It is tested in the `test_api_items_children_create.py` tests

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
lunika: yes, I changed it. Thanks

NathanVss (approved):
I approve because it solves an important security issue but we'll have to think of a dynamic solution, what if the previewable sets of file changes in the future? We should be able to on-the-fly set the content disposition of a file based on criterias

lunika:
Closed in favor of #355

PR #355, opened 2025-09-27
expose url_preview on item object

lunika:
## Purpose

We want to distinguish the url and the url_preview. The url_preview
property will be provided only if the mimetype is listed in the
ITEM_PREVIEWABLE_MIME_TYPES settings. Also an other route is added to
nginx forcing the content-disposition with the value attachment for
download url.

## Proposal

- [x] ✨(backend) expose url_preview on item object
- [x] ♻️(front) use preview url to preview a file
- [x] ✨(helm) add ingress managing preview url

### DIFF
--- a/src/nginx/servers.conf.erb
+++ b/src/nginx/servers.conf.erb
@@ -51,6 +51,25 @@ server {
         proxy_pass <%= ENV["AWS_S3_BUCKET_INTERNAL_URL"] %>;
         proxy_set_header Host <%= ENV["AWS_S3_BUCKET_INTERNAL_HOST"] %>;
         add_header Content-Security-Policy "default-src 'none'" always;
+        add_header Content-Disposition "attachment";
+    }
+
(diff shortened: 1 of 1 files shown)


docker/files/development/etc/nginx/conf.d/default.conf
```diff
@@ -17,6 +17,25 @@ server {
         proxy_set_header X-Amz-Date $authDate;
         proxy_set_header X-Amz-Content-SHA256 $authContentSha256;
 
+        # Get resource from Minio
+        proxy_pass http://minio:9000/drive-media-storage/;
+        proxy_set_header Host minio:9000;
+        add_header Content-Disposition "attachment";
```
NathanVss, line 23: Luckily we expose the filename in the url, so we don't have to specify filename here 😊

NathanVss (approved):
Yayy!

PR #357, opened 2025-09-29
Add Messages widget

NathanVss:
## Purpose

Capture d’écran 2025-09-29 à 11 44 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 - ✨(backend) create wopi applcation #2
 - ✨(backend) expose url_preview on item object
+- ✨(front) add messages widget #357
 
 ### Changed
 
--- a/src/frontend/apps/drive/src/features/drivers/types.ts
+++ b/src/frontend/apps/drive/src/features/drivers/types.ts
@@ -85,7 +85,7 @@ export type WopiInfo = {
   access_token: string;
   access_token_ttl: number;
   launch_url: string;
-}
+};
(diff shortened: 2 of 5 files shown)


sylvinus (approved):
LGTM!

src/backend/drive/settings.py
```diff
@@ -419,6 +419,24 @@ class Base(Configuration):
     FRONTEND_FEEDBACK_ITEMS = values.DictValue(
         {}, environ_name="FRONTEND_FEEDBACK_ITEMS", environ_prefix=None
     )
+    FRONTEND_FEEDBACK_MESSAGES_WIDGET = values.BooleanValue(
```
lunika: As it is a boolean, adding `ENABLED` somewhere in the name could be helpful I think

lunika: And it could be added to the frontend configuration file made in PR https://github.com/suitenumerique/drive/pull/299 instead of having a growing list of settings related to the frontend application ?

lunika (commented):
If you keep the settings, please update the `docs/env.md` file.

src/backend/drive/settings.py
NathanVss: I'll put all the pre existing frontend stuff in this configuration file in a new PR + we need to setup a way to be able to use this file with the ANCT deployment, that we haven't yet (@sylvinus), that's why we need to still rely on envvar for now

lunika (approved):

PR #358, opened 2025-09-29
🔖(minor) bump release version 0.6.0

lunika:
### Added

- ✨(backend) create wopi applcation #2
- ✨(backend) expose url_preview on item object #355

### Changed

- ♻️(backend) use PUT presigned-url to upload files #345


src/frontend/packages/sdk/src/Picker.ts
```diff
@@ -33,7 +33,7 @@ export const openPicker = async (
   } else {
     return {
       type: "cancelled",
-    };
```
NathanVss, line 36: Oupsy, remove this change from this commit?

PR #195, opened 2025-05-20
✨(back) integrate resource server API

lunika:
## Purpose

Expose endpoints for external api. This endpoints will allow to work on the `item` and `itemAccess` resources.

The external apis are available with these urls:

```
/external_api/v1.0/	rest_framework.routers.APIRootView	api-root
/external_api/v1.0/<drf_format_suffix:format>	rest_framework.routers.APIRootView	api-root
/external_api/v1.0/items/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-list
/external_api/v1.0/items/<pk>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-detail
/external_api/v1.0/items/<pk>/children/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-children
/external_api/v1.0/items/<pk>/children\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-children
/external_api/v1.0/items/<pk>/favorite/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-favorite
/external_api/v1.0/items/<pk>/favorite\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-favorite
/external_api/v1.0/items/<pk>/hard-delete/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-hard-delete
/external_api/v1.0/items/<pk>/hard-delete\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-hard-delete
/external_api/v1.0/items/<pk>/link-configuration/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-link-configuration
/external_api/v1.0/items/<pk>/link-configuration\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-link-configuration
/external_api/v1.0/items/<pk>/move/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-move
/external_api/v1.0/items/<pk>/move\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-move
/external_api/v1.0/items/<pk>/restore/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-restore
/external_api/v1.0/items/<pk>/restore\.<format>/	core.external_api.viewsets.ResourceServerItemViewSet	resource_server_items-restore
[...]

PR #379, opened 2025-10-17
✨(back) make Drive a resource server

NathanVss:
### Goal

We want to allow external SP to access drive resources, the first goal is to allow for items creation.

Examples:
- Save an attachment from https://github.com/suitenumerique/messages ( ping @jbpenrath )
- Store a video record from https://github.com/suitenumerique/meet ( ping @lebaudantoine )
- ...

### Implementation

This work is based on the library `django-lasuite` using the following implementation: https://github.com/suitenumerique/django-lasuite/blob/main/documentation/how-to-use-oidc-resource-server-backend.md

The resource server API is nested under a new subset of routes: `/external_api/v1.0/...`, the existing `/api/v1.0/...` is not meant to be used under resource server. 

We only exposed some routes, mainly routes that allow for:
- Items fetching
- Items creation, in order to allow for uploads to get a `policy`
- Items upload-ended to tell the app that the associated `policy` has received a file.

### More

We used some work already present in the existing PR: https://github.com/suitenumerique/drive/pull/195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to
 - ✨(front) add creating folder upload step
 - ✨(front) update to ui-kit v2
 - 🌐 add dutch translation
+- ✨(back) add resource server routes
 
 ### Fixed
 


src/backend/core/external_api/viewsets.py
```diff
@@ -0,0 +1,39 @@
...
+    @drf.decorators.action(
+        detail=False,
+        methods=["get"],
+        url_name="me",
+        url_path="me",
+        permission_classes=[IsAuthenticated & ResourceServerClientPermission],
+    )
+    def get_me(self, request):
+        """Get the current user."""
+        return super().get_me(request)
```
lunika: Without it, tests are passing. So maybe it's not needed to declare it or your tests are not covering well what you want.

src/backend/core/tests/conftest.py
```diff
@@ -25,3 +29,73 @@ def mock_user_teams():
...
+                "scope": "openid groups",
+                "active": True,
+            },
+        )
+
+        yield rsps
+
+
+@pytest.fixture(name="user_specific_sub")
+def fixture_user_specific_sub():
```
lunika: This is not needed IMO. You are not in the case where you are declaring the function `user_specific_sub` and declaring `user_specific_sub` as a function parameter in a test. You should try to rename your function without the `name` parameter fot the `@pytest.fixture` decorator.

src/backend/core/tests/conftest.py
```diff
@@ -25,3 +29,73 @@ def mock_user_teams():
...
+    client.post(
+        ...
+        HTTP_AUTHORIZATION=f"Bearer {build_authorization_bearer('some_token')}",
+    )
+    """
+    return base64.b64encode(token.encode("utf-8")).decode("utf-8")
+
+
+@pytest.fixture(name="user_token")
+def fixture_user_token():
```
lunika: Same here

src/backend/core/tests/external_api/items/test_external_api_items.py
```diff
@@ -0,0 +1,156 @@
...
+    response = client.get("/external_api/v1.0/items/")
+
+    assert response.status_code == 404
+
+
+def test_api_items_list_connected_resource_server(
+    user_token, resource_server_backend, user_specific_sub
+):
+    """Connected users should be allowed to list items from a resource server."""
+    reload_urls()
```
lunika: Why is it needed ?

src/backend/core/tests/external_api/items/test_external_api_items.py
```diff
@@ -0,0 +1,156 @@
+"""
+Tests for the Resource Server API for items.
```
lunika, line 2: You should also tests that other actions result in a `404`.

src/backend/core/tests/external_api/test_external_api_users.py
```diff
@@ -0,0 +1,90 @@
+"""
+Tests for the Resource Server API for users.
```
lunika, line 2: Same here for the other actions

src/backend/core/tests/utils/urls.py
```diff
@@ -0,0 +1,14 @@
+import importlib
+
+from django.urls import clear_url_caches
+
+
+def reload_urls():
```
lunika, line 8: This function already exists in the `test_e2e` module. Maybe you can extract it to make it more generic and use only one ?

bin/pytest
```diff
@@ -5,4 +5,4 @@ source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 _dc_run \
     -e DJANGO_CONFIGURATION=Test \
     app-dev \
-    pytest "$@"
```
lunika, line 8: IMO the `-s` option should not be a default.

compose.yaml
```diff
@@ -227,4 +233,10 @@ services:
     volumes:
       - ./docker/onlyoffice/local-development.json:/etc/onlyoffice/documentserver/local-production-linux.json
       # Enable debug
-      - ./docker/onlyoffice/log4js/production.json:/etc/onlyoffice/documentserver/log4js/production.json
\ No newline at end of file
+      - ./docker/onlyoffice/log4js/production.json:/etc/onlyoffice/documentserver/log4js/production.json
+
+networks:
```
lunika, line 238: IMO can be kept, same thing will be made in the search PR

src/backend/core/tests/external_api/items/test_external_api_items.py
NathanVss: As `resource_server_backend` is required as fixture, we need to reload the urls to take into account `OIDC_RESOURCE_SERVER_ENABLED` to make `/external_api` routes available. ( url are not reloaded between tests )

lunika (approved):
Awesome. Thanks

Issue #287, opened 2025-07-28
Advanced shared feature


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

As a user I want to determine, for each file/folder, how they are shared.

**Describe the solution you'd like**

This feature is shared between Docs and Drive. It will be implemented first in Docs and then backport in Drive.

PR #383, opened 2025-10-29
Advanced shared feature

lunika:
## Purpose

Manage advanced shared in its own workspace and shared workspace.

specs (in french) : https://docs.numerique.gouv.fr/docs/ab704b19-18e3-46e3-9eb1-39bb2548dd35/

https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=9171-8103

## Proposal

- [x] backend part
- [x] Add missing link-configuration endpoint

Implement #287

### DIFF
--- a/.github/workflows/build-mails.yml
+++ b/.github/workflows/build-mails.yml
@@ -1,0 +1,45 @@
+name: Build mail templates reusable workflow
+
+on:
+  workflow_call:
+
+jobs:
+  build-mails:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: src/mail
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Restore the mail templates
+        uses: actions/cache@v4
+        id: mail-templates
+        with:
+          path: "src/backend/core/templates/mail"
+          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
+
+      - name: Install yarn
+        if: steps.mail-templates.outputs.cache-hit != 'true'
+        run: npm install -g yarn
+
+      - name: Install node dependencies
+        if: steps.mail-templates.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile
+
+      - name: Build mails
+        if: steps.mail-templates.outputs.cache-hit != 'true'
+        run: yarn build
+
+      - name: Cache mail templates
+        if: steps.mail-templates.outputs.cache-hit != 'true'
+        uses: actions/cache@v4
+        with:
+          path: "src/backend/core/templates/mail"
+          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
--- a/.github/workflows/drive-frontend.yml
+++ b/.github/workflows/drive-frontend.yml
@@ -14,6 +14,9 @@ jobs:
     with:
       node_version: "22.x"
 
+  build-mails:
+    uses: ./.github/workflows/build-mails.yml
+
   lint-front:
     runs-on: ubuntu-latest
     needs: install-front
@@ -61,7 +64,7 @@ jobs:
 
   test-e2e:
     runs-on: ubuntu-latest
-    needs: install-front
+    needs: [install-front, build-mails]
     strategy:
       matrix:
         browser:
@@ -83,6 +86,12 @@ jobs:
           path: "src/frontend/**/node_modules"
           key: front-node_modules-${{ hashFiles('src/frontend/**/yarn.lock') }}
           fail-on-cache-miss: true
+
+      - name: Restore the mail templates
+        uses: actions/cache@v4
+        with:
+          path: "src/backend/core/templates/mail"
+          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
 
       - name: Install Playwright Browsers
         run: |
--- a/.github/workflows/drive.yml
+++ b/.github/workflows/drive.yml
@@ -61,44 +61,7 @@ jobs:
           fi
 
   build-mails:
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: src/mail
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
-
-      - name: Install Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: "18"
-
-      - name: Restore the mail templates
-        uses: actions/cache@v4
-        id: mail-templates
-        with:
-          path: "src/backend/core/templates/mail"
-          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
-
-      - name: Install yarn
-        if: steps.mail-templates.outputs.cache-hit != 'true'
-        run: npm install -g yarn
-
-      - name: Install node dependencies
-        if: steps.mail-templates.outputs.cache-hit != 'true'
-        run: yarn install --frozen-lockfile
-
-      - name: Build mails
-        if: steps.mail-templates.outputs.cache-hit != 'true'
-        run: yarn build
-
-      - name: Cache mail templates
-        if: steps.mail-templates.outputs.cache-hit != 'true'
-        uses: actions/cache@v4
-        with:
-          path: "src/backend/core/templates/mail"
-          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
+    uses: ./.github/workflows/build-mails.yml
 
   lint-back:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(global) implement advanced shared management system
+
+### Removed
+
+- 🔥(global) remove notion of workspace
+
 ## [v0.12.0] - 2026-02-06
 
 ### Added


src/backend/core/api/permissions.py
```diff
@@ -95,31 +94,51 @@ def has_permission(self, request, view):
...
+    def has_object_permission(self, request, view, obj):
+        """Check permission for a given object."""
+        abilities = obj.get_abilities(request.user)
+        action = view.action
+        return abilities.get(action, False)
+
 
-class AccessPermission(permissions.BasePermission):
+class ItemAccessPermission(IsAuthenticated):
     """Permission class for access objects."""
```
lunika: Update docstring, not accurate anymore.

src/backend/core/api/serializers.py
```diff
@@ -33,95 +34,106 @@ class Meta:
         read_only_fields = ["id", "full_name", "short_name"]
 
 
-class BaseAccessSerializer(serializers.ModelSerializer):
-    """Serialize template accesses."""
+class ItemLightSerializer(serializers.ModelSerializer):
+    """Minial item serializer for nesting in document accesses."""
```
```suggestion
    """Minial item serializer for nesting in item accesses."""
```

qbey:
```suggestion
    """Minimal item serializer for nesting in item accesses."""
```

src/backend/core/api/serializers.py
```diff
@@ -472,13 +529,69 @@ class LinkItemSerializer(serializers.ModelSerializer):
...
 
+    def validate(self, attrs):
+        """Validate that link_role and link_reach are compatible using get_select_options."""
+        link_reach = attrs.get("link_reach")
+        link_role = attrs.get("link_role")
+
+        if not link_reach:
+            raise serializers.ValidationError(
+                {"link_reach": _("This field is required.")}
+            )
```
qbey, line 680: How would it be possible to reach this? Are not the fields validated before?

src/backend/core/api/serializers.py
```diff
@@ -472,13 +529,69 @@ class LinkItemSerializer(serializers.ModelSerializer):
...
         ]
 
+    def validate(self, attrs):
+        """Validate that link_role and link_reach are compatible using get_select_options."""
+        link_reach = attrs.get("link_reach")
+        link_role = attrs.get("link_role")
+
+        if not link_reach:
+            raise serializers.ValidationError(
+                {"link_reach": _("This field is required.")}
```
qbey, line 679: Other validation errors are not translated, I guess the frontend never use the values from the backend, maybe no translation needed here.

src/backend/core/api/serializers.py
```diff
@@ -196,28 +231,32 @@ class Meta:
...
+    def get_user_role(self, item):
         """
         Return roles of the logged-in user for the current item,
         taking into account ancestors.
         """
         request = self.context.get("request")
         if request:
-            return item.get_roles(request.user)
+            return item.get_role(request.user)
         return []
```
qbey: Is there a type inconsistency here? get_user_role should not return a list I guess

src/backend/core/api/viewsets.py
```diff
@@ -1260,23 +1140,220 @@ class ItemAccessViewSet(
...
+        if role not in PRIVILEGED_ROLES:
+            accesses_qs = accesses_qs.filter(role__in=PRIVILEGED_ROLES)
+
+        accesses_qs = accesses_qs.annotate_user_roles(user)
+        accesses = list(accesses_qs.order_by("item__path", "created_at").iterator())
+
+        request_target_keys = {f"user:{user.id}"}
+        request_target_keys.update(
+            f"team:{team}" for team in getattr(user, "teams", [])
+        )
```
qbey: Where is it used?

PanchoutNathan:
@coderabbitai review

PanchoutNathan:
@coderabbitai full review

src/frontend/apps/drive/src/features/drivers/DTOs/AccessesDTO.ts
```diff
@@ -1,20 +1,25 @@
...
 export type DTOUpdateAccess = {
-    itemId: string;
-    accessId: string;
-    user_id: string;
-    role: Role;  
+  itemId: string;
+  accessId: string;
+  user_id: string;
+  role: Role;
 };
```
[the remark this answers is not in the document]
PanchoutNathan: Yes but not in this PR

src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
```diff
@@ -251,22 +231,17 @@ export const GlobalExplorerProvider = ({
...
-          ) {
-            const workspaces =
-              id === WorkspaceCategory.SHARED_SPACE
-                ? WorkspaceType.SHARED
-                : WorkspaceType.PUBLIC;
-            const response = await driver.getItems({
+        onLoadChildren={async (itemId, page) => {
+          // If the itemId is a favorite item, we need to get the favorite items.
+          const id = itemId.split("_")[0];
+          if (id === DefaultRoute.FAVORITES) {
```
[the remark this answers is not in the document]
PanchoutNathan: Currently, we append a _X suffix at the end of favorite item IDs to differentiate them when a node containing a favorite child is opened. This avoids duplicate IDs in the tree component, which would otherwise cause errors. Therefore, we need to remove the _X suffix for API requests.

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.scss
```diff
@@ -0,0 +1,17 @@
...
+    cursor: pointer;
+    font: inherit;
+
+    &:focus {
+      outline: 2px solid
+        var(--c--contextuals--content--semantic--brand--primary);
+      outline-offset: 2px;
+    }
+  }
+}
```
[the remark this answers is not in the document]
PanchoutNathan: nope

src/frontend/apps/drive/package.json
```diff
@@ -16,8 +16,8 @@
     "yarn": "1.22.22"
   },
   "dependencies": {
-    "@gouvfr-lasuite/ui-kit": "0.18.4",
-    "@openfun/cunningham-react": "4.0.0",
+    "@gouvfr-lasuite/ui-kit": "/Users/melde/Documents/ui-kit",
+    "@gouvfr-lasuite/cunningham-react": "4.1.0",
```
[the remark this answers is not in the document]
PanchoutNathan: false

src/backend/core/models.py
```diff
@@ -675,99 +686,164 @@ def invalidate_nb_accesses_cache(self):
...
-            else set()
+        has_access_role = bool(role) and not is_deleted
+        link_select_options = (
+            LinkReachChoices.get_select_options(**self.ancestors_link_definition)
+            if has_access_role
+            else {}
         )
 
-        roles = roles | public_roles | authenticated_roles
+        link_definition = get_equivalent_link_definition(
```
NathanVss: Could `computed_link_definition` be used here?

src/backend/core/models.py
```diff
@@ -675,99 +686,164 @@ def invalidate_nb_accesses_cache(self):
...
+                {"link_reach": self.link_reach, "link_role": self.link_role},
+            ]
+        )
 
-        can_get = bool(roles) and not is_deleted
+        link_reach = link_definition["link_reach"]
+        if link_reach == LinkReachChoices.PUBLIC or (
+            link_reach == LinkReachChoices.AUTHENTICATED and user.is_authenticated
+        ):
+            role = RoleChoices.max(role, link_definition["link_role"])
```
NathanVss, line 758: Could you explain ( and add a comment in the code ) the logic behind this ? Why do we need to make an special case here?

src/backend/core/tests/items/test_api_items_children_list.py
```diff
@@ -501,6 +535,10 @@ def test_api_items_children_list_authenticated_related_direct():
         "results": [
             {
                 "abilities": child1.get_abilities(user),
+                "ancestors_link_reach": "restricted",
+                "ancestors_link_role": None,
+                "computed_link_reach": child1.computed_link_reach,
+                "computed_link_role": child1.computed_link_role,
```
NathanVss, line 541: Maybe I missed it but I don't see tests that explicitely makes assertion on those two `computed_` attribute with hard coded values

src/backend/core/api/viewsets.py
```diff
@@ -1273,36 +1262,20 @@ def get_queryset(self):
...
-                        **{f"{self.resource_field_name}__accesses__team__in": teams}
-                    ),
-                    **{self.resource_field_name: self.kwargs["resource_id"]},
-                )
-                .annotate(user_roles=db.Subquery(user_roles_query))
-                .distinct()
-            )
         return queryset
 
+    def list(self, request, *args, **kwargs):
```
NathanVss, line 1593: It seems that you don't use `accesses_view` ( and other ) anymore? Am I right? 

If so, we could check for `item.get_abilities(user)['access_view']` instead of a hard computed permission in here like that

```

        if item.get_role(user) is None:
            return drf.response.Response([])
```

WDYT?

src/backend/core/api/viewsets.py
```diff
@@ -1249,49 +1250,99 @@ def item(self):
...
-        # Check if the access being deleted is the last owner access for the resource
-        if (
-            instance.role == "owner"
-            and resource.accesses.filter(role="owner").count() == 1
-        ):
-            return drf.response.Response(
-                {"detail": "Cannot delete the last owner access for the resource."},
-                status=drf.status.HTTP_403_FORBIDDEN,
+        if role not in PRIVILEGED_ROLES:
+            queryset = queryset.filter(role__in=PRIVILEGED_ROLES)
```
NathanVss: Why? ( could add a comment in the code )

src/backend/core/api/viewsets.py
NathanVss: +1

src/backend/core/models.py
```diff
@@ -1167,57 +1168,115 @@ def delete(self, *args, **kwargs):
...
-                roles = self.user_roles or []
-            except AttributeError:
-                try:
-                    roles = self.item.accesses.filter(
-                        models.Q(user=user) | models.Q(team__in=teams),
-                    ).values_list("role", flat=True)
-                except (self._meta.model.DoesNotExist, IndexError):
-                    roles = []
+        ancestors_role, current_role = self.get_user_roles_tuple(user)
+        role = RoleChoices.max(ancestors_role, current_role)
```
NathanVss: We could add a dedicated method `get_role` that handle this logic, it would improve readability with encapsulation

src/backend/core/api/viewsets.py
```diff
@@ -1278,48 +1277,42 @@ def list(self, request, *args, **kwargs):
...
             return drf.response.Response([])
 
-        ancestor_qs = (
-            self.item.ancestors() | models.Item.objects.filter(pk=self.item.pk)
-        ).filter(ancestors_deleted_at__isnull=True)
-        ancestor_items = list(ancestor_qs.order_by("path"))
-        ancestor_ids = [item.id for item in ancestor_items]
-
-        queryset = self.get_queryset().filter(item__in=ancestor_ids)
+        ancestors_qs = models.Item.objects.filter(
```
NathanVss, line 1609: Adding a bit more comments in for this route would be great to make understanding easier

src/backend/core/models.py
```diff
@@ -1285,24 +1285,24 @@ def get_abilities(self, user, is_explicit=True):
...
+            ancestors_role_priority = RoleChoices.get_priority(self.max_ancestors_role)
             set_role_to = [
                 candidate_role
                 for candidate_role in set_role_to
                 if RoleChoices.get_priority(candidate_role) >= ancestors_role_priority
             ]
         else:
+            ancestors_role_priority = RoleChoices.get_priority(
+                self.max_ancestors_role or self.role
+            )
```
NathanVss: When can that happen?

src/frontend/apps/drive/src/features/api/fetchApi.ts
NathanVss, line 1: It is possible to try deleting the pre existing main workspace causing a crash

Capture d’écran 2026-01-15 à 14 18 10

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -25,145 +34,67 @@ export const useMutationCreateFile = () => {
...
       return driver.createFolder(...payload);
     },
     onSuccess: (data, variables) => {
-      if (variables.parentId) {
-        queryClient.invalidateQueries({
-          queryKey: ["items", variables.parentId],
-        });
-      }
+      const queryKey = variables.parentId
+        ? ["items", variables.parentId, "children"]
```
NathanVss, line 115: It seems that if there is a parent the actual key is much more like 

```
      "items",
      itemId,
      "children",
      "infinite",

```

Nope ? As from [id].tsx we use `useInfiniteChildren`

src/frontend/apps/drive/src/features/explorer/hooks/useQueries.tsx
```diff
@@ -41,7 +46,18 @@ export const useItems = () => {
...
+export const useItem = (
+  itemId: string,
+  options?: HookUseQueryOptions<Item>
+) => {
+  return useQuery({
+    queryKey: ["items", itemId],
+    queryFn: () => getDriver().getItem(itemId),
+    ...options,
+  });
+};
```
NathanVss, line 74: Maybe you could use it in `files/[id].tsx` nope.?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -25,145 +34,67 @@ export const useMutationCreateFile = () => {
...
+        : ["items", "infinite", JSON.stringify({ is_creator_me: true })];
+      addItemToTopOfPaginatedList(queryKey, data);
+
+      queryClient.invalidateQueries({
+        queryKey: ["rootItems"],
+      });
```
NathanVss: Is this still needed?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+  const onSuccess = () => {
+    queryClient.invalidateQueries({
+      queryKey,
+    });
+  };
+
+  return { onMutate, onError, onSuccess };
+};
+
+export const useRefreshItemCache = () => {
```
NathanVss: Plz provide a comment to explain the purpose

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+  const driver = getDriver();
+  const treeContext = useTreeContext();
+  const refreshFavoriteCache = useRefreshFavoriteCache();
+  return useMutation({
+    mutationFn: (...payload: Parameters<typeof driver.deleteFavoriteItem>) => {
+      return driver.deleteFavoriteItem(...payload);
+    },
+    onSuccess: (_data, itemId: string) => {
+      // We add the path level to the id to avoid conflicts with the same id inside the tree for favorite items.
+      treeContext?.treeData.deleteNode(itemId + "_0");
```
NathanVss: Shouldn't we remove all `treeContext?.treeData.deleteNode(itemId + "_*");` if this item exists multiple times in the tree?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+    if (partialUpdate) {
+      updateItemInPaginatedList(["items"], itemId, partialUpdate);
+      queryClient.setQueryData(["item", itemId], (old: Item) => {
+        return {
+          ...old,
+          ...partialUpdate,
+        };
+      });
+    } else {
+      const item = await driver.getItem(itemId);
```
NathanVss: Why just not invalidating the "item", * query instead of doing this? Seems more "react-query-ish" nope?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
     onSuccess: (_, variables) => {
-      queryClient.invalidateQueries({
-        queryKey: ["itemInvitations", variables.itemId],
-      });
+      onSuccessAccessOrInvitation(variables.itemId, true);
     },
   });
 };
+
+export const useRefreshFavoriteCache = () => {
```
NathanVss: Seems pretty similar to `useRefreshFavoriteCache`, isn't quite the same as calling `useRefreshItemCache(id, {is_favorite: true})` or smth like that?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+      refreshFavoriteCache(itemId, false);
+    },
+  });
+};
+
+export const useGetQueryKey = () => {
+  const router = useRouter();
+  return (parentId?: string) => {
+    const isOnMyFilesRoute = isMyFilesRoute(router.pathname);
+    let queryKey = parentId ? ["items", parentId, "children"] : [];
```
NathanVss: is `["items", parentId, "children"]` key still used somewhere?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+      treeContext?.treeData.deleteNode(itemId + "_0");
+      refreshFavoriteCache(itemId, false);
+    },
+  });
+};
+
+export const useGetQueryKey = () => {
+  const router = useRouter();
+  return (parentId?: string) => {
+    const isOnMyFilesRoute = isMyFilesRoute(router.pathname);
```
NathanVss: Dependency inversion here imho. 

The route mecanism should be built on top of the model. 

The model is here relying also on the routing, which reveals a deps inversion bad pattern imho.

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+    let queryKey = parentId ? ["items", parentId, "children"] : [];
+    if (isOnMyFilesRoute) {
+      queryKey = getQueryKeyForRouteId(router.pathname);
```
NathanVss: plus: you are calling `getQueryKeyForRouteId` only if `isOnMyFilesRoute` which `getQueryKeyForRouteId` already calls itself internally. Looks cumbersome, why not invalidating all the queries `items,infinite`?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+    }
+    return queryKey;
+  };
+};
+
+export const useRemoveFromQueryCache = () => {
```
NathanVss: Does not seem to be used anymore

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+      queryKey,
+      (context as { previousItems: Item[] })?.previousItems
+    );
+  };
+
+  const onSuccess = () => {
+    if (queryKey.length === 0) {
+      return;
+    }
+    console.log("onSuccess", queryKey);
```
NathanVss: :)

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -265,81 +211,335 @@ export const useMutationDeleteWorskpace = () => {
...
+    queryClient.invalidateQueries({
+      queryKey,
+    });
+  };
+
+  return { onMutate, onError, onSuccess };
+};
+
+export const useUpdateMutationCallbacks = (
```
NathanVss: Isn't this a duplication of `updateItemInPaginatedList` ?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
NathanVss, line 1: As you can see according to my comments in this file I'm pretty sure it can be rationalized

src/frontend/apps/drive/src/features/explorer/hooks/useOptimisticPagination.ts
NathanVss, line 1: Maybe put this in hooks and call it `useOptimisticPagination.ts` smth easier?

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
NathanVss: Could you also order the type of function we have in this files ? The utils and actual mutator are mixed, not much easy to read, WDYT about putting actual mutators first and utils at the bottom of the file ? With a comment to seperator both

src/frontend/apps/drive/src/features/i18n/translations.json
NathanVss, line 1: Isn't "Espace" to be replaced with "Dossier" ?

Capture d’écran 2026-01-15 à 14 54 41
Capture d’écran 2026-01-15 à 14 55 13

src/frontend/apps/drive/src/utils/itemUtils.ts
NathanVss, line 1: instead of a new file WDYT about using `features/items/utils.ts`?

src/frontend/apps/drive/src/features/i18n/translations.json
NathanVss, line 1: The old main workspace now appears as "Workspace" in the explorer, should we do something special about this one for UX ? Special naming or pulling all its existing files into `My files` for the migration? @rl-83 @PanchoutNathan @lunika 

Capture d’écran 2026-01-15 à 15 01 14

NathanVss: I'm scared that the user will panic or don't understand

src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
```diff
@@ -311,15 +288,10 @@ export const GlobalExplorerProvider = ({
...
-        title={t("file_preview.title")}
-        files={previewFiles}
-        onChangeFile={handleChangePreviewItem}
-        handleDownloadFile={() => handleDownloadItem(previewItem)}
-        openedFileId={previewItem?.id}
-        sidebarContent={previewItem && <ItemInfo item={previewItem} />}
+      <CustomFilesPreview
+        currentItem={previewItem}
+        items={previewItems}
+        onSetPreviewItem={setPreviewItem}
```
NathanVss: `setPreviewItem` instead of `onSet` nope?

src/frontend/apps/drive/src/features/explorer/components/GlobalExplorerContext.tsx
```diff
@@ -404,13 +344,24 @@ const TreeProviderInitializer = ({
...
-export const itemToTreeItem = (item: Item, parentId?: string): TreeItem => {
+export const itemToTreeItem = (
+  item: Item,
+  parentId?: string,
+  isFavoriteItem?: boolean
+): TreeItem => {
+  const pathLevel = item.path.split(".").length - 1;
+  // We add the path level to the id to avoid conflicts with the same id inside the tree.
+  // This is useful when we have a favorite item that is a child of another item.
+  const id = item.id + (isFavoriteItem ? `_${pathLevel}` : "");
```
NathanVss: maybe I'm stupid or something but if the same item "A.B.C" appears multiple times in the tree, appending its pathLevel will always give the same result nope ? 

A.B.C + ("A.B.C.".split(".").length-1) will always give "A.B.C_3" nope?

src/frontend/apps/drive/src/features/explorer/components/modals/move/ExplorerMoveFolderModal.tsx
NathanVss, line 1: The first breadcrumb is not needed? It's a repetition and we cannot click on it, WDYT? @rl-83 

Capture d’écran 2026-01-15 à 15 15 51

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorer.tsx
```diff
@@ -101,14 +123,31 @@ export const EmbeddedExplorer = (props: EmbeddedExplorerProps) => {
...
   const items = useMemo(() => {
     // If itemChildren are not loaded yet, we want to return undefined in order to display loading state.
     if (itemChildren === undefined && props.currentItemId) {
       return undefined;
     }
+    if (isSearchItemsLoading) {
+      return itemsRef.current;
+    }
     let items = [];
     // If no itemId, we are in the root, we explorer spaces
```
NathanVss: Outdated comment, let's update it

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorer.tsx
```diff
@@ -101,14 +123,31 @@ export const EmbeddedExplorer = (props: EmbeddedExplorerProps) => {
...
-    if (props.currentItemId === null) {
+    if (searchQuery !== "") {
+      items = searchItems ?? [];
+    } else if (props.currentItemId === null) {
```
NathanVss, line 149: Maybe the logic below with main_workspace should evolve and does not makes sense anymore nope?

```
      if (user?.main_workspace) {
        items.push(user.main_workspace);
      }
      items = items.concat(rootItems ?? []);

      // Sort items to put main_workspace first
      items = items.sort((a, b) => {
        if (a.main_workspace && !b.main_workspace) return -1;
        if (!a.main_workspace && b.main_workspace) return 1;
        return 0;
      });
```

src/frontend/apps/drive/src/features/explorer/components/tree/DroppableNodeTree.tsx
NathanVss, line 1: BUG
https://github.com/user-attachments/assets/b06c5c59-2d65-4ae2-99ad-f70e2943c2bb

NathanVss: I cannot drag as easily as before, click and drag does seems to work anymore


https://github.com/user-attachments/assets/27f3a8b1-00ce-4dba-bb9f-8771c91aeeab

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeItemActions.tsx
NathanVss, line 1: BUG when clicking on share on my workspace
https://github.com/user-attachments/assets/2800b0b0-5598-432e-a585-69e829d62ec9

NathanVss:
BUG: Cannot retrieve favorite files
https://github.com/user-attachments/assets/81b0c1a8-b155-4f3c-8c19-ccd79a9f4b6a

( The frontend enforces a type=folder filter )

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -128,13 +132,20 @@ const useUpload = ({ item }: { item: Item }) => {
...
-                    queryKey: ["items", parentItem.id],
-                  });
+                  if (isMyFilesRoute(router.pathname)) {
+                    queryClient.invalidateQueries({
+                      queryKey: getMyFilesQueryKey(),
+                    });
+                  } else {
+                    queryClient.invalidateQueries({
+                      queryKey: ["items", parentItem.id],
+                    });
```
NathanVss: Why don't we have do perform 

```
                    queryClient.invalidateQueries({
                      queryKey: ["items", parentItem.id],
                    });
```

When being on "my files" route?

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -128,13 +132,20 @@ const useUpload = ({ item }: { item: Item }) => {
...
+                parentId: parentItem?.id,
               },
               {
                 onSuccess: async (createdFolder) => {
-                  queryClient.invalidateQueries({
-                    queryKey: ["items", parentItem.id],
-                  });
+                  if (isMyFilesRoute(router.pathname)) {
+                    queryClient.invalidateQueries({
+                      queryKey: getMyFilesQueryKey(),
```
NathanVss: Ditto ( dependency inversion )

src/backend/core/api/filters.py
NathanVss, line 1: backend file in `✨(frontend) refactor filters and enhance explorer pages` commit, is it expected?

src/frontend/apps/drive/src/features/i18n/translations.json
NathanVss: Does this still makes sense?

Capture d’écran 2026-01-15 à 15 44 17

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+        return (
+          (sameUser || sameTeam) &&
+          candidateAccess.max_role === access.max_ancestors_role
+        );
+      });
+
+      return {
+        ...access,
+        parent_id_max_role: parentAccess?.item.id,
```
NathanVss: Couldn't you use `max_ancestors_role_item_id` provided by the backend instead of computing it by yourself?

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.tsx
```diff
@@ -0,0 +1,472 @@
...
+        if (availableRoles.length === 0 && access.role !== Role.OWNER) {
+          return t("share_modal.options.top_message.to_lower_role");
+        }
+        return undefined;
+      }}
+      getAccessRoles={(access) => {
+        const availableRoles = access.abilities.set_role_to;
+        // const availableRoles = [Role.EDITOR, Role.ADMIN, Role.OWNER];
+
+        const isDisabled = (role: Role) => {
```
NathanVss, line 409: Missing white space

Capture d’écran 2026-01-15 à 15 56 40

NathanVss:
@lunika I noticed the `/recents` route to be quite slow. I notice via the toolbar that lots of requests are duplicated and most of then running on seq scan, I may lack improvements you provided for other routes, do you have an idea on how to improve that?

NathanVss (commented):
It was a looooooong way!! GG guys for this huge PR, enjoy reading my comments more than I did reviewing 10k lines 🤣

I will approve once we did testing on staging ;)

src/frontend/apps/drive/src/features/api/fetchApi.ts
lunika: More a backend comment for me I guess.

src/backend/core/api/viewsets.py
```diff
@@ -946,14 +933,19 @@ def search(self, request, *args, **kwargs):
             item__path__depth=1,
         )
 
+        # Remove items with upload_state SUSPICIOUS for non-creators
+        queryset = self._filter_suspicious_items(queryset, user)            
+
+        queryset = queryset.annotate_is_favorite(user)
```
lunika, line 1240: We know that filtering on an annotation does not scale well.
We removed this kind of filtering on the trashbin endpoint for example.
Take a look of the sql generated, you will see too many subrequests made.

src/backend/core/api/viewsets.py
```diff
@@ -701,6 +704,17 @@ def move(self, request, *args, **kwargs):
 
         item.move(target_item)
 
+        # If the item is moved to the root and the user does not have an access on the item,
+        # create an owner access for the user. Otherwise, the item will be invisible for the user.
+        if not target_item and not models.ItemAccess.objects.filter(item=item, user=user).exists():
```
lunika: What about people having access to this item ? Should we keep access (explicit or not) for at least users with privileged roles?

NathanVss: Discussed with @rl-83, it must be considered the same as a move inside a personal folder, so we don't keep the accesses

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
PanchoutNathan: No, I want to add it to all possible keys, so the queryKey is more of a prefix.

src/frontend/apps/drive/src/features/explorer/hooks/useQueries.tsx
PanchoutNathan: yep

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
PanchoutNathan: nope good catch

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
PanchoutNathan: Not for this one, because we're just removing the item's favorites feature, so it can only be at the first level.

src/backend/core/api/serializers.py
lunika: It can't, and there is a test enforcing it. Good catch

lunika: In fact no, there is a asserting that the `link_reach` must be present and the error message is changing in the test. So this code is reached when no `link_reach` is present in the request.

src/backend/core/models.py
lunika:
```
# Set the user role to the highest role between the item role and the link role
# Needed for a user with an access lower than link_role
# Needed for a user without access to determine the role he has.
```

src/backend/core/api/viewsets.py
lunika: We had a PII disclosure issue. Only users with privileged roles should see all the accesses. Other user should see only users with privileged roles.

src/backend/core/api/viewsets.py
lunika: Computing only the role is less expensive than computing all abilities for an item.

src/frontend/apps/drive/src/features/i18n/translations.json
PanchoutNathan: Yes but... it work with workspace attribute in back, maybe change to folder ? @lunika @NathanVss

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeItemActions.tsx
PanchoutNathan: @lunika

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
PanchoutNathan: i don't have ... parent ? haha. For my files route it's specific route and parentItem is undifined. I update all types

src/backend/core/models.py
lunika: When it is not an explicit and the related access if the first one in the tree.

src/backend/core/models.py
lunika: I think this comment is outdated. There is a `get_role` method, and I don't see anymore the commented code.

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeItemActions.tsx
lunika: @PanchoutNathan The route is not the good one. Look at the video, `_0` is added after the item id.

PR #391, opened 2025-11-06
Index to search

joehybird:
## Purpose

We want to add fulltext (and semantic in a second phase) search to Drive.

The goal is to enable efficient and scalable search across document content by pushing relevant data to a dedicated search backend, such as OpenSearch. The backend should be pluggable. 

## Proposal

- [x] Add indexing logic in a search indexer that can be declared as a backend
- [x] Implement indexing for the Find backend. See [corresponding PR in Find](https://github.com/suitenumerique/find/pull/16)
- [x] Implement search views as a proxy
- [x] Implement triggers to update search index when a document or its accesses change.

This is a backport of this [PR on docs](https://github.com/suitenumerique/docs/pull/1276)

### DIFF
--- a/docs/setup-find.md
+++ b/docs/setup-find.md
@@ -15,7 +15,7 @@ See [how-to-use-indexer.md](how-to-use-indexer.md) for details.
 
 ## Configure settings of Drive
 
-Add those Django settings the Drive application to enable the feature.
+Add those Django settings the Drive application to enable the indexation feature.
 
 ```python
 SEARCH_INDEXER_CLASS="core.services.search_indexers.SearchIndexer"
@@ -42,4 +42,10 @@ We also need to enable the **OIDC Token** refresh or the authentication will fai
 OIDC_STORE_ACCESS_TOKEN = True  # Store the access token in the session
 OIDC_STORE_REFRESH_TOKEN = True  # Store the encrypted refresh token in the session
 OIDC_STORE_REFRESH_TOKEN_KEY = "your-32-byte-encryption-key=="  # Must be a valid Fernet key (32 url-safe base64-encoded bytes)
+```
+
+And to enable the search of indexed files through the API, add this feature flag (default=True)
+
+```python
+FEATURES_INDEXED_SEARCH=True
 ```
--- a/env.d/development/common
+++ b/env.d/development/common
@@ -80,6 +80,9 @@ SEARCH_INDEXER_SECRET=find-api-key-for-driv-with-exactly-50-chars-length   # Key
 SEARCH_INDEXER_URL="http://find:8000/api/v1.0/documents/index/"
 SEARCH_INDEXER_QUERY_URL="http://find:8000/api/v1.0/documents/search/"
 
+# Enables the search of indexed files through the API
+FEATURES_INDEXED_SEARCH=True
+
 # Store OIDC tokens in the session
 # OIDC_STORE_ACCESS_TOKEN = True
 # OIDC_STORE_REFRESH_TOKEN = True # Store the encrypted refresh token in the session.
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1097,7 +1097,7 @@ class ItemViewSet(
         Applies filtering based on request parameter 'q' from `SearchItemFilter`.
         Depending of the configuration it can be:
          - A fulltext search through the opensearch indexation app "find" if the backend is
-           enabled (see SEARCH_INDEXER_CLASS)
+           enabled (see SEARCH_INDEXER_CLASS) and the feature flag INDEXED_SEARCH_ENABLED is True
          - A filtering by the model fields 'title' & 'type'.
         """
         queryset = self.queryset
@@ -1136,7 +1136,8 @@ class ItemViewSet(
 
         queryset = queryset.filter(path_list)
 
-        if indexer:
+        # use indexed search ONLY when the feature flag is enabled
+        if indexer and settings.FEATURES_INDEXED_SEARCH is True:
             # When the indexer is configured pop "title" from queryset search and use
             # fulltext results instead.
             return self._indexed_search(
--- a/src/backend/core/services/search_indexers.py
+++ b/src/backend/core/services/search_indexers.py
@@ -278,7 +278,8 @@ class SearchIndexer(BaseItemIndexer):
         mimetype = item.mimetype or ""
 
         if mimetype.startswith("text/"):
-            return default_storage.open(item.file_key, "rb").read().decode()
+            with default_storage.open(item.file_key, "rb") as fd:
+                return fd.read().decode()
 
         raise SuspiciousFileOperation(f"Unrecognized mimetype {mimetype}")
 
--- a/docs/env.md
+++ b/docs/env.md
@@ -42,6 +42,7 @@ This document lists all configurable environment variables for the Drive applica
 | `EMAIL_USE_SSL` | Use SSL for SMTP connection | `False` |
 | `EMAIL_USE_TLS` | Use TLS for SMTP connection | `False` |
 | `FEATURES_ALPHA` | Enable alpha features | `False` |
+| `FEATURES_INDEXED_SEARCH` | Enable the search of indexed files through the API | `True` |
 | `FRONTEND_THEME` | Frontend theme configuration | `None` |
 | `FRONTEND_FEEDBACK_BUTTON_SHOW` | Show feedback button | `False` |
 | `FRONTEND_FEEDBACK_BUTTON_IDLE` | Make feedback button idle (e.g. to bind to external library) | `False` |
--- a/src/backend/core/tests/items/test_api_items_search_fulltext.py
+++ b/src/backend/core/tests/items/test_api_items_search_fulltext.py
@@ -437,3 +437,36 @@ def test_api_items_search_pagination_endpoint_is_none(
         assert content["count"] == expected["count"]
         assert content["previous"] == previous_url
         assert content["next"] == next_url
+
+
+def test_api_items_search_feature_disabled(indexer_settings):
+    """Should not use indexed search if the feature is disabled"""
+    indexer_settings.FEATURES_INDEXED_SEARCH = False
+
+    assert get_file_indexer() is not None
+
+    user = factories.UserFactory()
+
+    client = APIClient()
+    client.force_login(user)
+
+    docs = factories.ItemFactory.create_batch(
+        5,
+        title="alpha",
+        users=[user],
+        parent=user.get_main_workspace(),
+    )
+
+    response = client.get(
+        "/api/v1.0/items/search/",
+        data={
+            "title": "alpha",
+        },
+    )
+
+    assert response.status_code == 200
+
+    content = response.json()
+    results = content.pop("results")
+
+    assert [r["id"] for r in results] == [str(d.pk) for d in docs]
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -159,6 +159,10 @@ class Base(Configuration):
         },
     }
 
+    FEATURES_INDEXED_SEARCH = values.BooleanValue(
+        default=True, environ_name="FEATURES_INDEXED_SEARCH", environ_prefix=None
+    )
+
     # Posthog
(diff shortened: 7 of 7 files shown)


src/backend/demo/management/commands/create_demo.py
```diff
@@ -0,0 +1,182 @@
...
+
+from faker import Faker
+
+from core import factories, models
+
+from demo import defaults
+
+fake = Faker()
+
+logger = logging.getLogger("impress.commands.demo.create_demo")
```
lunika:
```suggestion
logger = logging.getLogger(__name__)
```

env.d/development/common
```diff
@@ -73,3 +73,9 @@ WOPI_CLIENTS="collabora,onlyoffice"
 WOPI_COLLABORA_DISCOVERY_URL="http://collabora:9980/hosting/discovery"
 WOPI_ONLYOFFICE_DISCOVERY_URL=http://onlyoffice/hosting/discovery
 WOPI_SRC_BASE_URL=http://app-dev:8000
+
+# Indexer
+SEARCH_INDEXER_CLASS="core.services.search_indexers.SearchIndexer"
+SEARCH_INDEXER_SECRET=find-api-key-for-driv-with-exactly-50-chars-length   # Key generated by create_demo in Find app.
+SEARCH_INDEXER_URL="http://find:8000/api/v1.0/documents/index/"
+SEARCH_INDEXER_QUERY_URL="http://find:8000/api/v1.0/documents/search/"
```
lunika: Should be commented ?

src/backend/core/management/commands/index.py
```diff
@@ -0,0 +1,41 @@
...
+    """Index all files to remote search service"""
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        """Launch and log search index generation."""
+        indexer = get_file_indexer()
+
+        if not indexer:
+            raise CommandError("The indexer is not enabled or properly configured.")
```
lunika: If there is no indexer, only an info message should be printed informing that there is no indexer service configured ? I don't think it should be an error

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,351 @@
...
+        items (list of Item): List of items.
+
+    Returns:
+        dict[str, set[str]]: Mapping from ancestor path to its descendant paths (including itself).
+    """
+    ancestor_map = defaultdict(set)
+
+    for item in items:
+        path = str(item.path)
+        ancestors = path.split(".")
```
lunika: A path is already a list, you don't need to cast it in string and then split it.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,355 @@
...
+
+            if not items_batch:
+                break
+
+            last_id = items_batch[-1].id
+            accesses_by_item_path = get_batch_accesses_by_users_and_teams(items_batch)
+            serialize = self.serialize_item
+
+            serialized_batch = [
+                serialize(item, accesses_by_item_path) for item in items_batch
```
lunika:
```suggestion
                self.serialize_item(item, accesses_by_item_path) for item in items_batch
```

Why not using `self.serialize_item` ? I don't understand why you did that ?

lunika (commented):
Just for your information, the library used in Drive to manage the item tree is not the same from docs. We use django-ltree and use the postgresql ltree extension. This is a significant change between the two projects, changing how to make queries and how to work with the `path` parameter.

src/backend/core/services/search_indexers.py
joehybird: Old python optimisation habit to use local variables instead of attributes for simple loops. Not really useful here.

src/backend/core/services/search_indexers.py
joehybird: ok, thx for the tip :)

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,354 @@
...
+        Args:
+            document (Document): The document instance.
+            accesses (dict): Mapping of document ID to user/team access.
+
+        Returns:
+            dict: A JSON-serializable dictionary.
+        """
+        doc_path = str(item.path)
+        is_deleted = bool(item.ancestors_deleted_at or item.deleted_at)
+        content = self.to_text(item) if not is_deleted and self.has_text(item) else ""
```
lunika: I don't understand the logic here. If the item is deleted or it is not possible to have a text content, you will serialize it with an empty content ? Why you don't ignore it ?

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        Return True if the file mimetype can be converted into text for indexation
+        """
+        mimetype = item.mimetype or ""
+
+        return (
+            item.upload_state == models.ItemUploadStateChoices.READY
+            and item.type == models.ItemTypeChoices.FILE
+            and mimetype
+            and any(mimetype.startswith(allowed) for allowed in self.allowed_mimetypes)
```
lunika: This will not always work. If I want to index a specific mimetype, meaning not startig with `{family}/`, this will not work. Next document we will probably parse can be pdf and the mimetype is `application/pdf`. We will not allow `application/` but strictly `application/pdf`.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def index(self, queryset=None):
+        """
+        Fetch documents in batches, serialize them, and push to the search backend.
+        """
+        last_id = None
+        count = 0
+        queryset = queryset or models.Item.objects.all()
+        queryset = queryset.filter(
+            main_workspace=False,
```
lunika, line 205: Why files from a main workspace can't be indexed ?

src/backend/core/tasks/search.py
```diff
@@ -0,0 +1,110 @@
...
+    """Celery Task : Sends indexation query for a document."""
+    indexer = get_file_indexer()
+
+    if indexer is None:
+        return
+
+    try:
+        item = models.Item.objects.get(
+            pk=item_id,
+            main_workspace=False,
```
lunika, line 45: Why filtering on the main workspace ?

src/backend/core/tasks/search.py
```diff
@@ -0,0 +1,110 @@
...
+    except models.Item.DoesNotExist:
+        # Skip the task if the document does not exist.
+        return
+
+    accesses = get_batch_accesses_by_users_and_teams((item,))
+
+    data = indexer.serialize_item(item=item, accesses=accesses)
+
+    logger.info("Start file %s indexation", item_id)
+    indexer.push(data)
```
lunika: Why instead of executing all this steps not calling `indexed.index` with the adequate queryset ? `index` is responsible to do the same.

src/backend/core/services/search_indexers.py
joehybird: I have to mark a deleted file as `is_active=False` in the opensearch database. But we still have only one endpoint. I should add a comment to explain this.

src/backend/core/tasks/search.py
joehybird: You have a point here. I forgot to refactor this part after adding the queryset as an argument of `index()`

src/backend/core/api/viewsets.py
```diff
@@ -1005,18 +1011,63 @@ def breadcrumb(self, request, *args, **kwargs):
         serializer = self.get_serializer(breadcrumb, many=True)
         return drf.response.Response(serializer.data, status=drf.status.HTTP_200_OK)
 
+    # pylint: disable-next=too-many-arguments,too-many-positional-arguments
+    def _fulltext_search(self, queryset, indexer, request, text):
```
lunika: something more relevant ? `_external_search` ? This is not really related to full text search but using an external service managing search

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,388 @@
...
+            nb_results (int, optional):
+                The number of results to return per page.
+                Defaults to settings.SEARCH_INDEXER_QUERY_LIMIT.
+        """
+        nb_results = nb_results or self.search_limit
+        response = self.search_query(
+            data={
+                "q": text,
+                "visited": visited,
+                "services": ["drive"],
```
lunika: `drive` should be a constant

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,388 @@
...
+                "nb_results": nb_results,
+            },
+            token=token,
+        )
+
+        return [d["_id"] for d in response]
```
lunika: IMO this should not be part of this method. It is the class implementing this `BaseItemIndexer` class that know what the reponse contains. If we create an other backend and the response in the query does not have this `_id` property, it will not works. Probably the `search` method should be abstract and every backend should implement its own logic.

docs/setup-find.md
```diff
@@ -0,0 +1,45 @@
+# Setup the Find search for Drive
+
+This configuration will enable the fulltext search feature for Docs :
+- Each save on **core.Item** or **core.Item** will trigger the indexer
+- Once indexer service configured, the `api/v1.0/item/search/` will work as a proxy with the Find API for fulltext search.
+
+## Create an index service for Drive
+
+Configure a **Service** for Docs application with these settings
```
lunika:
```suggestion
Configure a **Service** for Drive application with these settings
```

docs/setup-find.md
```diff
@@ -0,0 +1,45 @@
...
+
+- **Name**: `drive`<br>_request.auth.name of the Docs application._
+- **Client id**: `drive`<br>_Name of the token audience or client_id of the Docs application._
+
+See [how-to-use-indexer.md](how-to-use-indexer.md) for details.
+
+## Configure settings of Drive
+
+Add those Django settings the Docs application to enable the feature.
```
lunika:
```suggestion
Add those Django settings the Drive application to enable the feature.
```

src/backend/core/api/viewsets.py
```diff
@@ -1005,6 +1011,43 @@ def breadcrumb(self, request, *args, **kwargs):
         serializer = self.get_serializer(breadcrumb, many=True)
         return drf.response.Response(serializer.data, status=drf.status.HTTP_200_OK)
 
+    # pylint: disable-next=too-many-arguments,too-many-positional-arguments
+    @method_decorator(refresh_oidc_access_token)
+    def _indexed_search(self, request, queryset, indexer, text):
+        """
+        Returns a queryset from the results the fulltext search of Find
```
lunika: It doesn't return a queryset but a response.

joehybird: Good point, forgot to change the doc :sweat_smile:

lunika:
@joehybird We added a security check tool. It spotted 2 security risks, can you look at them ? Thanks
Don'e spend time on the code duplication, it's only about the tests and it's legit to have code duplication at this place.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+
+    def to_text(self, item):
+        """
+        Convert a file content into an indexable text.
+        """
+        mimetype = item.mimetype or ""
+
+        if mimetype.startswith("text/"):
+            return default_storage.open(item.file_key, "rb").read().decode()
```
qbey: Don't we need to "close" after reading?

joehybird:
> @joehybird We added a security check tool. It spotted 2 security risks, can you look at them ? Thanks Don'e spend time on the code duplication, it's only about the tests and it's legit to have code duplication at this place.

One is in the unit test : a shuffle on a fake list of results to test the ordering by score. I think we can ignore it.
The other is a random in the create_demo command. Not used in production so I wonder :thinking:

src/backend/core/api/viewsets.py
```diff
@@ -1048,7 +1100,20 @@ def search(self, request, *args, **kwargs):
             path_list |= db.Q(path__descendants=top_level_item)
 
         queryset = queryset.filter(path_list)
+
+        if indexer:
```
qbey: It would be nice to be able to disable indexed search even if documents are indexed (to be able to start indexing document without actually use search on them).

=> Maybe open an issue if you don't want to do it in this PR.

qbey (approved):

Issue #352, opened 2025-09-24
Children of delete folders still appear in search


NathanVss:
## Bug Report

**Problematic behavior**
Children of delete folders still appear in search.

**Expected behavior/code**
They shouldn't.

**Steps to Reproduce**
1. Create folder A
2. Create folder A CHILD
3. Delete A
4. Search for "A", you got "A CHILD" found.

enansaif:
Hi there, I was just checking out the issue. I am sharing what I found.

Adding a condition to filter out items with deleted ancestors in the ```filter_scope``` function in ```backend/core/api/filters.py → SearchItemFilter```, helps exclude children of deleted folders from search results.

Before:
```
if ScopeChoices.DELETED in value:
    to_filter |= Q(deleted_at__isnull=False)
if ScopeChoices.NOT_DELETED in value:
    to_filter |= Q(deleted_at__isnull=True)
```
Proposed:
```
if ScopeChoices.DELETED in value:
    to_filter |= Q(deleted_at__isnull=False) | Q(ancestors_deleted_at__isnull=False)
if ScopeChoices.NOT_DELETED in value:
    to_filter |= Q(deleted_at__isnull=True, ancestors_deleted_at__isnull=True)
```
With this change, child items no longer appear in search results when their parent folder has been deleted (unless the location is set to the recycle bin).
However, I noticed that if the search location is set to recycle bin, the parent folder correctly shows it is in Trash, but the child folder still appears as part of its old workspace. This might need more investigation.

Does this solution look acceptable, and what would you suggest as the next steps? Thanks.

enansaif:
Hi again, I’ve explored this issue and already have a working solution locally.
Could I please be assigned to this issue so I can open a PR following the project’s contribution guidelines?

lunika:
Hi @enansaif 
You have been assigned to this issue. 
Thanks for your help and your contribution

lunika:
Managed in #393

PR #393, opened 2025-11-09
Fix/deleted child search

enansaif:
## Purpose
Fix issue where children of deleted folders still appear in search results. https://github.com/suitenumerique/drive/issues/352

## Proposal
Update `SearchItemFilter` -> `filter_scope` to exclude items whose ancestors are deleted.

- [x] Add condition to filter out items with `ancestors_deleted_at` not null.
- [x] Add test case to confirm children of deleted folders are excluded from normal search.
- [x] Add test case to confirm deleted children appear only when searching in recycle bin.

### DIFF
--- a/src/backend/core/api/filters.py
+++ b/src/backend/core/api/filters.py
@@ -102,9 +102,7 @@ class SearchItemFilter(ItemFilter):
         if ScopeChoices.ALL in value:
             return queryset
         if ScopeChoices.DELETED in value:
-            to_filter |= Q(deleted_at__isnull=False) | Q(
-                ancestors_deleted_at__isnull=False
-            )
+            to_filter |= Q(ancestors_deleted_at__isnull=False)
         if ScopeChoices.NOT_DELETED in value:
             to_filter |= Q(deleted_at__isnull=True, ancestors_deleted_at__isnull=True)
 


enansaif:
Requesting reviews @NathanVss @lunika. Thanks!

src/backend/core/api/filters.py
```diff
@@ -102,9 +102,11 @@ def filter_scope(self, queryset, name, value):
         if ScopeChoices.ALL in value:
             return queryset
         if ScopeChoices.DELETED in value:
-            to_filter |= Q(deleted_at__isnull=False)
+            to_filter |= Q(deleted_at__isnull=False) | Q(
+                ancestors_deleted_at__isnull=False
+            )
```
lunika:
```suggestion
            to_filter |= Q(ancestors_deleted_at__isnull=False)
```

Only relying on `ancestors_deleted_at` is enough here.

lunika:
And the body of your commits are too long. You should not exceed 80 characters.

enansaif:
Thank you for the feedback @lunika. I have removed the redundant check and shortend the commit messages.

lunika (approved):

lunika:
Thank you for your contribution

PR #417, opened 2025-11-24
🔥(docker) remove git in production layer

lunika:
## Purpose

In the production layer git is kept but we don't need it. We can remove it.

## Proposal

- [x] 🔥(docker) remove git in production layer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -144,6 +144,9 @@ FROM core AS backend-production
 
 ARG DRIVE_STATIC_ROOT=/data/static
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -134,6 +134,9 @@ FROM core AS backend-production
 
 ARG DRIVE_STATIC_ROOT=/data/static
 
+# Remobe git, we don't need it in the production image
```
qbey:
```suggestion
# Remove git, we don't need it in the production image
```

PR #441, opened 2025-11-28
✨(back) enhance items django admin

NathanVss:
We lack information on the item admin. Let's add some fields to allow filtering, and also display facets. Also added ability to update the upload_state in case we need to do some debugging.

Capture d’écran 2025-11-28 à 15 21 55


lunika (changes requested):
Pas de souci

PR #446, opened 2025-12-01
🐛(backend) do not allow renaming a file while not ready

lunika:
## Purpose

We don't want to allow renaming a file while its state is not ready. The malware_detection backend is using the file path to work, if we rename a file during the analysis then we can lead to an error like file does not exists.

## Proposal

- [x] 🐛(backend) do not allow renaming a file while not ready


lunika:
> no changelog? ^^

Added. Thanks

PR #453, opened 2025-12-04
✨(backend) add a command to update file_hash in malware_detection_info

lunika:
## Purpose

We want to update existing item in suspicious upload_state and to the malware_detection_info the file hash. This is an info we need to track suspicious files.

## Proposal

- [x] ✨(backend) add a command to update file_hash in malware_detection_info

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "django-cors-headers==4.9.0",
     "django-countries==8.2.0",
     "django-filter==25.2",
-    "django-lasuite[all]==0.0.20",
+    "django-lasuite[all]==0.0.21",
     "django-ltree@git+https://github.com/mariocesar/django-ltree@5d955bc82021a50c522ee524106f6709e3b414e4",
     "django-parler==2.3",
     "django-redis==6.0.0",
--- a/src/backend/uv.lock
+++ b/src/backend/uv.lock
@@ -497,7 +497,7 @@ wheels = [
 
 [[package]]
 name = "django-lasuite"
-version = "0.0.20"
+version = "0.0.21"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "django" },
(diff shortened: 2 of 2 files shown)


src/backend/core/management/commands/update_suspicious_item_file_hash.py
```diff
@@ -0,0 +1,27 @@
...
+            upload_state=ItemUploadStateChoices.SUSPICIOUS,
+            malware_detection_info__file_hash__isnull=True,
+        ).iterator()
+
+        for item in items:
+            self.stdout.write(
+                f"Triggering update suspicious item file hash for item {item.id}"
+            )
+            update_suspicious_item_file_hash.delay(item.id)
+        self.stdout.write("Update suspicious item file hash command completed")
```
NathanVss, line 27: Let's imagine for some reason we can't get the hash of a suspicious item ( remove by hand on the storage for instance ) -> the file would loop forever

src/backend/core/management/commands/update_suspicious_item_file_hash.py
```diff
@@ -0,0 +1,27 @@
+"""Update suspicious item file hash"""
+
+from django.core.management.base import BaseCommand
+
+from core.models import Item, ItemUploadStateChoices
+from core.tasks.item import update_suspicious_item_file_hash
+
+
+class Command(BaseCommand):
```
NathanVss, line 9: I don't see any cron job to start this command, what do you think regarding its execution ? One time exec by hand, cron, etc ?

src/backend/core/tasks/item.py
```diff
@@ -97,3 +98,28 @@ def rename_file(item_id, new_title):
...
+            "updating suspicious item file hash: Item %s does not exist", item_id
+        )
+        return
+    if item.upload_state != ItemUploadStateChoices.SUSPICIOUS:
+        logger.error(
+            "updating suspicious item file hash: Item %s is not suspicious", item_id
+        )
+        return
+    with default_storage.open(item.file_key, "rb") as file:
+        file_hash = hashlib.file_digest(file, "sha256").hexdigest()
```
NathanVss, line 122: Do you know if this handles well file with a size > ram allocation of the worker?

src/backend/core/management/commands/update_suspicious_item_file_hash.py
lunika: It is a one shot command. You should not rerun it with this PR on the django-lasuite project https://github.com/suitenumerique/django-lasuite/pull/51

src/backend/core/tasks/item.py
lunika: This is the same code used in the django-lasuite project and we didn't have for now problem with it. But our workers have far more RAM than the max upload file size authorized.

src/backend/core/management/commands/update_suspicious_item_file_hash.py
lunika: The same item will trigger the task every time the command is executed, yes.

PR #458, opened 2025-12-05
✨(backend) enable full customization of external api

NathanVss:
We have some instances that requires more open resource server features. This change adds a setting that allows to cherry-pick which api routes needs to be available or not.

### DIFF
--- a/docs/resource_server.md
+++ b/docs/resource_server.md
@@ -2,13 +2,15 @@
 
 Drive implements resource server, so it means it can be used from an external app to perform some operation using the dedicated API.
 
+> **Note:** This feature might be subject to future evolutions. The API endpoints, configuration options, and behavior may change in future versions.
+
 ## Prerequisites
 
 In order to activate the resource server on Drive you need to setup the following environement variables
 
 ```
 OIDC_RESOURCE_SERVER_ENABLED=True
-OIDC_OP_URL= 
+OIDC_OP_URL=
 OIDC_OP_INTROSPECTION_ENDPOINT=
 OIDC_RS_CLIENT_ID=
 OIDC_RS_CLIENT_SECRET=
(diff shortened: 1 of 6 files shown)


docs/resource_server.md
```diff
@@ -18,6 +18,37 @@ OIDC_RS_ALLOWED_AUDIENCES=
...
+    "item_invitation": {
+        "enabled": False,
+        "actions": [],
+    },
+}
+```
+
+**Endpoints:**
+
+- `items`: Controls `/external_api/v1.0/items/`. Available actions: `list`, `retrieve`, `create`, `update`, `partial_update`, `destroy`, `children`, `upload_ended`, `move`, `restore`, `trashbin`, `hard_delete`, `tree`, `breadcrumb`, `link_configuration`, `favorite`, `media_auth`, `wopi`
```
sylvaingi, line 48: @NathanVss is the `/users` endpoint implicitly enabled?

NathanVss: Yes it is, so you can get infos about the user associated with the access token

docs/resource_server.md
```diff
@@ -18,6 +20,37 @@ OIDC_RS_ALLOWED_AUDIENCES=
...
+
+Configure the `EXTERNAL_API` setting to control which routes and actions are available in the external API. Set it via the `EXTERNAL_API` environment variable (as JSON) or in Django settings.
+
+Default configuration:
+
+```python
+EXTERNAL_API = {
+    "items": {
+        "enabled": True,
+        "actions": ["list", "retrieve", "children", "upload_ended"],
```
lunika, line 33: Seeing `children` makes me think we can't choose between listing children and creating a new one. This is out of the scope of this PR but we should find a way to split this method.

src/backend/core/external_api/viewsets.py
```diff
@@ -1,11 +1,20 @@
...
-from core.api.viewsets import ItemViewSet, UserViewSet
+from core.api.viewsets import (
+    InvitationViewset,
+    ItemAccessViewSet,
+    ItemViewSet,
+    UserViewSet,
+)
 from core.external_api.permissions import ResourceServerClientPermission
 
+# pylint: disable=too-many-ancestors
```
lunika, line 16: Where do you have too many ancestors ?

src/backend/core/external_api/viewsets.py
```diff
@@ -14,7 +23,11 @@ class ResourceServerItemViewSet(ItemViewSet):
 
     permission_classes = [ResourceServerClientPermission & ItemAccessPermission]
 
-    resource_server_actions = ["list", "retrieve", "children", "upload_ended"]
+    @property
+    def resource_server_actions(self):
+        """Build resource_server_actions from settings."""
+        items_config = settings.EXTERNAL_API.get("items", {})
+        return list(items_config.get("actions", []))
```
lunika: You should probably create a private function in a mixin and call it here to avoid code repetition


```suggestion
    @property
    def resource_server_actions(self):
        """Build resource_server_actions from settings."""
        return self._get_resource_server_action("items")
        
     # Move this in a Mixin   
    def _get_resource_server_actions(self, resource_name):
        external_api_config = settings.EXTERNAL_API.get(resource_name, {})
        return list(external_api_config.get("actions", []))
```

docs/resource_server.md
NathanVss: Agreed +1

src/backend/core/external_api/viewsets.py
NathanVss: It's transitive up from ResourceServerItemAccessViewSet -> ItemAccessViewSet 😅

PR #465, opened 2025-12-09
✨(front) make auth urls and redirection configurable

mellingo:
## Purpose

Auth urls (login, logout and connectedUser) can be configured through Driver.
If unauthenticated, redirect to homepage_url from ThemeCustomization if present.


mellingo:
Hello! We already discussed this topic with @NathanVss and @PanchoutNathan. I hope you will be able to see this PR even though I cannot assign a reviewer

PR #468, opened 2025-12-15
🐛(email) avoid trying to send emails if no provider is configured

NathanVss:


src/backend/core/tests/test_models_items.py
```diff
@@ -639,6 +640,39 @@ def test_models_items__email_invitation__failed(mock_logger, _mock_send_mail):
     assert isinstance(exception, smtplib.SMTPException)
 
 
+@mock.patch("core.models.send_mail")
+@mock.patch.object(Logger, "debug")
```
lunika: You can use the `caplog` fixture instead of mocking the logger. It is made for that purpose. https://docs.pytest.org/en/stable/how-to/logging.html

src/backend/core/tests/test_models_items.py
```diff
@@ -639,6 +640,39 @@ def test_models_items__email_invitation__failed(mock_logger, _mock_send_mail):
...
+@override_settings(EMAIL_HOST=None)
```
lunika: Same here, use the settings fixture.
Probably we already use the same way you made in the project, but we should use the settings fixture instead.

src/backend/core/tests/tasks/items/test_rename_file.py
```diff
@@ -88,6 +88,5 @@ def test_rename_file_item_not_ready(caplog):
     )
     with caplog.at_level("ERROR", logger="core.tasks.item"):
         rename_file(item.id, "new_title")
-    item.refresh_from_db()
```
lunika, line 91: Why this change ?

NathanVss: Mistake. But the test was still passing fine 😬

src/backend/core/tests/test_models_items.py
```diff
@@ -12,6 +12,7 @@
 from django.core import mail
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
+from django.test import override_settings
```
lunika:
```suggestion
```

src/backend/core/tests/tasks/items/test_rename_file.py
lunika: Yes this is normal. I want to test that the title has not changed so for that I have to refresh the object and assert that the title is still the same.

PR #479, opened 2026-01-01
Remove N+1 query patterns on items children view

sampaccoud:
## Purpose

Our Drive instance is lacking snappiness and I noticed it was getting really slow with folders containing many items which is the symptom of unwanted N+1 query patterns on list views.

## Proposal

Remove N+1 query patterns on the items children list endpoint, which triggers 600+ database queries per request (3 extra queries  x 200 items returned).

- Avoid N+1 queries when serializing item creators by preloading the creator relation for all children.
  :point_right: @lunika do we really need the details of the creator on all list items? Returning just the ID would be even more efficient.

- Avoid N+1 queries when computing nb_accesses by precomputing the parent access count and annotating each child with its total number of accesses in the list query (parent count + direct count).

- Avoid N+1 queries when computing link reach and link role by precomputing ancestors link values once and passing it to serializers as a shared mapping.

This keeps the children view query count constant whatever the number of items returned (200 by default!) while preserving existing API behavior.

@NathanVss @PanchoutNathan for further snappiness optimization, I noticed that the frontend is making repeated queries to some list views. It doesn't seem normal and could use some cache or an optimized logic in the frontend?

That will be my first TGV pull request of the year :wink: HNY all! :tada:

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -540,16 +540,15 @@ class ItemViewSet(
         queryset = self.annotate_user_roles(queryset)
         return queryset
 
-    def get_response_for_queryset(self, queryset):
+    def get_response_for_queryset(self, queryset, context=None):
         """Return paginated response for the queryset if requested."""
-
+        context = context or self.get_serializer_context()
         page = self.paginate_queryset(queryset)
         if page is not None:
-            serializer = self.get_serializer(page, many=True)
-            result = self.get_paginated_response(serializer.data)
-            return result
+            serializer = self.get_serializer(page, many=True, context=context)
+            return self.get_paginated_response(serializer.data)
 
-        serializer = self.get_serializer(queryset, many=True)
+        serializer = self.get_serializer(queryset, many=True, context=context)
         return drf.response.Response(serializer.data)
 
     def retrieve(self, request, *args, **kwargs):
@@ -931,7 +930,17 @@ class ItemViewSet(
             + Coalesce(db.Count("accesses", distinct=True), 0),
         )
 
-        return self.get_response_for_queryset(queryset)
+        # Pass ancestors' links paths mapping to the serializer as a context variable
+        # in order to allow saving time while computing abilities on the instance
+        paths_links_mapping = item.compute_ancestors_links_paths_mapping()
+
+        return self.get_response_for_queryset(
+            queryset,
+            context={
+                "request": request,
+                "paths_links_mapping": paths_links_mapping,
+            },
+        )
 
     @drf.decorators.action(detail=True, methods=["get"])
     def tree(self, request, pk=None):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to
 - ✨(api) modify items/search endpoint to use indexed items in Find
 - 🐛(email) avoid trying to send emails if no provider is configured
 - ♻️(backend) improve mimetype detection
+- ♻️(backend) remove N+1 query patterns on items children view
 
 ### Fixed
 
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -605,6 +605,26 @@ class Item(TreeModel, BaseModel):
 
         return super().save(*args, **kwargs)
 
+    def compute_ancestors_links_paths_mapping(self):
+        """
+        Compute the ancestors links for the current document up to the highest readable ancestor.
+        """
+        ancestors = (
+            (self.ancestors() | self._meta.model.objects.filter(pk=self.pk))
+            .filter(ancestors_deleted_at__isnull=True)
+            .order_by("path")
+        )
+        ancestors_links = []
+        paths_links_mapping = {}
+
+        for ancestor in ancestors:
+            ancestors_links.append(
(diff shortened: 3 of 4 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -918,10 +919,18 @@ def children(self, request, *args, **kwargs):
...
         queryset = filters.OrderingFilter().filter_queryset(
             self.request, queryset, self
         )
+
+        # Pre-compute number of accesses
+        item_nb_accesses = item.nb_accesses
+        queryset = queryset.annotate(
+            _nb_accesses=db.Value(item_nb_accesses)
+            + Coalesce(db.Count("accesses", distinct=True), 0),
+        )
```
NathanVss, line 931: With the new pattern, each time we query the children we perform a count on each of them whether we already have a cache value stored already. 

For the first /children request, we indeed avoid the N+1 query issue, but regarding the next ones before we were relying on cache, and now everytime time the DB has to perform a count on children accesses. We never rely on cache anymore with this approach, WYDT?

src/backend/core/api/viewsets.py
```diff
@@ -918,10 +919,18 @@ def children(self, request, *args, **kwargs):
...
+        # Apply ordering only now that everything is filtered and annotated
         queryset = filters.OrderingFilter().filter_queryset(
             self.request, queryset, self
         )
+
+        # Pre-compute number of accesses
+        item_nb_accesses = item.nb_accesses
+        queryset = queryset.annotate(
+            _nb_accesses=db.Value(item_nb_accesses)
+            + Coalesce(db.Count("accesses", distinct=True), 0),
```
lunika, line 930: IMO this is wrong. `item.nb_accesses` is doing a count on all item+ancestors accesses. `db.Count("accesses")` will only return the direct access on the item.

src/backend/core/api/viewsets.py
```diff
@@ -541,16 +541,15 @@ def filter_queryset(self, queryset):
         queryset = self.annotate_user_roles(queryset)
         return queryset
 
-    def get_response_for_queryset(self, queryset):
+    def get_response_for_queryset(self, queryset, context=None):
```
lunika, line 543: IMO this commit will conflict with what is made in PR #383, we are doing something similar.

src/backend/core/api/viewsets.py
sampaccoud: In fact, the cache is only interesting to avoid going up the ancestors' chain. In this case, we have to retrieve nb_accesses for N leaf items. So even if all of these leaves already have a cache (which is far from being sure...) you would still make N requests to Redis which are much more expensive than the "internal" count I do. 
The other thing to understand is that there should be very few accesses on a leaf item... as accesses are most likely configured at the folder level. So my counts are super cheap and select on the primary key which has an index.
Finally, doing this will avoid using cache keys for leaf items which is good because the number of files is huge and our cache is limited. It's good to avoid rotation on other cache keys.

sampaccoud:
> IMO we should keep the first commit about the creator. We should talk about the second one, you are saving queries but removing cache. What should we really do ? And the third one is managed in PR #383

did you see my question about the creator in PR description? I didn't find the use cas for including the creator's details in the serializer.

src/backend/core/api/viewsets.py
sampaccoud: yes but I'm adding `item_nb_accesses` to it which accounts for all accesses on its ancestors. Look at the test I wrote which involves a grand parent, a parent, an item and some children. Maybe I'm missing something but the counts in this test are random so it would fail.

src/backend/core/api/viewsets.py
sampaccoud: Whichever merges first then :wink:

lunika:
> > IMO we should keep the first commit about the creator. We should talk about the second one, you are saving queries but removing cache. What should we really do ? And the third one is managed in PR #383
> 
> did you see my question about the creator in PR description? I didn't find the use cas for including the creator's details in the serializer.

The creators details are used by the frontend application when you click on the info picto of an item

image

sampaccoud:
> > > IMO we should keep the first commit about the creator. We should talk about the second one, you are saving queries but removing cache. What should we really do ? And the third one is managed in PR #383
> > 
> > 
> > did you see my question about the creator in PR description? I didn't find the use cas for including the creator's details in the serializer.
> 
> The creators details are used by the frontend application when you click on the info picto of an item

Since it's behind a click, it could rely on a call to the detail endpoint and we can free a lot of resources on list views which are 90% of the queries in Drive?

lunika:
Yes but we have to expose an API endpoint for retrieving users like `/api/v1.0/users/{id}` and it is not something we want IMO.

src/backend/core/api/viewsets.py
NathanVss: The point about redis queries makes sense indeed, not sure if outperform count in DB instead, we may want to measure that. But after all, why not merge the two approach and store the count in the DB as a cache value on each row?

lunika: > store the count in the DB as a cache value on each row?

Because it's a nightmare to maintain and error prone

PR #480, opened 2026-01-01
Create ItemAccess when children are created via the API

NielsCodes:
## Purpose

When children of a directory are created via the `POST /api/v1.0/items/{id}/children/` endpoint, no `ItemAccess` is created. This is inconsistent with the `POST /api/v1.0/items/` behaviour, where the `ItemAccess` **is** created.

Missing the `ItemAccess` of course causes issues down the line, for instance the user not being able to delete objects they just created.

## Proposal
Make creation of child items consistent with the creation of non-child items, by creating an `ItemAccess` record.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -898,6 +898,11 @@ class ItemViewSet(
                     parent=item,
                     **serializer.validated_data,
                 )
+                models.ItemAccess.objects.create(
+                    item=child_item,
+                    user=request.user,
+                    role=models.RoleChoices.OWNER,
+                )
 
             # Set the created instance to the serializer
             serializer.instance = child_item


lunika:
This behavior is intended. The permissions (accesses) are inherited, there is no need to create a new access on the newly created item.
Permission will be fully managed and more fined grain in #383

NielsCodes:
@lunika let me give you a bit more context. We're developing a desktop tool that allows bi-directional syncing between Drive and the local filesystem. This means we need to be able to perform all CRUD operations via the API. That's true for both top-level and nested files. 

Here is the problem:

✅ Creating an item with the `/items` endpoint allows that same item to be deleted
❌ Creating an item with the `/children` endpoint returns a 403 when trying to delete said item

Both of these actions are performed on a Workspace that is owned by the current user. Therefore this seems like inconsistent behaviour to me. 

Knowing this, do you think that the PR you mentioned will enable us to finish our use case?

lunika:
Can you reproduce using the Drive frontend ? 

Accesses are inherited, that means you will have the exact same access as your parent unless you have an explicit access created on the item.
Adding an explicit access when creating a child breaks this inheritance.
The PR I mention implement a complete scenario but does not change the way accesses are computed. Right now we are looking for the role of the highest ancestor the current user has access. Roles and user abilities are computed here https://github.com/suitenumerique/drive/blob/main/src/backend/core/models.py#L734 And this function is used in the permission to know if the current user can do the action. In your case it should be `destroy`.

Can you add a test to help me to understand ans reproduce your error ?

PR #481, opened 2026-01-05
♻️(backend) improve mimetype detection

lunika:
## Purpose

The mimetype detection were sometimes not accurate and return a wrong indication. For example, PPTX files were identified as an archive because they are zip files. This version relies on the file extension is the guessed MIME type is not relevant.

## Proposal

- [x] ♻️(backend) improve mimetype detection

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/utils/mimeTypes.ts
+++ b/src/frontend/apps/drive/src/features/explorer/utils/mimeTypes.ts
@@ -218,11 +218,7 @@ export const getMimeCategory = (
 ): MimeCategory => {
   // Special case: some calc files have application/zip mimetype. For those we should check their extension too.
   // Otherwise they will be shown as zip files.
-  if (
-    mimetype === "application/zip" &&
-    extension &&
-    CALC_EXTENSIONS.includes(extension)
-  ) {
+  if (extension && CALC_EXTENSIONS.includes(extension)) {
     return MimeCategory.CALC;
   }
 


lunika:
@coderabbitai full review

src/backend/core/api/utils.py
```diff
@@ -156,3 +158,58 @@ def get_item_file_head_object(item):
...
+    # If content detection gives us a generic type and we have extension info
+    if mimetype_from_content in generic_types and mimetype_from_extension:
+        # Use extension-based detection if it's more specific
+        if mimetype_from_extension not in generic_types:
+            return mimetype_from_extension
+
+    # If content detection failed or returned None, use extension
+    if not mimetype_from_content or mimetype_from_content in generic_types:
+        if mimetype_from_extension:
+            return mimetype_from_extension
```
NathanVss: Wouldn't this make more sense as we already test for `mimetype_from_content in generic_types` in the condition above?

```suggestion
    # If content detection failed or returned None, use extension
    if not mimetype_from_content:
        if mimetype_from_extension:
            return mimetype_from_extension
```

NathanVss (commented):
It seems that with this update we will be able to simplify the frontend code that does that: https://github.com/suitenumerique/drive/blob/691599af70f8dde3870addccdf5a1a021807d4bc/src/frontend/apps/drive/src/features/explorer/utils/mimeTypes.ts#L221-L232

Let me add a commit on top of that

src/backend/core/api/utils.py
```diff
@@ -156,3 +158,58 @@ def get_item_file_head_object(item):
...
+        "text/plain",
+    }
+
+    # If content detection gives us a generic type and we have extension info
+    if mimetype_from_content in generic_types and mimetype_from_extension:
+        # Use extension-based detection if it's more specific
+        if mimetype_from_extension not in generic_types:
+            return mimetype_from_extension
+
+    # If content detection failed or returned None, use extension
```
NathanVss:
```suggestion
# If content detection failed or returned None or is generic, use extension if it exists
```

PanchoutNathan (approved):

PR #486, opened 2026-01-06
✨(backend) Restrict file upload based on file type

lunika:
## Purpose

We want to limit what type of file can be uploaded. For this we combine two condition. The first one is based on the extension and the second is based on the mimetype. If one fails, the upload is rejected.

## Proposal

- [x] ✨(backend) add an allowed file extension list
- [x] ♻️(backend) delete file for not ready items
- [x] ✨(backend) add an allowed file mimetype list

### DIFF
--- a/src/backend/uv.lock
+++ b/src/backend/uv.lock
@@ -1740,11 +1740,11 @@ wheels = [
 
 [[package]]
 name = "urllib3"
-version = "2.6.1"
+version = "2.6.3"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


src/backend/core/api/serializers.py
```diff
@@ -429,6 +430,14 @@ def validate(self, attrs):
                     code="item_create_file_filename_required",
                 )
 
+            if settings.RESTRICT_FILE_EXTENSIONS:
+                _root, extension = splitext(attrs["filename"])
```
NathanVss, line 437: Shouldn't why rely on guessed mimetype? Otherwise I can upload a exe file that has been rename to toto.png

lunika: We should rely on both. I will add new commits for this.

NathanVss (approved):

PR #492, opened 2026-01-07
✨(front) add download and preview events

### LINKED ISSUE #491: Add posthog events to track file opening and download

NathanVss:
We need to provide more tracking events on posthog.

Closes #491

### DIFF
--- a/src/frontend/apps/drive/src/features/items/hooks/useDownloadItem.tsx
+++ b/src/frontend/apps/drive/src/features/items/hooks/useDownloadItem.tsx
@@ -7,6 +7,7 @@ import {
   addToast,
   ToasterItem,
 } from "@/features/ui/components/toaster/Toaster";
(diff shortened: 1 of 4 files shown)


src/frontend/apps/drive/src/features/items/hooks/useDownloadItem.tsx
```diff
@@ -46,6 +47,16 @@ export const useDownloadItem = () => {
       description = t("file_download_modal.description");
     }
 
+
+    const triggerDownload = () => {
+      posthog.capture("file_download", {
```
lunika, line 52: If PostHog is not configured, will it fail ?

NathanVss: Nope it will not fail

PR #496, opened 2026-01-14
✨(global) create new files directly from Drive

NathanVss:
## Purpose

We wan to allow creating files directly from drive.

Capture d’écran 2026-01-14 à 15 24 12

## Todo

- [x] Tests
- [x] Tidy up commits
- [x] Once rebased with advanced share, we must allow template creation from /items/from-template too

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-drive/create-file-from-template.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-drive/create-file-from-template.spec.ts
@@ -1,0 +1,91 @@
+import test, { expect } from "@playwright/test";
+import { clearDb, login } from "./utils-common";
+import { expectRowItem } from "./utils-embedded-grid";
+import { createFolderInCurrentFolder } from "./utils-item";
+import { navigateToFolder } from "./utils-navigate";
+
+test.describe("Create file from template", () => {
+  test.beforeEach(async ({ page }) => {
+    await clearDb();
+    await login(page, "drive@example.com");
+    await page.goto("/");
+  });
+
+  test("Create a text document (odt)", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New text document" }).click();
+
+    await page.getByRole("textbox", { name: "File name" }).fill("My document");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My document");
+  });
+
+  test("Create a spreadsheet (ods)", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New spreadsheet" }).click();
+
+    await page
+      .getByRole("textbox", { name: "File name" })
+      .fill("My spreadsheet");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My spreadsheet");
+  });
+
+  test("Create a presentation (odp)", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New slides" }).click();
+
+    await page
+      .getByRole("textbox", { name: "File name" })
+      .fill("My presentation");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My presentation");
+  });
+});
+
+test.describe("Create file from template in a folder", () => {
+  test.beforeEach(async ({ page }) => {
+    await clearDb();
+    await login(page, "drive@example.com");
+    await page.goto("/");
+    await createFolderInCurrentFolder(page, "Test folder");
+    await navigateToFolder(page, "Test folder", ["My files", "Test folder"]);
+  });
+
+  test("Create a text document (odt) in a folder", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New text document" }).click();
+
+    await page.getByRole("textbox", { name: "File name" }).fill("My document");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My document");
+  });
+
+  test("Create a spreadsheet (ods) in a folder", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New spreadsheet" }).click();
+
+    await page
+      .getByRole("textbox", { name: "File name" })
+      .fill("My spreadsheet");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My spreadsheet");
+  });
+
+  test("Create a presentation (odp) in a folder", async ({ page }) => {
+    await page.getByRole("button", { name: "New" }).click();
+    await page.getByRole("menuitem", { name: "New slides" }).click();
+
+    await page
+      .getByRole("textbox", { name: "File name" })
+      .fill("My presentation");
+    await page.getByRole("button", { name: "Create" }).click();
+
+    await expectRowItem(page, "My presentation");
+  });
+});


NathanVss:
@coderabbitai full review

src/backend/wopi/viewsets.py
```diff
@@ -154,18 +153,17 @@ def _put_file_content(self, request, pk=None):
...
-
         if lock_value:
             lock_service = LockService(item)
             current_lock_value = lock_service.get_lock(default="")
             if current_lock_value != lock_value:
                 return Response(status=409, headers={X_WOPI_LOCK: current_lock_value})
         else:
             # Check if the body is 0 bytes
-            body_size = int(request.META.get("CONTENT_LENGTH") or 0)
+            body_size = item.size or 0
```
lunika, line 163: Hmmm, no, you should not change this. We are checking the size of the request. Not the size of the item.

src/backend/core/api/serializers.py
```diff
@@ -476,7 +478,49 @@ def get_numchild_folder(self, _item):
...
 
     def update(self, instance, validated_data):
-        raise NotImplementedError("Update method can not be used.")
+        raise NotImplementedError("Update method can not be used.")        
+
+
+class CreateItemFromTemplateSerializer(serializers.Serializer):
+    """Serializer used to create a new item from a template file"""
+
+    extension = serializers.CharField(
```
lunika: I would suggest to use a `ChoiceField` here with the exact list of allowed extensions. Using this field will allow you to remove the `validate_extension`.

src/backend/core/api/serializers.py
```diff
@@ -476,7 +478,49 @@ def get_numchild_folder(self, _item):
...
+        max_length=10,
+        required=True,
+        help_text="File extension (e.g., 'odt', 'ods', ...)",
+    )
+    title = serializers.CharField(
+        max_length=255,
+        required=True,
+        allow_blank=True,
```
```suggestion
```

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+            )
+            raise drf.exceptions.ValidationError(
+                {"extension": _("Error reading template file.")},
+                code="template_file_read_error",
+            )
+
+        # If the title already has the extension, use it, otherwise add it
+        if title.endswith(f".{extension}"):
+            filename = title
+        else:
```
lunika:
```suggestion
```

You should not allow changing the extension. The extension must be one of the extensions we allow, not one set by the end user.

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+        except Exception as e:
+            logger.error(
+                "Error uploading template file to storage for item %s: %s",
+                child_item.id,
+                str(e),
+            )
+            # Delete created item
+            child_item.soft_delete()
+            child_item.hard_delete()
+            process_item_deletion.delay(child_item.id)
```
lunika:
```suggestion
            self._complete_item_deletion(child)
```

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
             },
         )
 
+    @drf.decorators.action(
+        detail=True,
+        methods=["post"],
+        url_path="children/from-template",
+    )
+    @transaction.atomic
```
lunika: Why do you need this ?

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+        """
+
+        item = self.get_object()
+
+        serializer = serializers.CreateItemFromTemplateSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        validated_data = serializer.validated_data
+
+        extension = validated_data["extension"]
+        title = validated_data.get("title")
```
lunika: You are mixing how you access an item in `validated_data`. You should choose one and use it for both case. 

Moreover I think that the declaration of `validated_data` is useless and you can use here `serializer.validated_data` directly

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+
+        The template file is read from core/file_templates/template.{extension}
+        and automatically uploaded to storage.
+        """
+
+        item = self.get_object()
+
+        serializer = serializers.CreateItemFromTemplateSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        validated_data = serializer.validated_data
```
lunika:
```suggestion
```

Useless, you can use directly `serializer.validated_data`

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+            raise drf.exceptions.ValidationError(
+                {"detail": _("Error uploading file to storage.")},
+                code="storage_upload_error",
+            )
+
+        mimetype, file_size = child_item.detect_mimetype()
+        child_item.upload_state = models.ItemUploadStateChoices.READY
+        child_item.mimetype = mimetype
+        child_item.size = file_size
+        child_item.save(update_fields=["upload_state", "mimetype", "size"])
```
lunika: IMO you can have all these infos without using the s3 head operation and reading the first file bytes.
You already know the filesize (it is `len(template_content)`) and the mimetype can be directly guss from the extension, this parameter is totally controlled by the API.

src/backend/core/api/viewsets.py
```diff
@@ -961,6 +945,116 @@ def children(self, request, *args, **kwargs):
...
+
+        # Serialize and return the created item
+        response_serializer = serializers.CreateItemSerializer(
```
lunika: You should use the `ItemSerializer` here. The `CreateItemSerializer` give an upload url, you don't need it here.

src/backend/core/models.py
```diff
@@ -1062,6 +1063,44 @@ def move(self, target, ignore_parent_numchild_update=False):
...
+        else:
+            file_head = s3_client.get_object(
+                Bucket=default_storage.bucket_name, Key=self.file_key
+            )["Body"].read()
+
+        # Use improved MIME type detection combining magic bytes and file extension
+        # Import is here to avoid circular import
+        from core.api import utils
+        mimetype = utils.detect_mimetype(file_head, filename=self.filename)        
+        return (mimetype, file_size)
```
lunika: What you return is not accurate with the method name. Why `detect_mimetype` returns a tuple containing the file size ?

src/backend/core/models.py
```diff
@@ -1062,6 +1063,44 @@ def move(self, target, ignore_parent_numchild_update=False):
...
 
+    def detect_file_size(self):
+        s3_client = default_storage.connection.meta.client
+        head_response = s3_client.head_object(
+            Bucket=default_storage.bucket_name, Key=self.file_key
+        )
+        file_size = head_response["ContentLength"]
+        return file_size
+
+    def detect_mimetype(self):
```
lunika: IMO both method should not live in the model, I would put them in an `utils` module.

src/backend/core/models.py
```diff
@@ -1062,6 +1063,44 @@ def move(self, target, ignore_parent_numchild_update=False):
...
+                Bucket=default_storage.bucket_name, Key=self.file_key
+            )["Body"].read()
+
+        # Use improved MIME type detection combining magic bytes and file extension
+        # Import is here to avoid circular import
+        from core.api import utils
+        mimetype = utils.detect_mimetype(file_head, filename=self.filename)        
+        return (mimetype, file_size)
+
+    def refresh_mimetype(self):
```
lunika: Seemd unused, you should probably remove it.

src/backend/wopi/viewsets.py
NathanVss: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putfile

`When a host receives a PutFile request on a file that's not locked, the host checks the current size of the file. If it's 0 bytes, the PutFile request should be considered valid and should proceed.`

For my reading, it's about the current size = the size before the PutFile operation. Which can make sense in a way any wopi editor can initialize a file ( an empty file is free to write ). 

Otherwise it would mean any wopi editor without lock could erase any existing file, which do not make sense to me

src/backend/core/api/viewsets.py
NathanVss: If something fails after create_child ( e.g. during mimetype detection ) we will not have dangling uncomplete item left. WDYT?

src/backend/core/api/viewsets.py
NathanVss: Indeed!!

src/backend/core/api/serializers.py
```diff
@@ -479,6 +481,43 @@ def update(self, instance, validated_data):
...
+
+        if not os.path.exists(template_path):
+            raise serializers.ValidationError(
+                _("Template file not found for extension '{extension}'.").format(
+                    extension=value
+                ),
+                code="template_file_not_found",
+            )
+
+        return value
```
lunika: Still don't understand why you continue to do this validation. 
You are testing again in the viewset if the file exists.

src/backend/core/api/utils.py
```diff
@@ -219,3 +213,51 @@ def detect_mimetype(file_buffer: bytes, filename: str | None = None) -> str:
...
+            Range="bytes=0-2047",
+        )
+        file_head = range_response["Body"].read()
+    else:
+        file_head = s3_client.get_object(
+            Bucket=default_storage.bucket_name, Key=item.file_key
+        )["Body"].read()
+
+    mimetype = detect_mimetype(file_head, filename=item.filename)
+    return (mimetype, file_size)
```
lunika:
```
```

As this is not used finally I suggest to remove this. Also because in the `upload_ended` method the head_response is now used for other things than having the file size.

src/backend/core/api/viewsets.py
```diff
@@ -689,28 +693,7 @@ def upload_ended(self, request, *args, **kwargs):
...
-            file_head = range_response["Body"].read()
-        else:
-            file_head = s3_client.get_object(
-                Bucket=default_storage.bucket_name, Key=item.file_key
-            )["Body"].read()
-
-        # Use improved MIME type detection combining magic bytes and file extension
-        logger.info("upload_ended: detecting mimetype for file: %s", item.file_key)
-        mimetype = utils.detect_mimetype(file_head, filename=item.filename)
+        mimetype, file_size = utils.detect_item_mimetype_and_file_size(item)
```
lunika: You should rollback. `head_response` is used in the `main` branch.

src/backend/core/api/viewsets.py
```diff
@@ -990,6 +945,104 @@ def children(self, request, *args, **kwargs):
...
+        # Upload template file to storage
+        try:
+            default_storage.save(child_item.file_key, BytesIO(template_content))
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.error(
+                "Error uploading template file to storage for item %s: %s",
+                child_item.id,
+                str(e),
+            )
+            self._complete_item_deletion(child_item)
```
lunika: I think this will fail. The `file` probably does not exist on the storage. It will not directly fail during the request process but after once celery will process the deletion

src/backend/core/api/viewsets.py
```diff
@@ -990,6 +945,104 @@ def children(self, request, *args, **kwargs):
...
+            raise drf.exceptions.ValidationError(
+                {"detail": _("Error uploading file to storage.")},
+                code="storage_upload_error",
+            ) from e
+
+        child_item.upload_state = models.ItemUploadStateChoices.READY
+        child_item.mimetype = utils.detect_mimetype(template_content, filename)
+        child_item.size = len(template_content)
+        child_item.save(update_fields=["upload_state", "mimetype", "size"])
```
lunika:
```suggestion
        child_item.save(update_fields=["upload_state", "mimetype", "size", "updated_at])
```

Maybe not a hard requirement, but `updated_at` will not be updated

src/backend/core/api/viewsets.py
```diff
@@ -990,6 +945,104 @@ def children(self, request, *args, **kwargs):
...
+            raise drf.exceptions.PermissionDenied(
+                detail=can_upload.get(
+                    "message", "You do not have permission to upload files."
+                )
+            )
+
+        # Get template file path
+        core_app_config = apps.get_app_config("core")
+        template_dir = os.path.join(core_app_config.path, "file_templates")
+        template_path = os.path.join(template_dir, f"template.{extension}")
```
lunika: Merge both in one `os.path.join` call ?
```suggestion
        template_dir = os.path.join(core_app_config.path, "file_templates",  f"template.{extension}")
```

lunika: And having them in an `assets` directory, maybe in the `drive` app should be better (`src/backend/drive/assets/file_templates`).

lunika: Or a better idea, and probably overkill but interesting. You define a setting where the templates should live; this allows you to put them outside the application and replace them by using a volume, for example.

src/backend/core/api/viewsets.py
```diff
@@ -990,6 +945,104 @@ def children(self, request, *args, **kwargs):
...
+        except OSError as e:
+            logger.error(
+                "Error reading template file %s: %s",
+                template_path,
+                str(e),
+            )
+            raise drf.exceptions.ValidationError(
+                {"extension": _("Error reading template file.")},
+                code="template_file_read_error",
+            ) from e
```
lunika: `FileNotFoundError` inherit from `OSError` and providing the information to the client that the file does not exist is 
- Maybe not useful, he can take no action on this.
- Potentially provide information that can be used to determine something by an attacker

I suggest to keep only the `OSError` except block

src/backend/core/api/viewsets.py
```diff
@@ -990,6 +945,104 @@ def children(self, request, *args, **kwargs):
...
+            )
+            self._complete_item_deletion(child_item)
+            raise drf.exceptions.ValidationError(
+                {"detail": _("Error uploading file to storage.")},
+                code="storage_upload_error",
+            ) from e
+
+        child_item.upload_state = models.ItemUploadStateChoices.READY
+        child_item.mimetype = utils.detect_mimetype(template_content, filename)
+        child_item.size = len(template_content)
```
lunika: Why not doing this in the `create_child` call ? Of the `default_storage.save` is failing, you are forced to delete everything.

lunika:
I had an idea this morning.
Instead of adding a new action, a new serializer, etc. Why not adding the ChoiceField in the `CreateItemSerializer`, this field is not required. 
You add a private method in the viewset creating the file.
In bot `perform_create` and `children` method, if in the validated_data from the serializer, the `extension` field is not null, then you call this private method. And boom ça fait des chocapics.

src/backend/core/api/viewsets.py
NathanVss: ok for the `assets` dir. Let's post pone the folder based on a settings in a separate PR when needed, we'll need to dynamically load the available extension etc ( backend for validation and on frontend for the suggested file type, in this case we'll need icons etc etc )

lunika (approved):

lunika:
Add an entry in the changelog please :-)

PanchoutNathan (dismissed):
## Frontend Review

Overall, the frontend changes are well-structured. The approach is clean: new entries in the "New" dropdown, a modal for the file name, and an API call to create the file from a template. Here are some observations:

src/frontend/apps/drive/src/features/explorer/hooks/useMutations.ts
```diff
@@ -39,6 +39,32 @@ export const useMutationCreateFile = () => {
   });
 };
 
+export const useMutationCreateFileFromTemplate = () => {
+  const driver = getDriver();
+  const queryClient = useQueryClient();
```
PanchoutNathan: 🔴 **Cache invalidation inconsistency**

`useMutationCreateFileFromTemplate` uses `queryClient.invalidateQueries` directly with `["items", variables.parentId]`, while `useMutationCreateFile` (line 27) uses the `useRefreshQueryCacheAfterMutation` helper which invalidates **both** `["items", "infinite"]` **and** `["items", parentId, "children"]`.

**Consequence**: after creating a file from a template, the `["items", "infinite"]` list (top-level "My files" view) won't be refreshed. If the user creates a file then navigates to a view using that query, they won't see it until the cache expires.

**Suggestion**: use `useRefreshQueryCacheAfterMutation` like `useMutationCreateFile` does:

```ts
export const useMutationCreateFileFromTemplate = () => {
  const driver = getDriver();
  const refresh = useRefreshQueryCacheAfterMutation();
  return useMutation({
    mutationFn: async (
      ...payload: Parameters<typeof driver.createFileFromTemplate>
    ) => {
      return driver.createFileFromTemplate(...payload);
    },
    onSuccess: (data, variables) => {
      refresh(variables.parentId);
    },
    meta: {
      showErrorOn403: true,
    },
  });
};
```

src/frontend/apps/drive/src/features/explorer/components/modals/ExplorerCreateFileModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return "odp";
+    case ExplorerCreateFileType.CALC:
+      return "ods";
+  }
+};
+
+export const ExplorerCreateFileModal = (
+  props: Pick<ModalProps, "isOpen" | "onClose"> & {
+    parentId: string;
+    type: ExplorerCreateFileType;
```
PanchoutNathan, line 36: 🔴 **`parentId` is required here but optional on `ExplorerCreateFolderModal`**

This declares `parentId: string` (required) while `ExplorerCreateFolderModal` declares `parentId?: string` (optional). However in `ExplorerTree.tsx:199`, `itemId` (which can be `undefined`) is passed as `parentId` to this component.

Also, `StandardDriver.createFileFromTemplate` already handles the falsy `parentId` case (`data.parentId ? ... : 'items/'`), so the type should reflect that:

```ts
parentId?: string;
```

What's strange is that today you **can** create children at root level (the flow works end-to-end), yet TypeScript doesn't flag this type mismatch — `itemId` is presumably typed as `string | undefined` and is being passed to a prop typed as `string`. This should be investigated: either the `itemId` type is wrong somewhere, or TypeScript is not catching this due to the spread pattern (`{...createFileModal}` might be obscuring it).

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeActions.tsx
```diff
@@ -30,9 +47,46 @@ export const ExplorerTreeActions = ({
...
-              isHidden: !showMenu,
+              isHidden: !canCreateChildren,
               callback: openCreateFolderModal,
             },
+            {
+              icon: renderFileIcon({
+                type: ItemType.FILE,
+                filename: "doc.odt",
+                mimetype:
+                  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
```
PanchoutNathan, line 58: 🟡 **Misleading mimetypes**

Microsoft Office mimetypes are used here (`application/vnd.openxmlformats-officedocument...`) while the files actually created are **ODF** format (`.odt`). It works visually because both formats fall into the same icon category via `MIME_MAP`, but it's conceptually wrong and fragile. Same issue on the two other entries below (`.odp` and `.ods`).

**Suggestion**: use the actual ODF mimetypes (which this very PR adds to `mimeTypes.ts`!):
- `application/vnd.oasis.opendocument.text` for `.odt`
- `application/vnd.oasis.opendocument.spreadsheet` for `.ods`
- `application/vnd.oasis.opendocument.presentation` for `.odp`

src/frontend/apps/drive/src/features/explorer/components/modals/ExplorerCreateFileModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    createFileFromTemplate.mutate(
+      {
+        parentId: props.parentId,
+        extension: extension,
+        title: data.filename,
+      },
+      {
+        onSuccess: () => {
+          form.reset();
```
PanchoutNathan, line 54: 🟡 **Form doesn't reset on close without submission**

`form.reset()` is only called here in `onSuccess`. If the user opens the modal, types a name, then closes without submitting (Cancel button or clicking outside), the form retains the previous value on the next open.

**Suggestion**: wrap `onClose` to also reset:
```tsx
const handleClose = () => {
  form.reset();
  props.onClose();
};
```
Then use `handleClose` on the Cancel button and pass it as `onClose` to the Modal.

src/frontend/apps/drive/src/features/explorer/components/modals/ExplorerCreateFileModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <form
+          onSubmit={form.handleSubmit(onSubmit)}
+          id="create-file-form"
+          className="mt-s"
+        >
+          <RhfInput
+            label={t("explorer.actions.createFile.modal.label")}
+            fullWidth={true}
+            autoFocus={true}
+            {...form.register("filename")}
```
PanchoutNathan, line 87: 🟢 (nit) No client-side validation — the user can submit an empty name. The backend will reject it, but UX would be better with at least:
```ts
{...form.register("filename", { required: true })}
```
(Consistent with `ExplorerCreateFolderModal` which also lacks it, so not blocking.)

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeActions.tsx
```diff
@@ -30,9 +47,46 @@ export const ExplorerTreeActions = ({
               icon: <img src={createFolderSvg.src} alt="" />,
               label: t("explorer.tree.create.folder"),
               value: "info",
```
PanchoutNathan: 🟢 (nit) All four dropdown options have `value: "info"`. If `DropdownMenu` uses `value` as a unique key, this could cause issues. Consider distinct values: `"create-folder"`, `"create-doc"`, `"create-slides"`, `"create-calc"`.

src/frontend/apps/drive/src/features/explorer/components/modals/ExplorerCreateFileModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+        },
+      }
+    );
+  };
+
+  return (
+    <Modal
+      {...props}
+      size={ModalSize.SMALL}
+      title={t("explorer.actions.createFile.modal.title")}
```
PanchoutNathan: 🟢 (nit) The modal title is always "Create file" regardless of the type (doc, spreadsheet, presentation). A dynamic title per type would help users confirm they clicked the right option.

src/frontend/apps/drive/src/features/explorer/components/icons/ItemIcon.scss
```diff
@@ -16,8 +16,8 @@
   }
 
   &.medium {
-    width: 32px;
-    height: 32px;
+    width: 24px;
```
PanchoutNathan, line 19: 🟢 (nit) This changes `.medium` from `32px` to `24px` **globally** — it affects all places using `ItemIcon` with medium size, not just the new dropdown. Has this been visually verified everywhere?

src/frontend/apps/drive/src/features/explorer/components/tree/ExplorerTreeActions.tsx
```diff
@@ -30,9 +47,46 @@ export const ExplorerTreeActions = ({
               icon: <img src={createFolderSvg.src} alt="" />,
```
PanchoutNathan, line 47: for all icons

PanchoutNathan: There must be an error in the Figma; we have the same thing everywhere.

PR #497, opened 2026-01-15
🏗️(ds_proxy) introduce how to use ds_proxy with Drive

lunika:
## Purpose

We want to add documentation showing how to use DS_Proxy with Drive. With proxy is fully optional and is here if you want an encryption layer between Drive and the object storage.

## Proposal

- [x] 🏗️(ds_proxy) introduce how to use ds_proxy with Drive


lunika:
@coderabbitai full review

PR #502, opened 2026-01-22
✨(frontend) implement silent login feature and configuration integration

NathanVss:
## Purpose

We want to be able to provide a silent login mecanism to the app. Let's do it by leveraging `prompt=None` during the OIDC flow. 

Highly inspired by the work done by @lebaudantoine on https://github.com/suitenumerique/meet


lunika (approved):
Ok for the backend review

PR #505, opened 2026-01-26
🔥(backend) remove usage of atomic transaction for item creation

lunika:
## Purpose

We used atomic transaction in the ItemViewset when a new item is created
(perform_create and children actions). This code can be removed, seen we
decided to use UUID for the item path. The race condition is possible when
multiple items are created does not exist anymore; we don't need to
Know the previously created item to guess the next path.

## Proposal

Description...

- [x] 🔥(backend) remove usage of atomic transaction for item creation
- [x] ⏪️(backend) revert AWS_S3_REGION_NAME configuration


src/backend/core/api/viewsets.py
```diff
@@ -572,7 +572,6 @@ def retrieve(self, request, *args, **kwargs):
 
         return drf.response.Response(serializer.data)
 
-    @transaction.atomic
```
NathanVss, line 575: Just in case something fails, why wouldn't we keep the atomicity of item creation?

env.d/development/common
```diff
@@ -25,7 +25,7 @@ DJANGO_EMAIL_PORT=1025
 STORAGES_STATICFILES_BACKEND=django.contrib.staticfiles.storage.StaticFilesStorage
 AWS_S3_ACCESS_KEY_ID=drive
 AWS_S3_SECRET_ACCESS_KEY=password
-AWS_S3_REGION_NAME=eu-east-1
+AWS_S3_REGION_NAME=us-east-1
```
NathanVss, line 28: If not needed anyway, why change to `us`?

NathanVss (approved):

PR #508, opened 2026-01-28
🔒️(backend) prevent mismatch mimetype between object storage and app

lunika:
## Purpose

If there is a mismatch mimetype between the object in the object storage and items in the application, this can lead to previewing files in the browser that are not allowed. We want to prevent this, the mimetype in the object storage is overridden in that case.

## Proposal

- [x] 🔒️(backend) prevent mismatch mimetype between object storage and app

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -23,6 +23,7 @@ from django.utils.text import slugify
 
 import posthog
 import rest_framework as drf
+from botocore.exceptions import ClientError
 from corsheaders.middleware import (
     ACCESS_CONTROL_ALLOW_METHODS,
     ACCESS_CONTROL_ALLOW_ORIGIN,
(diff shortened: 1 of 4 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -732,6 +733,25 @@ def upload_ended(self, request, *args, **kwargs):
 
         item.save(update_fields=["upload_state", "mimetype", "size"])
 
+        if head_response["ContentType"] != mimetype:
+            logger.info(
+                "upload_ended: content type mismatch betweeb object storage and item,"
```
NathanVss: typo "between"

src/backend/core/api/viewsets.py
```diff
@@ -732,6 +733,25 @@ def upload_ended(self, request, *args, **kwargs):
...
+                " updating from %s to %s",
+                head_response["ContentType"],
+                mimetype,
+            )
+            s3_client.copy_object(
```
NathanVss: Maybe add error handling if the copy operation raise an error for any reason?

PR #509, opened 2026-01-28
Mirror files on two buckets

lunika:
## Purpose

We want to allow the mirroring of the uploaded files between two distinct buckets that can be present in different cloud service providers.
To implement it, we created a dedicated S3 storage class inherited from the S3 storage backend. The `save` method is overridden to call a Celery task responsible for making the mirroring.
The mirroring consists of downloading the file and then re-uploading it in the other location. To maximize compatibility with storages not supporting multipart upload and/or download, we implemented it the more simple and compatible way. It is not the optimal one but with this we are mostly compatible with all storages.

## Proposal

- [x] ♻️(backend) remove usage of boto3 put_object method
- [x] ✨(backend) create a storage mirroring files on s3
- [x] ✨(backend) implement mirror_file task
- [x] ✨(backend) call mirror_file task in upload_ended viewset

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -26,8 +26,6 @@ This document lists all configurable environment variables for the Drive applica
 | `AWS_S3_MIRRORING_RESPONSE_CHECKSUM_VALIDATION` | AWS S3 response checksum calculation config for the mirroring bucket | `when_supported` |
 | `AWS_STORAGE_BUCKET_NAME` | AWS S3 bucket name for file storage | `drive-media-storage` |
 | `CACHES_DEFAULT_TIMEOUT` | Default cache timeout in seconds | `30` |
-| `CELERY_BROKER_URL` | Celery broker URL for task queue | `redis://redis:6379/0` |
-| `CELERY_TASK_ROUTES` | Celery task routing configuration. Use this to route specific tasks to dedicated queues, e.g. `{"core.tasks.storage.mirror_file": {"queue": "mirror"}}` | `{}` |
 | `CORS_ALLOW_ALL_ORIGINS` | Allow all origins for CORS | `False` |
 | `CORS_ALLOWED_ORIGINS` | List of allowed origins for CORS | `[]` |
 | `CORS_ALLOWED_ORIGIN_REGEXES` | List of allowed origin regexes for CORS | `[]` |
@@ -41,6 +39,8 @@ This document lists all configurable environment variables for the Drive applica
 | `DB_PASSWORD` | Database password | `pass` |
 | `DB_PORT` | Database port | `5432` |
 | `DB_USER` | Database user | `dinum` |
+| `DJANGO_CELERY_BROKER_URL` | Celery broker URL for task queue | `redis://redis:6379/0` |
+| `DJANGO_CELERY_TASK_ROUTES` | Celery task routing configuration. Use this to route specific tasks to dedicated queues, e.g. `{"core.tasks.storage.mirror_file": {"queue": "mirror"}}` | `{}` |
 | `EMAIL_BACKEND` | Email backend for sending emails | `django.core.mail.backends.smtp.EmailBackend` |
 | `EMAIL_BRAND_NAME` | Brand name for email templates | `None` |
 | `EMAIL_FROM` | Default sender email address | `from@example.com` |
--- a/src/helm/drive/templates/_helpers.tpl
+++ b/src/helm/drive/templates/_helpers.tpl
@@ -140,15 +140,6 @@ Requires top level scope
 {{- end }}
 
 {{/*
-Full name for celery
-
-Requires top level scope
-*/}}
-{{- define "drive.celery.fullname" -}}
-{{ include "drive.fullname" . }}-celery
-{{- end }}
-
-{{/*
 Full name for the frontend
 
 Requires top level scope
--- a/src/helm/drive/templates/backend_celery_worker.yaml
+++ b/src/helm/drive/templates/backend_celery_worker.yaml
@@ -1,48 +1,50 @@
 {{- $envVars := include "drive.common.env" (list . .Values.backend) -}}
 {{- $fullName := include "drive.backend.fullname" . -}}
-{{- $component := "backend-celery-worker" -}}
+{{- range $worker := .Values.backend.celeryWorkers }}
+{{- $component := printf "backend-celery-worker-%s" $worker.name }}
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ $fullName }}-celery
-  namespace: {{ .Release.Namespace | quote }}
+  name: {{ $fullName }}-celery-{{ $worker.name }}
+  namespace: {{ $.Release.Namespace | quote }}
   annotations:
-    {{- with .Values.backend.dpAnnotations }}
+    {{- with $.Values.backend.dpAnnotations }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
   labels:
-    {{- include "drive.common.labels" (list . $component) | nindent 4 }}
+    {{- include "drive.common.labels" (list $ $component) | nindent 4 }}
 spec:
-  replicas: {{ .Values.backend.celery.replicas }}
+  replicas: {{ $worker.replicas }}
   selector:
     matchLabels:
-      {{- include "drive.common.selectorLabels" (list . $component) | nindent 6 }}
+      {{- include "drive.common.selectorLabels" (list $ $component) | nindent 6 }}
   template:
     metadata:
       annotations:
-        {{- with .Values.backend.podAnnotations }}
+        {{- with $.Values.backend.podAnnotations }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
       labels:
-        {{- include "drive.common.selectorLabels" (list . $component) | nindent 8 }}
+        {{- include "drive.common.selectorLabels" (list $ $component) | nindent 8 }}
     spec:
       {{- if $.Values.image.credentials }}
       imagePullSecrets:
-        - name: {{ include "drive.secret.dockerconfigjson.name" (dict "fullname" (include "drive.fullname" .) "imageCredentials" $.Values.image.credentials) }}
+        - name: {{ include "drive.secret.dockerconfigjson.name" (dict "fullname" (include "drive.fullname" $) "imageCredentials" $.Values.image.credentials) }}
       {{- end}}
-      shareProcessNamespace: {{ .Values.backend.shareProcessNamespace }}
+      shareProcessNamespace: {{ $.Values.backend.shareProcessNamespace }}
       containers:
-        {{- with .Values.backend.sidecars }}
+        {{- with $.Values.backend.sidecars }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
-        - name: {{ .Chart.Name }}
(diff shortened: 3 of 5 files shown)


src/backend/drive/settings.py
NathanVss, line 1: Not sure the code matches the commit content?

NathanVss: the commit: 

```
🏗️(ds_proxy) introduce how to use ds_proxy with Drive
We want to add documentation showing how to use DS_Proxy with Drive.
With proxy is fully optionnal and is here if you want to an encryption
layer between Drive and the object storage.
```

src/backend/core/storage/s3_mirroring_storage.py
```diff
@@ -0,0 +1,18 @@
+"""S3 storage mirroring files on multiple S3 buckets."""
+
+from storages.backends.s3 import S3Storage
+
+from core.tasks.storage import mirror_file
+
+# pylint: disable=abstract-method
+
+
+class S3MirroringStorage(S3Storage):
```
NathanVss, line 10: Not sure why you created `S3MirroringStorage` and at the same time added `mirror_file.delay(item.file_key)` in upload_ended? 

I thought that the only requirement if we want to enable mirroring is to configure the storage backend in the settings to be `S3MirroringStorage` nope?

src/backend/core/tasks/storage.py
```diff
@@ -0,0 +1,65 @@
...
+            region_name=settings.AWS_S3_MIRRORING_REGION_NAME,
+            signature_version=settings.AWS_S3_MIRRORING_SIGNATURE_VERSION,
+            request_checksum_calculation=settings.AWS_S3_MIRRORING_REQUEST_CHECKSUM_CALCULATION,
+            response_checksum_validation=settings.AWS_S3_MIRRORING_RESPONSE_CHECKSUM_VALIDATION,
+        ),
+    )
+
+
+@app.task
+def mirror_file(name):
```
NathanVss: WDYT about adding a retry mecanism? ( it's built-in via the celery annotation )

src/backend/core/tasks/storage.py
```diff
@@ -0,0 +1,65 @@
...
+    mirror_bucket = settings.AWS_S3_MIRRORING_STORAGE_BUCKET_NAME
+
+    logger.info("Starting mirror of file %s to bucket %s", name, mirror_bucket)
+
+    with default_storage.open(name, mode="rb") as source_file:
+        mirror_s3_client.put_object(
+            Bucket=mirror_bucket,
+            Key=name,
+            Body=source_file,
+        )
```
NathanVss: Are you sure in case of mime type mismatch ( https://github.com/suitenumerique/drive/pull/508 ) the mirror ContentType would be aligned?

src/backend/core/tests/tasks/storage/test_mirror_file.py
```diff
@@ -0,0 +1,80 @@
...
+import pytest
+
+from core.tasks.storage import (
+    MissingMirroringConfigurationError,
+    get_mirror_s3_client,
+    mirror_file,
+)
+
+
+def test_get_mirror_s3_client_no_confif_should_raise_an_exception(settings):
```
NathanVss: typo

src/backend/core/tests/storage/test_s3_mirroring_storage.py
NathanVss, line 1: end of file name _test not needed?

src/backend/drive/settings.py
```diff
@@ -231,6 +235,43 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    # Miroring S3 settings
```
NathanVss: "Mirroring"

src/backend/core/storage/s3_mirroring_storage.py
NathanVss: Wouldn't the explicit `mirror_file.delay(item.file_key)` in the viewset + `S3MirroringStorage` configured result in a double buggy mirroring?

lunika: It is explain in the related commit:

```
when an upload is ended, we have to call the mirror_file task. The
upload is made in the client browser so we can't use the S3 storage
class dedicated to the mirroring.
```

So, unless I'm wrong, there is no double buggy mirroring.

src/backend/core/tasks/storage.py
lunika: Good point. I will add it.

src/backend/drive/settings.py
lunika: I don't understand why this commit is here. I probably messed up my rebase with the main branch.

lunika: I removed this commit.

src/backend/core/tasks/storage.py
```diff
@@ -43,24 +41,45 @@ def get_mirror_s3_client():
...
+            Key=item_key,
             Body=source_file,
+            ContentType=mirror_task.item.mimetype,
+            Metadata=head_object["Metadata"],
         )
 
-    logger.info("Successfully mirrored file %s to bucket %s", name, mirror_bucket)
+    logger.info("Successfully mirrored file %s to bucket %s", item_key, mirror_bucket)
+    mirror_task.status = MirrorItemTaskStatusChoices.COMPLETED
+    mirror_task.save(update_fields=["status", "updated_at"])
```
NathanVss, line 108: What about a strategy to clear the mirror task in DB that are COMPLETED? Do we want to keep them forever? Or a least provide a command to clear the completed ones?

src/backend/core/models.py
```diff
@@ -1063,6 +1072,30 @@ def move(self, target, ignore_parent_numchild_update=False):
             self._meta.model.objects.filter(pk=old_parent_id).update(**update)
 
 
+class MirrorItemTask(BaseModel):
```
NathanVss, line 1048: I was wondering about the that we could use a library like https://django-celery-results.readthedocs.io/en/latest/ instead of doing our own implementation of task models. Or maybe there is a specific reason?

lunika: Yes, this table will be used to mirror the entire existing items.

src/backend/core/tasks/storage.py
```diff
@@ -69,16 +69,31 @@ def mirror_file(mirror_task_id):
...
+        botocore.exceptions.BotoCoreError,
+        botocore.exceptions.ClientError,
+    ) as exc:
+        if self.request.retries >= self.max_retries:
+            mirror_task.status = MirrorItemTaskStatusChoices.FAILED
+            mirror_task.error_details = str(exc)
+            mirror_task.save(update_fields=["status", "error_details", "updated_at"])
+            return
+
+        mirror_task.status = MirrorItemTaskStatusChoices.PENDING
```
NathanVss, line 93: nit: might maybe be useful to log the retries count, wdyt?

Dockerfile
NathanVss, line 1: nit: would be great to have theses edits in another commit

docs/s3_mirroring.md
```diff
@@ -25,4 +25,14 @@ For the second object storage, used for mirroring, you have to set this settings
...
+
+## Migrate from one bucket to the other.
+
+The mirroring feature has been developed to ease the migration from one bucket located in one cloud provider to an other bucket located in an other cloud provider. 
+
+The scenario to makge the migration is this one :
+
+- Enable the mirroring by configuring it (see documentation before). The mirrored bucket should be the future one used after the migration completed
+- Run the management command `prepare_mirroring_items_history`, its purpose is to put in the `mirror_item_task` all the existing items in order to mirror them
+- Then, run the management command `process_pending_mirror_tasks` periodically. This management commen will put in a celery queue the tasks to process by batch of 100 items. You can configure a cron to trigger this management command every 5 or 10 minutes for example. You should monitor it to determint the interval between tasks.
```
NathanVss: `command`

src/backend/core/management/commands/process_pending_mirror_tasks.py
NathanVss, line 1: Maybe I'm wrong: it seems that it is possible to trigger tasks that are already being retried?

Even if it is the case it shouldn't be a big deal? As if I understand correctly this command is supposed to be sort of a last resort mecanism to start again stuck tasks?

src/backend/core/management/commands/process_pending_mirror_tasks.py
```diff
@@ -0,0 +1,49 @@
...
+
+        pending_tasks = MirrorItemTask.objects.filter(
+            status=MirrorItemTaskStatusChoices.PENDING
+        )[:slots_available]
+
+        for task in pending_tasks:
+            mirror_file.delay(task.id)
+
+        self.stdout.write(
+            f"Enqueued {pending_tasks.count()} pending mirror tasks for processing "
```
NathanVss: Edge case maybe but the `count()` could have changed between this log statement and before the for loop. Maybe use an counter in the loop?

src/backend/core/management/commands/prepare_mirroring_items_history.py
```diff
@@ -0,0 +1,64 @@
...
+        "each item that is a file."
+    )
+
+    def handle(self, *args, **options):
+        """Prepare the mirroring items history."""
+
+        self.stdout.write(self.style.SUCCESS("Preparing mirroring items history..."))
+
+        items_id = (
+            Item.objects.prefetch_related("mirror_tasks")
```
NathanVss, line 28: Is this prefetch really needed as you already check for `mirror_tasks__isnull` below?

src/backend/core/tests/commands/test_prepare_mirroring_items_history.py
```diff
@@ -0,0 +1,62 @@
...
+    ItemTypeChoices,
+    ItemUploadStateChoices,
+    MirrorItemTask,
+    MirrorItemTaskStatusChoices,
+)
+
+pytestmark = pytest.mark.django_db
+
+
+def test_prepare_mirroring_items_history_no_items():
```
NathanVss: The naming of this test is not really self explanatory imho.  Maybe "test_prepare_mirroring_items_history_no_tasks_duplication" or smth like this?

src/backend/core/admin.py
```diff
@@ -211,3 +212,38 @@ class InvitationAdmin(admin.ModelAdmin):
...
+    list_display = (
+        "id",
+        "item",
+        "status",
+        "updated_at",
+        "created_at",
+    )
+
+    list_filter = ("status",)
+    search_filter = (
```
NathanVss: `search_fields` maybe nope?

src/backend/core/admin.py
```diff
@@ -211,3 +212,38 @@ class InvitationAdmin(admin.ModelAdmin):
...
+    )
+    readonly_fields = (
+        "id",
+        "created_at",
+        "updated_at",
+    )
+    ordering = ("-created_at",)
+    actions = ("trigger_new_mirroring",)
+
+    def trigger_new_mirroring(self, request, queryset):
```
NathanVss, line 242: What if it is already processing?

src/backend/core/management/commands/prepare_mirroring_items_history.py
lunika: Needed to do the left join.

src/backend/core/admin.py
lunika: It will be reprocess. This is also the goal of this action, force reprocessing task blocked, for any reason, in processing.

Dockerfile
lunika: I moved it to the commit is should lives. And edited the body to explain why it is added here.

PR #518, opened 2026-02-03
✨(backend) add last_release_note_seen field to user

NathanVss:
Allow users to track which release note they last saw. The /me endpoint now supports PATCH so users can update their last_release_note_seen value. This will let the frontend know whether to display new release notes.


src/backend/core/api/serializers.py
```diff
@@ -32,6 +32,7 @@ class Meta:
             "full_name",
             "short_name",
             "language",
+            "last_release_note_seen",
```
lunika, line 43: Add a validation on the expected value. This will avoid to have a field where someone will inject an xss.

src/backend/core/models.py
```diff
@@ -227,6 +227,14 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
         help_text=_("Claims from the OIDC token."),
     )
 
+    last_release_note_seen = models.CharField(
+        _("last release note seen"),
+        max_length=85,
+        blank=True,
+        null=True,
+        default=None,
```
```suggestion
```

src/backend/core/tests/test_api_users.py
```diff
@@ -266,6 +266,7 @@ def test_api_users_retrieve_me_authenticated():
         "full_name": user.full_name,
         "short_name": user.short_name,
         "language": user.language,
+        "last_release_note_seen": None,
```
lunika, line 269: add an other test on the `/me` endpoint with this value set ?

src/backend/drive/settings.py
```diff
@@ -1027,6 +1027,9 @@ class Base(Configuration):
     FRONTEND_SILENT_LOGIN_ENABLED = values.BooleanValue(
         default=False, environ_name="FRONTEND_SILENT_LOGIN_ENABLED", environ_prefix=None
     )
+    FRONTEND_HIDE_RELEASE_NOTE = values.BooleanValue(
```
lunika: In order to be consistent, you should name this variable ending with `_ENABLED`. Maybe `FRONTEND_RELEASE_NOTE_ENABLED` ?

lunika: Once the modification is done, you also have to update the `docs/env.md` file to add this new variable

PR #520, opened 2026-02-03
🐛(backend) fix files with # in filename causing SignatureDoesNotMatch

### LINKED ISSUE #449: SignatureDoesNotMatch with filename containing "#"
## Bug Report

**Problematic behavior**
A filename with a special char gave a signature error:

Image

When renaming the file without the '#' is worked again.

NathanVss:
Calling unquote() before urlparse() decoded %23 to # which urlparse then treated as a fragment separator, truncating the file path. By parsing the URL first and only decoding the extracted path/query components, special characters like # no longer interfere with URL structure parsing.

Closes #449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to
 ### Fixed
 
 - 🐛(backend) correctly configure celery beat to run wopi configuration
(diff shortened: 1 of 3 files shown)


NathanVss:
@coderabbitai full review

PR #525, opened 2026-02-05
🐛(global) update ui when renaming file from wopi editor + customize wopi parameters

NathanVss:
Add support for renaming files from the WOPI editor (OnlyOffice) and a way to disable rename for clients that don't support it properly (Collabora). It also allows for future customisation that could be needed as an open source project.

The issue is that Collabora doesn't send a PostMessage when a file get renamed via the WOPI protocol, so the frontend has no way to know the file was renamed. OnlyOffice does send a File_Rename PostMessage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 - ✨(global) add FRONTEND_CSS_URL and FRONTEND_JS_URL settings
 - ✨(backend) add a download action returning the media url
 - ✨(frontend) add right click feature
+- ✨(backend) allow customization of wopi parameters
 
 ### Fixed
 
@@ -22,6 +23,7 @@ and this project adheres to
 - 🐛(backend) fix WOPI PutFile to check stored file size
 - 🐛(frontend) fix 401 page infinite redirect loop after login
 - 🐛(backend) fix OIDC redirect allowed hosts format in dev config
+- 🐛(global) update ui when renaming file from wopi editor
 
 ### Changed
 


env.d/development/common
```diff
@@ -77,6 +77,9 @@ WOPI_CLIENTS="collabora,onlyoffice"
 WOPI_COLLABORA_DISCOVERY_URL="http://collabora:9980/hosting/discovery"
 WOPI_ONLYOFFICE_DISCOVERY_URL=http://onlyoffice/hosting/discovery
 WOPI_SRC_BASE_URL=http://app-dev:8000
+# Disable rename for Collabora by default because if does not post messages when renaming a file.
```
lunika:
```suggestion
# Disable rename for Collabora by default because it does not post messages when renaming a file.
```

src/backend/drive/settings.py
```diff
@@ -1571,6 +1572,13 @@ def post_setup(cls):
         if cls.WOPI_CLIENTS:
             for wopi_client in cls.WOPI_CLIENTS:
                 wopi_client_upper = wopi_client.upper()
+
+                options_raw = os.environ.get(f"WOPI_{wopi_client_upper}_OPTIONS", "{}")
+                try:
+                    options = json.loads(options_raw)
+                except json.JSONDecodeError:
+                    options = {}
```
lunika: We don't using `values.DictValue` ? Check what is made for the `discovery_url`.

src/backend/wopi/viewsets.py
```diff
@@ -4,6 +4,7 @@
 import uuid
 from os.path import splitext
 
+from django.conf import settings as django_settings
```
lunika: Why not using `settings` ?

src/backend/wopi/viewsets.py
```diff
@@ -93,6 +95,19 @@ def retrieve(self, request, pk=None):
...
+        if wopi_client:
+            client_name = wopi_client.get("client", "")
+            client_config = django_settings.WOPI_CLIENTS_CONFIGURATION.get(
+                client_name, {}
+            )
+            client_options = client_config.get("options", {})
+            if "SupportsRename" in client_options:
+                properties["SupportsRename"] = client_options["SupportsRename"]
+                if not client_options["SupportsRename"]:
+                    properties["UserCanRename"] = False
```
lunika: This seems too specific; we will have to add more conditions here in the future.

why not constructing the `properties` dict line 70 by using the `options` present in the client config ?

For example : `"SupportsRename": client_options.get("SupportsRename", True)`

NathanVss: Agreed, + `SupportsRename` is enough.

src/backend/wopi/tests/viewset/test_check_file_info.py
```diff
@@ -3,13 +3,17 @@
 from io import BytesIO
 
 from django.contrib.auth.models import AnonymousUser
+from django.core.cache import cache as django_cache
```
lunika: any reason to rename it ?

NathanVss: Automatic import

Issue #443, opened 2025-12-01
new row for relation "drive_item" violates check constraint "drive_item_numchild_check"


NathanVss:
## Bug Report

**Problematic behavior**

When moving a folder the following error occurred:

```
new row for relation "drive_item" violates check constraint "drive_item_numchild_check"
DETAIL:  Failing row contains (<path>, <uuid>, 2025-12-01 09:03:19.939631+00, 2025-12-01 09:03:19.939662+00, <name>, restricted, reader, null, null, null, folder, null, -1, -1, <uuid>, null, f, null, null, null, {}).
```

<img width="762" height="372" alt="Image" src="https://github.com/user-attachments/assets/147b70be-30d4-47a5-a0f9-7cbee316c632" />

And then after few moves the user told that this error was gone. No idea on how to reproduce it.

**Expected behavior/code**
It should not crash.

selimslab:
Opened PR #527 to fix

PR #527, opened 2026-02-07
Prevent numchild going under zero

selimslab:
Fix bug #443 

numchild must be zero or more, can't be negative


lunika:
Hi,

thanks for your contribution. Your solution does not fix the root cause. I think we are facing a race condition, both on creation (when the increment is made) and the deletion/removal/restore

lunika:
Managed in #566

PR #532, opened 2026-02-10
🐛(backend) remove folders in the recents view

lunika:
## Purpose

The recents view should only return files. We have to explicitely filter on type=file and exlude filtering on the type column.

## Proposal

- [x] 🐛(backend) remove folders in the recents view


lunika:
This should not be merged The endpoint allow filtering on the type, so the frontend should use this type filtering to have only files or folders depending the case.

PR #545, opened 2026-02-18
📈(global) add posthog tracking related to advanced share

NathanVss:
Here's the list of new PostHog events:

  Backend:
  - item_moved
  - item_move_missing_permission
  - item_favorited
  - item_unfavorited
  - item_access_created
  - item_access_updated
  - item_access_deleted
  - item_invitation_created
  - item_invitation_updated
  - item_invitation_deleted

  Frontend:
  - click_copy_link
  - click_redirect_to_parent_item

Also add tests to cover the events.

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-drive/share.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-drive/share.spec.ts
@@ -1,4 +1,4 @@
-import { test as base, BrowserContext, Page } from "@playwright/test";
+import { test as base, BrowserContext, expect, Page } from "@playwright/test";
 import { clearDb, login } from "./utils-common";
 import {
   clickToMyFiles,
@@ -7,6 +7,8 @@ import {
 } from "./utils-navigate";
 import { createFolderInCurrentFolder } from "./utils-item";
 import {
+  clickCopyLinkButton,
+  clickOnMemberItemRole,
   closeShareModal,
   expectAllowedLinkReach,
   expectAllowedRoles,
@@ -19,7 +21,11 @@ import {
   expectRowItem,
   expectRowItemIsNotVisible,
 } from "./utils-embedded-grid";
-import { clickOnBreadcrumbButtonAction } from "./utils-explorer";
+import {
+  clickOnBreadcrumbButtonAction,
+  expectExplorerBreadcrumbs,
+} from "./utils-explorer";
+import { setupPosthogEventCapture } from "./utils/posthog-utils";
 
 type TwoUsers = {
   userA: { context: BrowserContext; page: Page };
@@ -82,7 +88,11 @@ MultiUserTest(
     await shareCurrentItemWithWebkitUser(userA.page, "Editor");
     await closeShareModal(userA.page);
     await createFolderInCurrentFolder(userA.page, "Sub folder");
-    await navigateToFolder(userA.page, "Sub folder", ["My files", "Folder", "Sub folder"]);
+    await navigateToFolder(userA.page, "Sub folder", [
+      "My files",
+      "Folder",
+      "Sub folder",
+    ]);
     await clickOnBreadcrumbButtonAction(userA.page, "Share");
 
     await expectAllowedRoles(
@@ -111,7 +121,11 @@ MultiUserTest(
     await expectLinkReachSelected(userA.page, "Connected");
     await closeShareModal(userA.page);
     await createFolderInCurrentFolder(userA.page, "Sub folder");
-    await navigateToFolder(userA.page, "Sub folder", ["My files", "Folder", "Sub folder"]);
+    await navigateToFolder(userA.page, "Sub folder", [
+      "My files",
+      "Folder",
+      "Sub folder",
+    ]);
     await openShareModal(userA.page);
     await expectLinkReachSelected(userA.page, "Connected");
     await expectAllowedLinkReach(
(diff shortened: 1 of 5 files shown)


src/backend/core/tests/items/test_api_item_accesses_create.py
```diff
@@ -470,3 +472,33 @@ def test_api_item_accesses_create_authenticated_owner_syncronize_descendants_acc
 
     # access on item should be kept
     assert models.ItemAccess.objects.filter(item=item, user=other_user).count() == 1
+
+
+# Posthog events
+
+
+@override_settings(POSTHOG_KEY="fake-key")
+def test_api_item_accesses_create_posthog_event():
```
lunika:
```suggestion
@override_settings(POSTHOG_KEY="fake-key")
def test_api_item_accesses_create_posthog_event(settings):
```

Except for the test for the config api we always use the `settings` fixture to override settings.

src/backend/core/tests/items/test_api_item_invitations.py
```diff
@@ -860,3 +860,103 @@ def test_api_item_invitations_create_lower_email():
 
     assert response.status_code == 201
     assert response.json()["email"] == "guest@example.com"
+
+
+# Posthog events
+
+
+@override_settings(POSTHOG_KEY="fake-key")
```
lunika: Same in this file, use the `settings` fixture

src/backend/core/tests/items/test_api_items_favorite.py
```diff
@@ -1,5 +1,9 @@
 """Test favorite item API endpoint for users in drive's core app."""
 
+from unittest import mock
+
+from django.test import override_settings
```
lunika: same here

src/backend/core/tests/items/test_api_items_move.py
```diff
@@ -3,8 +3,11 @@
 """
 
 import random
+from unittest import mock
 from uuid import uuid4
 
+from django.test import override_settings
```
lunika: same here

src/backend/core/utils/analytics.py
```diff
@@ -0,0 +1,23 @@
...
+    if settings.POSTHOG_KEY:
+        properties = properties.copy()
+        item = kwargs.get("item")
+        if item:
+            properties["item_id"] = item.id
+            properties["item_title"] = item.title
+            properties["item_size"] = item.size
+            properties["item_mimetype"] = item.mimetype
+            properties["item_type"] = item.type
+        posthog.capture(
```
lunika, line 21: The capture sometimes fails, in sentry we see lot http error on this call. Are we sure this is not blocking the Drive request ?

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.tsx
```diff
@@ -451,6 +452,15 @@ export const ItemShareModal = ({
                   `${window.location.origin}/explorer/items/${itemId}`,
                 );
               }
+              posthog.capture("click_copy_link", {
```
PanchoutNathan, line 455: ### Property naming: back/front alignment ✅ but internal inconsistency ⚠️

The naming `item_id`, `item_size`, `item_mimetype` etc. is **consistent with the new backend utility** `posthog_capture` which enriches events with the same `item_` prefix:

```python
# src/backend/core/utils/analytics.py
properties["item_id"] = item.id
properties["item_title"] = item.title
properties["item_size"] = item.size
properties["item_mimetype"] = item.mimetype
properties["item_type"] = item.type
```

👍 for the alignment with this new convention.

**However**, the `click_redirect_to_parent_item` event (further down in this same file) uses `id` instead of `item_id`, which creates an inconsistency within this same PR. See my comment below.

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.tsx
```diff
@@ -451,6 +452,15 @@ export const ItemShareModal = ({
...
+                item_id: itemId,
+                item_title: item?.title,
```
PanchoutNathan, line 457: ### ❓ `item_title` sends the document title in plain text to PostHog

Existing events (`file_download`, `file_preview_opened`) do not send the title. The backend does via the `posthog_capture` utility, but this is potentially sensitive data (user document names).

Is this an intentional product/GDPR decision? If so, no issue — but it might be worth documenting (or at least validating) since this is user data being sent to a third-party service.

src/frontend/apps/drive/src/features/explorer/components/modals/share/ItemShareModal.tsx
```diff
@@ -507,6 +517,9 @@ const RedirectionToParentItem = ({
   const router = useRouter();
 
   const handleRedirectToParentItem = () => {
+    posthog.capture("click_redirect_to_parent_item", {
```
PanchoutNathan, line 520: ### ⚠️ Naming inconsistency with `click_copy_link`

Here we use `id`:
```ts
posthog.capture("click_redirect_to_parent_item", {
  id: itemId,
});
```

But `click_copy_link` (above) uses `item_id`:
```ts
posthog.capture("click_copy_link", {
  item_id: itemId,
});
```

In PostHog, this will create two different properties for the same concept. I suggest aligning on `item_id` here as well, to be consistent with the `item_` prefix convention introduced by the backend utility in this same PR.

```suggestion
    posthog.capture("click_redirect_to_parent_item", {
      item_id: itemId,
    });
```

src/frontend/apps/e2e/__tests__/app-drive/utils/posthog-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  await page.route(`${FAKE_POSTHOG_HOST}/**`, async (route) => {
+    try {
+      const postData = JSON.parse(route.request().postData() ?? "{}");
+      if (postData.event) {
+        events.push(postData.event);
+      }
+    } catch {
+      // ignore non-JSON requests (e.g. /decide)
+    }
+    await route.fulfill({
```
PanchoutNathan, line 36: ### Nit: empty `catch`

The comment explains the intent well, but an empty `catch` can hide real issues (e.g. a parsing bug that would never be detected).

An alternative would be a more explicit check rather than try/catch:

```suggestion
    } catch {
      // non-JSON requests (e.g. /decide) — safe to ignore
    }
```

Minor — the existing comment is sufficient. Just a readability suggestion.

src/frontend/apps/e2e/__tests__/app-drive/share.spec.ts
```diff
@@ -7,10 +7,13 @@ import {
 } from "./utils-navigate";
 import { createFolderInCurrentFolder } from "./utils-item";
 import {
+  clickCopyLinkButton,
+  clickOnMemberItemRole,
   closeShareModal,
   expectAllowedLinkReach,
   expectAllowedRoles,
   expectLinkReachSelected,
+  getShareModal,
```
PanchoutNathan: unused

src/backend/core/utils/analytics.py
NathanVss: I hope not, new need to investigate the sentry trace

PR #547, opened 2026-02-18
✨(backend) allow item creation on the external API

NathanVss:
## Summary

- Enable the `create` action on the external API items endpoint so resource servers can upload files at root level
- Add an e2e test covering the full flow: create item at root + upload-ended


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changes
```
lunika:
```suggestion
### Changed
```

PR #556, opened 2026-02-20
🐛(frontend) fix 401 page infinite redirect loop after login

NathanVss:
- Extract the redirect-after-login logic into a dedicated `useRedirectAfterLogin` hook for reusability
- Fix the 401 page login button redirecting back to `/401` after authentication, causing an infinite loop — it now redirects to `/` so the home page handles post-login navigation
- Fix `OIDC_REDIRECT_ALLOWED_HOSTS` dev config which included the `http://` scheme, causing redirect validation to fail
- Add an e2e test covering the full flow: session expiry → 401 page → re-login → redirect back to the original page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to
 
 - ✨(frontend) sync backend user language to browser on load
 - 🐛(backend) fix WOPI PutFile to check stored file size
+- 🐛(frontend) fix 401 page infinite redirect loop after login
+- 🐛(backend) fix OIDC redirect allowed hosts format in dev config
 
 ### Changed
 


src/frontend/apps/drive/src/hooks/useRedirectAfterLogin.tsx
```diff
@@ -0,0 +1,21 @@
+import { SESSION_STORAGE_REDIRECT_AFTER_LOGIN_URL } from "@/features/api/fetchApi";
+import { useAuth } from "@/features/auth/Auth";
+import { useEffect } from "react";
+
+export const useRedirectAfterLogin = () => {
+  const { user } = useAuth();
+  // Redirect to the attempted url if it exists, otherwise redirect to the last visited item.
```
PanchoutNathan, line 7: to the last visited item ? How ? I i see only the my-files route

NathanVss: You can see the added test to understand how to replicate this and reproduce on your side

PR #558, opened 2026-02-21
Prevent uploaded files exceeding the max allowed file size

lunika:
## Purpose

We can prevent starting a new upload if the file size is higher than the max size allowed. To also ensure it is respected, the same check is made in the upload_ended action.

## Proposal

- [x] ✨(backend) reject uploaded file higher than DATA_UPLOAD_MAX_MEMORY_SIZE
- [x]  ✨(frontend) block upload if file size higher than DATA_UPLOAD_MAX_MEMORY_SIZE
- [x]  ✨(backend) expose DATA_UPLOAD_MAX_MEMORY_SIZE in the config endpoint

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -32,6 +32,7 @@ This document lists all configurable environment variables for the Drive applica
 | `CRISP_WEBSITE_ID` | Crisp chat widget website ID | `None` |
 | `CSRF_TRUSTED_ORIGINS` | List of trusted origins for CSRF | `[]` |
 | `DATA_DIR` | Directory for storing application data | `/data` |
+| `DATA_UPLOAD_MAX_MEMORY_SIZE` | max upload file size, in bytes | `2147483648` (2 GB) |
 | `DATABASE_URL` | Database connection URL (overrides individual DB settings) | `None` |
 | `DB_ENGINE` | Database engine | `django.db.backends.postgresql` |
 | `DB_HOST` | Database host | `localhost` |
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -714,6 +714,18 @@ class ItemViewSet(
         )
         file_size = head_response["ContentLength"]
 
+        if file_size > settings.DATA_UPLOAD_MAX_MEMORY_SIZE:
+            self._complete_item_deletion(item)
+            logger.info(
+                "upload_ended: file size (%s) for file %s higher than the allowed max size",
+                file_size,
+                item.file_key,
+            )
+            raise drf.exceptions.ValidationError(
+                detail="The file size is higher than the allowed max size.",
+                code="file_size_exceeded",
+            )
+
         if file_size > 2048:
             range_response = s3_client.get_object(
                 Bucket=default_storage.bucket_name,
--- a/src/backend/core/tests/items/test_api_item_upload_ended.py
+++ b/src/backend/core/tests/items/test_api_item_upload_ended.py
@@ -363,3 +363,32 @@ def test_api_upload_ended_mismatch_mimetype_with_object_storage(caplog):
     )
     assert head_object["ContentType"] == "application/pdf"
     assert head_object["Metadata"] == {"foo": "bar"}
+
+
+def test_api_upload_ended_file_size_exceeded(settings, caplog):
+    """
+    Test when the file size exceed the allowed max upload file size
+    should return a 400 and delete the file.
+    """
+
+    settings.DATA_UPLOAD_MAX_MEMORY_SIZE = 0
+
+    user = factories.UserFactory()
+    client = APIClient()
+    client.force_login(user)
+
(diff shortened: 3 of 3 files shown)


src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -15,6 +15,20 @@ import { useQueryClient } from "@tanstack/react-query";
...
+    return `${(bytes / (1024 * 1024 * 1024)).toFixed(1)} GB`;
+  }
+  if (bytes >= 1024 * 1024) {
+    return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
+  }
+  if (bytes >= 1024) {
+    return `${(bytes / 1024).toFixed(1)} KB`;
+  }
+  return `${bytes} B`;
+};
```
PanchoutNathan: `formatFileSize` duplicates `formatSize` which already exists in `src/features/explorer/utils/utils.ts:126` (used in `ItemInfo.tsx`), and the project also uses `pretty-bytes` (in `FileUploadToast.tsx`). Consider reusing one of them for consistency.

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -341,14 +356,35 @@ export const useUploadZone = ({ item }: { item: Item }) => {
...
+      const tooLargeFiles = maxSize
+        ? upload.files.filter((file) => file.size > maxSize)
+        : [];
+      for (const file of tooLargeFiles) {
+        addToast(
+          <ToasterItem type="error">
+            <span>
+              {t("explorer.actions.upload.file_too_large", {
+                name: file.name,
+                maxSize: formatFileSize(maxSize!),
```
PanchoutNathan: Nit: the `maxSize!` non-null assertion could be avoided by restructuring slightly — e.g. putting `formatFileSize` inside an `if (maxSize)` block or using an early return. Not blocking, but `!` silences the type checker.

src/frontend/apps/drive/src/features/explorer/hooks/useUpload.tsx
```diff
@@ -341,14 +356,35 @@ export const useUploadZone = ({ item }: { item: Item }) => {
       const upload = filesToUpload(acceptedFiles);
       await handleHierarchy(upload);
 
+      // Filter out files that exceed the maximum upload size.
+      const maxSize = config.DATA_UPLOAD_MAX_MEMORY_SIZE;
+      const validFiles = maxSize
+        ? upload.files.filter((file) => file.size <= maxSize)
+        : upload.files;
```
PanchoutNathan: The behavior when `DATA_UPLOAD_MAX_MEMORY_SIZE` is absent/undefined is to silently disable the size check — this is correct for integrators who don't configure a limit. However, note that since `maxSize` is a `number`, a value of `0` would be falsy and would also disable the check. If `0` were to mean "no files allowed", the current code wouldn't handle it. A more explicit check (`maxSize !== undefined && maxSize !== null`) would be safer, or at minimum a comment to document this choice.

src/backend/core/api/viewsets.py
```diff
@@ -667,6 +667,18 @@ def upload_ended(self, request, *args, **kwargs):
...
+        if file_size > settings.DATA_UPLOAD_MAX_MEMORY_SIZE:
+            self._complete_item_deletion(item)
+            logger.info(
+                "upload_ended: file size (%s) for file %s higher than the allowed max size",
+                file_size,
+                item.file_key,
+            )
+            raise drf.exceptions.ValidationError(
+                detail="The file size is higher than the allwed max size.",
+                code="file_size_exceedeed",
```
PanchoutNathan: Two typos here:
- `"allwed"` → `"allowed"`
- `code="file_size_exceedeed"` → `code="file_size_exceeded"`

docs/env.md
```diff
@@ -32,6 +32,7 @@ This document lists all configurable environment variables for the Drive applica
 | `CRISP_WEBSITE_ID` | Crisp chat widget website ID | `None` |
 | `CSRF_TRUSTED_ORIGINS` | List of trusted origins for CSRF | `[]` |
 | `DATA_DIR` | Directory for storing application data | `/data` |
+| `DATA_UPLOAD_MAX_MEMORY_SIZE` | max upload file size, in bytes | `2097152` (2 GB) |
```
PanchoutNathan: `2097152 (2 GB)` → 2,097,152 bytes = **2 MB**, not 2 GB.

PanchoutNathan:
**UX suggestion for @rl-83:** the current flow requires a drop (drag & drop) to trigger the size check. For a single file upload (via the "Import" button), wouldn't it be better to use react-dropzone's `maxSize` attribute or check the file size as soon as the file is selected, **before** kicking off the whole flow (folder creation, progress toast, etc.)? That would give the user more immediate feedback. For multi-file uploads, the current flow with per-file error toasts is well suited.

PR #561, opened 2026-02-23
Add a url_permalink to prevent url changing when a file is renamed

### LINKED ISSUE #510: Prevent item url from changing when renaming it
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
As the `url` attribute of the item contains its filename, when we rename it the previous url does not longer work.

The issue is that when we share the url, we are not sure it is eternal ..

**Describe the solution you'd like**
A url that always stays the same. Ideally it shouldn't contain the filename, but only the uuid of the item.

lunika:
## Purpose

When a file is renamed, the value of the file url is also changing. We can't change this property because the nginx auth_request is just a proxy in front of the s3 bucket.
The workaround implemented here is to add a `download` action always redirecting to the good media url.
To expose this new action, a new property `url_permalink` is added.

## Proposal

- [x] ✨(backend) add a download action returning the media url
- [x] ✨(backend) expose a url_permalink property in the Item serializer.

Fix #510


NathanVss (approved):
Neat :)

PR #566, opened 2026-02-25
🐛(backend) stop storing numchild in database use annotation instead

### LINKED ISSUE #443: new row for relation "drive_item" violates check constraint "drive_item_numchild_check"

lunika:
## Purpose

In the item model the properties numchild and numchild_folder must be maintain by doing a +/- 1 operation in several actions. Maintaining is error prone, hard to catch and can have lot of root cause we can't identified. We choose to use annotation instead and to compute the value when needed.

## Proposal

- [x] 🐛(backend) stop storing numchild in database use annotation instead

fix #443

### DIFF
--- a/src/backend/uv.lock
+++ b/src/backend/uv.lock
@@ -959,14 +959,14 @@ wheels = [
 
 [[package]]
 name = "joserfc"
-version = "1.4.3"
+version = "1.6.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "cryptography" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/dc/32/8315032da02a124a326c2c98724387a4c9dcfa248efeb1307db8f2662331/joserfc-1.4.3.tar.gz", hash = "sha256:dcbd447ff70c64d75f4cbdc5fb1d48f6914bff67b97fa0f779dd5c581d1bb81d", size = 200514, upload-time = "2025-11-19T05:52:27.942Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/ce/90/b8cc8635c4ce2e5e8104bf26ef147f6e599478f6329107283cdc53aae97f/joserfc-1.6.3.tar.gz", hash = "sha256:c00c2830db969b836cba197e830e738dd9dda0955f1794e55d3c636f17f5c9a6", size = 229090, upload-time = "2026-02-25T15:33:38.167Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/d9/4e/b998a5f8f042a17a6a87e7ed7fb85082d62f7d5c1d307ac6c544a2c7293c/joserfc-1.4.3-py3-none-any.whl", hash = "sha256:3cb2f1c1e3285c44a50ca9cb3287f37c2c665442028a4983aed87a75285164dc", size = 66957, upload-time = "2025-11-19T05:52:26.15Z" },
+    { url = "https://files.pythonhosted.org/packages/12/4f/124b3301067b752f44f292f0b9a74e837dd75ff863ee39500a082fc4c733/joserfc-1.6.3-py3-none-any.whl", hash = "sha256:6beab3635358cbc565cb94fb4c53d0557e6d10a15b933e2134939351590bda9a", size = 70465, upload-time = "2026-02-25T15:33:36.997Z" },
 ]
 
 [[package]]


src/backend/core/migrations/0020_remove_item_numchild_path_nlevel_idx.py
```diff
@@ -0,0 +1,31 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name="item",
+            name="numchild",
+        ),
+        migrations.RemoveField(
+            model_name="item",
+            name="numchild_folder",
+        ),
```
qbey: Migrations are run before deployment: doing this will cause a downtime while not all backend pods deployed. You should consider removing columns in the next release (we need to improve migration management)

src/backend/core/models.py
```diff
@@ -631,6 +665,33 @@ def nb_accesses(self):
...
+        """Calculate the number of non-deleted children."""
+        try:
+            return self._numchild
+        except AttributeError:
+            return Item.objects.filter(
+                path__descendants=self.path,
+                deleted_at__isnull=True,
+                ancestors_deleted_at__isnull=True,
+                path__depth=self.depth + 1,
+            ).count()
```
qbey: I don't like that : I think we should explicitly raise here, because this could introduce N+1 queries if wrongly used. We have to enforce the dev to do the annotate before.

src/backend/core/models.py
```diff
@@ -631,6 +665,33 @@ def nb_accesses(self):
...
+        try:
+            return self._numchild_folder
+        except AttributeError:
+            return Item.objects.filter(
+                path__descendants=self.path,
+                deleted_at__isnull=True,
+                ancestors_deleted_at__isnull=True,
+                path__depth=self.depth + 1,
+                type=ItemTypeChoices.FOLDER,
+            ).count()
```
qbey: Same here

src/backend/core/models.py
```diff
@@ -377,6 +379,50 @@ def annotate_user_roles(self, user):
...
+            .annotate(count=models.Count("pk"))
+            .values("count"),
+            output_field=models.IntegerField(),
+        )
+
+        return self.annotate(
+            _numchild=Coalesce(numchild_sq, 0, output_field=models.IntegerField()),
+            _numchild_folder=Coalesce(
+                numchild_folder_sq, 0, output_field=models.IntegerField()
+            ),
```
qbey: Not sure why `Coalesce` is needed here, but I guess you know ^^

jmaupetit: Maybe because the sub-query can return `None` and we only want integers here?

src/backend/core/models.py
```diff
@@ -377,6 +379,50 @@ def annotate_user_roles(self, user):
...
+        """
+        direct_children_qs = (
+            Item.objects.filter(
+                path__descendants=models.OuterRef("path"),
+                deleted_at__isnull=True,
+                ancestors_deleted_at__isnull=True,
+            )
+            .annotate(_depth_diff=NLevel("path") - NLevel(models.OuterRef("path")))
+            .filter(_depth_diff=1)
+            .order_by()
```
jmaupetit, line 424: Not sure ordering matters here?

```suggestion

```

src/backend/core/models.py
```diff
@@ -506,6 +539,7 @@ class Meta:
         ]
         indexes = [
             GistIndex(fields=["path"]),
+            models.Index(NLevel(models.F("path")), name="drive_item_path_nlevel_idx"),
```
jmaupetit, line 574: 🥇

src/backend/core/models.py
lunika: Without I have this error `django.db.utils.ProgrammingError: column "u0.created_at" must appear in the GROUP BY clause or be used in an aggregate function`

src/backend/core/models.py
lunika: No it can be safely removed. The subqueries will always return an integer.

jmaupetit:
There is something surprising on the linters configuration (pylint detects unused imports and not ruff). Also not sure why pylint and ruff are needed. I'll look into that.

PR #567, opened 2026-02-25
🔥(backend) remove unused ServerToServerAuthentication backend

lunika:
## Purpose

The backend ServerToServerAuthentication is not used in the application and will not be used. We decided to remove it.

## Proposal

- [x] 🔥(backend) remove unused ServerToServerAuthentication backend


NathanVss (approved):
As we have now `rest_framework_api_key` LGTM. If we need this feature back we could use this library.

PR #573, opened 2026-02-26
🐛(backend) sanitize filename to be compatible with filesystems

### LINKED ISSUE #555: Lacking filename validation allows invalid S3 object keys (non‑printable / unsafe characters)
Hi,

## Bug Report

**Problematic behavior**
It appears that Drive currently allows file names containing characters that are not valid for some S3-compatible object storage providers.

In particular, filenames containing non-printable characters (e.g., newline) or unsafe sequences are not validated before attempting upload. This results in a failure at the S3 layer instead of being caught earlier with proper validation.

For example our instance contains a file with the name `">␊.txt` (coming from a YesWeHack test) that is uploadable on some S3 providers (e.g. Scaleway) but not others (e.g. Outscale)

**Expected behavior**
The AWS documentation [clearly states](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html) that some characters should be avoided (see "Characters to avoid" chapter). IMHO drive should be the one restraining file names and catch these early to align with AWS S3 best practices.

**Additional context**
This was caught while attempting to rclone sync from a Scaleway S3 bucket to an Outscale S3 bucket :
```
item/fd23b497-7ff6-4be4-ac85-d2b99c45dbfc/"><img src=x onerror=alert()>␊.txt: Failed to copy: operation error S3: PutObject, https response error StatusCode: 400, RequestID: 46db91e3c64f8d436e5f, HostID: 46db91e3c64f8d436e5f, api error InvalidInput: object keys cannot contain non-printable characters
```

lunika:
## Purpose

Filename were used as it, provided by the user. We want to force sanitization on them to ensure they will be always compatible with filesystems.

## Proposal

- [x] 🐛(backend) sanitize filename to be compatible with filesystems

Fix #555

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -511,14 +511,18 @@ class ItemSerializer(ListItemSerializer):
 
     def update(self, instance, validated_data):
         """Validate that the title is unique in the current path."""
-        if (
-            validated_data.get("title")
-            and instance.title != validated_data.get("title")
-            and instance.depth > 1
+        if validated_data.get("title") and instance.title != validated_data.get(
+            "title"
         ):
-            validated_data["title"] = instance.manage_unique_title(
-                validated_data.get("title")
-            )
+            if instance.depth > 1:
+                validated_data["title"] = instance.manage_unique_title(
+                    validated_data.get("title")
+                )
+
+            if instance.type == models.ItemTypeChoices.FILE:
+                # Just check for validation, the real filename
+                # will be use later in the rename_file task
+                utils.sanitize_filename(validated_data["title"])
 
         return super().update(instance, validated_data)
 
@@ -643,14 +647,13 @@ class CreateItemSerializer(ItemSerializer):
                         {"filename": _("This field is required for files.")},
                         code="item_create_file_filename_required",
                     )
-
                 if settings.RESTRICT_UPLOAD_FILE_TYPE:
-                    _root, ext = splitext(attrs["filename"])
-                    if ext.lower() not in settings.FILE_EXTENSIONS_ALLOWED:
+                    _root, extension = splitext(attrs["filename"])
+                    if extension.lower() not in settings.FILE_EXTENSIONS_ALLOWED:
                         logger.info(
                             "create_item: file extension not allowed %s "
                             "for filename %s",
-                            ext,
+                            extension,
                             attrs["filename"],
                         )
                         raise serializers.ValidationError(
@@ -660,6 +663,8 @@ class CreateItemSerializer(ItemSerializer):
 
                 # When it's a file we force the title with the filename
                 attrs["title"] = attrs["filename"]
+                # Use the sanitize_filename utils
+                attrs["filename"] = utils.sanitize_filename(attrs["filename"])
 
         if (
             attrs["type"] == models.ItemTypeChoices.FOLDER
--- a/src/backend/core/api/utils.py
+++ b/src/backend/core/api/utils.py
@@ -2,9 +2,13 @@
 
 import logging
 import mimetypes
+import re
+import unicodedata
 from datetime import datetime
+from os.path import splitext
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.core.files.storage import default_storage
 
 import boto3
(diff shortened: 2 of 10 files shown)


src/backend/core/tests/items/test_api_items_children_create.py
```diff
@@ -599,3 +599,70 @@ def test_api_items_children_create_entitlements_backend_returns_falsy(
...
+        else:
+            item = factories.ItemFactory(parent=item, type=ItemTypeChoices.FOLDER)
+
+    now = timezone.now()
+    with freeze_time(now):
+        response = client.post(
+            f"/api/v1.0/items/{item.id!s}/children/",
+            {
+                "type": ItemTypeChoices.FILE,
+                "filename": "><img src=x onerror=alert()>␊.txt",
```
lebaudantoine, line 629: let's parametrize this with more examples of invalid filenames wdyt? 

LLM generated (approximative) suggestion : 

```
@pytest.mark.parametrize("filename,expected_sanitized", [
    # XSS / HTML injection (your current case)
    ("><img src=x onerror=alert()>␊.txt", "img_srcx_onerroralert.txt"),
    # Path traversal
    ("../../../etc/passwd.txt", "etcpasswd.txt"),
    # Null bytes
    ("file\x00name.txt", "filename.txt"),
    # Only special chars (edge: what's left after sanitization?)
    ("!@#$%^&*().txt", ".txt"),  # or whatever your sanitizer produces
    # Unicode / homoglyph
    ("filеname.txt", "filename.txt"),  # Cyrillic 'е'
    # Extremely long filename
    ("a" * 300 + ".txt", "a" * 255 + ".txt"),  # assuming 255 char limit
    # Spaces and dots
    ("my file name.txt", "my_file_name.txt"),  # or however spaces are handled
])
```

src/backend/core/tasks/item.py
```diff
@@ -69,7 +70,7 @@ def rename_file(item_id, new_title):
 
     _, extension = splitext(item.filename)
 
-    new_filename = f"{new_title}{extension}"
+    new_filename = f"{get_valid_filename(new_title)}{extension}"
```
lebaudantoine: (not related to this PR)

What about having a possible double extension?
If new_title already contains the file extension (e.g. "report.pdf"), and extension from item.filename is also ".pdf", you'll end up with "report.pdf.pdf".

src/backend/core/tests/items/test_api_items_children_create.py
NathanVss: +1

NathanVss: More broadly I think it would be great to ensure we at least test against all the "Characters to avoid" from https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

src/backend/core/api/serializers.py
```diff
@@ -669,6 +670,7 @@ def validate(self, attrs):
 
                 # When it's a file we force the title with the filename
                 attrs["title"] = attrs["filename"]
+                attrs["filename"] = get_valid_filename(attrs["filename"])
```
NathanVss: It seems that `get_valid_filename('>>')` called with special only chars returns an empty string, we should make sure it is well handled

src/backend/core/tasks/item.py
lunika: if you try to change the extension, yes you will have a double extension. We don't allow changing it.

src/backend/core/api/utils.py
```diff
@@ -220,3 +224,27 @@ def detect_mimetype(file_buffer: bytes, filename: str | None = None) -> str:
...
+
+
+def sanitize_filename(filename):
+    """
+    Sanitize a filename to be compliant to use on filesystem.
+
+    """
+    name, extension = splitext(filename)
+    #  Convert to ASCII. Taken from the django slugify helper.
+    name = unicodedata.normalize("NFKD", name).encode("ascii", "ignore").decode("ascii")
```
NathanVss, line 237: Why is it needed? Might be great to add a comment to explain that.

NathanVss (approved):
Nlç€ =)

src/backend/core/api/utils.py
lunika: Maybe it is not a strict requirement, it just ensure that no strange thinks are added in the filename. Like emoji

PR #576, opened 2026-02-26
🐛(backend) allow inviting external person on item with no direct access

### LINKED ISSUE #575: Impossible to invite external person on an item without explicit access
## Bug Report

**Problematic behavior**

On an item without explicit access, a user is not able to invite an external person. The invitations endpoint always return a 403

**Expected behavior/code**

A user should be able to invite external person on every items in a tree. 

**Steps to Reproduce**
1. Create a root folder (name: `Folder A`)
2. Create an other folder in this root folder (name: `Folder B`)
3. Click on share on the `Folder B`
4. write an email not present in the database and invite him
5. You have an error saying you have not enough permission

lunika:
## Purpose

When a user with priviledged role on an item but inherited he can not invite external person. We refacto the invitation viewset, serializer and permission to check roles against all the tree.

## Proposal

- [x] 🐛(backend) allow inviting external person on item with no direct access

Fix #575

### DIFF
--- a/src/backend/core/api/permissions.py
+++ b/src/backend/core/api/permissions.py
@@ -1,13 +1,12 @@
 """Permission handlers for the drive core app."""
 
 from django.core import exceptions
-from django.db.models import Q
 from django.http import Http404
 
 from lasuite.drf.models.choices import PRIVILEGED_ROLES
 from rest_framework import permissions
 
-from core.models import ItemAccess, RoleChoices, get_trashbin_cutoff
+from core.models import RoleChoices, get_trashbin_cutoff
 
 ACTION_FOR_METHOD_TO_PERMISSION = {
     "versions_detail": {"DELETE": "versions_destroy", "GET": "versions_retrieve"},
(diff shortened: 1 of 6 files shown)


src/backend/core/tests/items/test_api_item_invitations.py
```diff
@@ -591,6 +591,36 @@ def test_api_item_invitations_create_cannot_invite_existing_users_case_insensiti
     }
 
 
+def test_api_items_invitations_on_item_without_explicit_access():
+    """Creating an invitation on an item withtou explicit access on it should work"""
```
NathanVss: typo

src/backend/core/api/permissions.py
```diff
@@ -72,27 +71,19 @@ def has_permission(self, request, view):
...
         # Apply permission checks only for creation (POST requests)
-        if view.action != "create":
-            return True
+        if view.action == "create":
+            # Check if resource_id is passed in the context
+            role = getattr(view, view.resource_field_name).get_role(user)
+            if role not in PRIVILEGED_ROLES:
+                raise exceptions.PermissionDenied(
+                    "You are not allowed to manage accesses for this resource."
+                )
```
NathanVss, line 87:
```suggestion
                raise exceptions.PermissionDenied(
                    "You are not allowed to manage invitations for this resource."
                )
```
?

src/backend/core/api/permissions.py
NathanVss: This method look the same as 

```
class ItemAccessPermission(IsAuthenticated):
    """Permission class for the ItemAccessViewSet."""

    def has_permission(self, request, view):
        """check create permission for accesses in documents tree."""
        if super().has_permission(request, view) is False:
            return False

        if view.action == "create":
            role = getattr(view, view.resource_field_name).get_role(request.user)
            if role not in PRIVILEGED_ROLES:
                raise exceptions.PermissionDenied(
                    "You are not allowed to manage accesses for this resource."
                )

        return True
```

Maybe it would be better to find a way to factorize it?

src/backend/core/models.py
```diff
@@ -1420,29 +1440,29 @@ def is_expired(self):
...
+        try:
+            roles = self.user_roles or []
+        except AttributeError:
+            roles = ItemAccess.objects.filter(
+                models.Q(user=user) | models.Q(team__in=user.teams),
+                item__path__ancestors=self.item.path,
+            ).values_list("role", flat=True)
+
+        return RoleChoices.max(*roles)
+
```
NathanVss, line 1457: this method look the same in Invitation, Access and Item, maybe should be factorized ?

src/backend/core/tests/items/test_api_item_invitations.py
lunika: I'm making so many typo lately. I'm exhausting myself.

src/backend/core/models.py
lunika: There are quite similar. I will do that in an other PR

PR #586, opened 2026-03-03
Hide pending items from listing views and improve upload progress

NathanVss:
- **Backend**: Exclude items with `upload_state=PENDING` from list, children, recents and search endpoints so that incomplete uploads are not visible to other users.
- **Frontend**: Scale the upload progress indicator to 0–90% during the actual upload, reserving the last 10% for the `upload-ended` confirmation instead of skipping 100% entirely.

### DIFF
--- a/src/backend/e2e/management/commands/e2e_fixture_search.py
+++ b/src/backend/e2e/management/commands/e2e_fixture_search.py
@@ -87,6 +87,9 @@ class Command(BaseCommand):
                 type=data["type"],
                 creator=data["creator"],
                 parent=parent,
+                update_upload_state=models.ItemUploadStateChoices.READY
+                if data["type"] == models.ItemTypeChoices.FILE
+                else None,
                 users=[(data["creator"], models.RoleChoices.OWNER)]
                 if depth == 0
                 else None,


src/backend/core/api/viewsets.py
```diff
@@ -378,6 +378,10 @@ def _filter_suspicious_items(self, queryset, user):
 
         return queryset.exclude(upload_state=models.ItemUploadStateChoices.SUSPICIOUS)
 
+    def _exclude_pending_items(self, queryset):
+        """Exclude items with PENDING upload_state from listing views."""
+        return queryset.exclude(upload_state=models.ItemUploadStateChoices.PENDING)
```
lunika, line 383: Is it not more efficient to add this line `348` in the definition of the queryset used by the viewset ?

lunika:
And you should add a management command cleaning item with upload_state `pending` older than `x` hours/days.

src/backend/core/api/viewsets.py
NathanVss: I could, I wanted to make sure we carefuly pick the route in which we want to exclude pending items. I'm going to make the changes.

NathanVss: I could not set it directly on the queryset attribute because some routes like "upload_ended" would not find pending items. 

So instead I defined it inside `get_queryset()` after the condition of details view.

src/backend/core/management/commands/clean_pending_items.py
```diff
@@ -0,0 +1,41 @@
...
+            upload_state=ItemUploadStateChoices.PENDING,
+            created_at__lt=threshold,
+            hard_deleted_at__isnull=True,
+        )
+
+        count = 0
+        for item in items.iterator():
+            item.soft_delete()
+            item.hard_delete()
+            process_item_deletion.delay(item.id)
```
lunika:
```suggestion
            item.delete()
```

in the deletion task, the file is deleted on the storage but here probably there is no file.
Maybe, the task ahould be improved with a try/catch block around the `default_storage.delete(item.file_key)` instruction. Can you open an issue with this ?

src/helm/drive/values.yaml
```diff
@@ -259,7 +259,13 @@ backend:
...
   #       - python manage.py flush --no-input
   ## @param backend.cronjobs Cronjob name, schedule, command
-  cronjobs: []
+  cronjobs:
+    - name: clean-pending-items
+      schedule: "30 */12 * * *"
+      command:
+        - "python"
+        - "manage.py"
+        - "clean_pending_items"
```
lunika:
```suggestion
      command:
        - "/bin/sh"
        - "-c"
        - python manage.py clean_pending_items
```

src/helm/drive/values.yaml
```diff
@@ -259,7 +259,13 @@ backend:
   #       - "-c"
   #       - python manage.py flush --no-input
   ## @param backend.cronjobs Cronjob name, schedule, command
-  cronjobs: []
+  cronjobs:
+    - name: clean-pending-items
+      schedule: "30 */12 * * *"
```
lunika: Once a day is probably enough ?

src/backend/core/management/commands/clean_pending_items.py
NathanVss: There: https://github.com/suitenumerique/drive/issues/600

PR #591, opened 2026-03-05
Update backend linters configuration

jmaupetit:
## Purpose

Compared to Pylint, Ruff is really efficient. Hence in a development workflow using only this linter (in check mode) is sufficient to produce valid code that will pass (all?) checks. But some rules are missing.

## Proposal

- [x] split ruff check and fix actions
- [x] configure ruff maximal line length to 100
- [x] activate pyflakes ruff rules
- [x] ~~activate pydocstyle rules~~ postponed to a new PR
- [x] ~~activate pycodestyle rules (`E`, `W`)~~ postponed to a new PR

### DIFF
--- a/src/backend/core/api/filters.py
+++ b/src/backend/core/api/filters.py
@@ -40,9 +40,7 @@ class WorkspacesChoices(TextChoices):
 class SearchItemFilter(ItemFilter):
     """Filter class dedicated to the Item viewset search method."""
 
-    workspace = django_filters.UUIDFilter(
-        method="filter_workspace", label=_("Workspace")
-    )
+    workspace = django_filters.UUIDFilter(method="filter_workspace", label=_("Workspace"))
 
     scope = django_filters.MultipleChoiceFilter(
         field_name="scopes",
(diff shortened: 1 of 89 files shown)


src/backend/core/apps.py
```diff
@@ -16,4 +16,3 @@ def ready(self):
         Import signals when the app is ready.
         """
         # pylint: disable=import-outside-toplevel, unused-import
-        from . import signals  # noqa: PLC0415
```
jmaupetit, line 19: Shouldn't be removed!

Makefile
```diff
@@ -201,9 +201,14 @@ lint-ruff-format: ## format back-end python sources with ruff
 
 lint-ruff-check: ## lint back-end python sources with ruff
 	@echo 'lint:ruff-check started…'
-	@$(COMPOSE_RUN_APP_NO_DEPS) ruff check . --fix
+	@$(COMPOSE_RUN_APP_NO_DEPS) ruff check .
 .PHONY: lint-ruff-check
 
+lint-ruff-check-fix: ## fix back-end python sources with ruff
+	@echo 'lint:ruff-check started…'
```
jmaupetit:
```suggestion
	@echo 'lint:ruff-check-fix started…'
```

jmaupetit:
As it may lead to huge changes, I've postponed Pydocstyle and Pycodestyle rules to a separated PR. We need to discuss outcomes/constraints internally first.

jmaupetit:
Also, Sonarqube suggestions make sense, but we need to tackle this separately (see #596).

PR #612, opened 2026-03-13
🐛(backend) fix hard delete of files created by other users

### LINKED ISSUE #604: We can't hard delete a file which is from someone else
## Bug Report

**Problematic behavior**
We can't hard delete a file which is from someone else, it gives a 403.

**Expected behavior/code**
We should be able to do so as long this file exists in our trashbin.

**Steps to Reproduce**
1. User A: create a folder and share it to user B
2. User B: create a file inside this folder and delete it
3. User A: go to trash and you should see the file user B deleted
4. User A: hard delete this file -> 403 error from /hard-delete route.

neilcroft:
## Purpose

  Fix #604 — A folder owner/admin gets a 403 when trying to hard delete a trashed file created by another user.

  ## Proposal

  The `can_hard_delete` ability used `is_owner_or_admin` which included a `not is_deleted` condition, making it always `False` for trashed items.

  - [x] Extract `is_owner_or_admin` as a pure role check (owner or admin, regardless of deletion state)
  - [x] Introduce `can_manage` for operations that require the item to not be deleted
  - [x] Use `is_owner_or_admin` in `can_hard_delete` so owners/admins can hard delete trashed items
  - [x] Add test covering the scenario

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -818,7 +818,7 @@ class Item(TreeModel, BaseModel):
         # Characteristics that are based only on specific access
         is_owner = role == RoleChoices.OWNER
         is_deleted = self.ancestors_deleted_at
-        is_owner_or_admin = (is_owner or role == RoleChoices.ADMIN) and not is_deleted
+        is_owner_or_admin = is_owner or role == RoleChoices.ADMIN
 
         # Compute access roles before adding link roles because we don't
         # want anonymous users to access versions (we wouldn't know from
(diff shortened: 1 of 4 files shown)


src/backend/core/models.py
```diff
@@ -838,7 +838,8 @@ def get_abilities(self, user):
             role = RoleChoices.max(role, link_definition["link_role"])
         can_get = bool(role) and not is_deleted
         retrieve = can_get or is_owner
-        can_update = (is_owner_or_admin or role == RoleChoices.EDITOR) and not is_deleted
+        can_manage = is_owner_or_admin and not is_deleted
+        can_update = (can_manage or role == RoleChoices.EDITOR) and not is_deleted
```
```suggestion
        can_update = (is_owner_or_admin or role == RoleChoices.EDITOR) and not is_deleted
```

should not changed ? there still the `and not is_deleted`

neilcroft: Yes you're right!

PR #628, opened 2026-03-31
Mirror files when boto3.copy_object is used

lunika:
## Purpose

When a file is renamed or duplicated. We don't use the storage backend to do that but directly the S3 client to use the [copy_object](https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/copy_object.html) to not transfer the file but to make the operation directly on the bucket.

Doing this, the copy on the bucket is not mirrored on the mirror bucket if this feature is enabled.

This PR add 2 commits, one for the rename function and another for the duplicate function

## Proposal

- [x] ✨(backend) mirror a file once renamed
- [x] ✨(backend) mirror a file when duplicated

### DIFF
--- a/src/backend/core/tasks/item.py
+++ b/src/backend/core/tasks/item.py
@@ -13,6 +13,7 @@ import botocore
 
 from core.api.utils import sanitize_filename
 from core.models import Item, ItemTypeChoices, ItemUploadStateChoices
+from core.tasks.storage import rename_file_on_mirroring_bucket
 
 from drive.celery_app import app
 
@@ -104,6 +105,8 @@ def rename_file(item_id, new_title):
         Bucket=default_storage.bucket_name,
         Key=from_file_key,
     )
+
+    rename_file_on_mirroring_bucket.delay(item.id, from_file_key, to_file_key)
 
 
 @app.task
--- a/src/backend/core/tasks/item.py
+++ b/src/backend/core/tasks/item.py
@@ -13,7 +13,7 @@ import botocore
 
 from core.api.utils import sanitize_filename
 from core.models import Item, ItemTypeChoices, ItemUploadStateChoices
-from core.tasks.storage import rename_file_on_mirroring_bucket
+from core.tasks.storage import duplicate_file_on_mirroring_bucket, rename_file_on_mirroring_bucket
 
 from drive.celery_app import app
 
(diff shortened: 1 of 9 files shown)


src/backend/core/tasks/storage.py
```diff
@@ -104,3 +104,43 @@ def mirror_file(self, mirror_task_id):
...
+
+@app.task
+def rename_file_on_mirroring_bucket(item_id, from_file_key, to_file_key):
+    """
+    Rename a file on the mirroring bucket. If the original file does not exist, mirror it.
+    Copy it otherwise
+    """
+    mirror_s3_client = get_mirror_s3_client()
+    if not mirror_s3_client:
+        logger.info(
```
jmaupetit, line 117: If mirroring is not configured on purpose, maybe this will pollute application's logs. Should the logger level be decreased to DEBUG?

src/backend/core/tasks/item.py
jmaupetit, line 1: `s/pugged/plugged` in the commit message body.

src/backend/core/tasks/storage.py
```diff
@@ -104,3 +104,43 @@ def mirror_file(self, mirror_task_id):
...
+
+    try:
+        mirror_s3_client.head_object(Bucket=mirror_bucket, Key=from_file_key)
+    except botocore.exceptions.ClientError:
+        item = Item.objects.get(pk=item_id)
+        mirror_task = MirrorItemTask.objects.create(
+            item=item, status=MirrorItemTaskStatusChoices.PENDING
+        )
+
+        mirror_file.delay(mirror_task.id)
```
jmaupetit, line 131: So the prerequisite is that the `file_key` (_e.g._ item's filename) has been updated first using the new filename, and thus, there is a sequence dependency between the `rename_file` and the `rename_file_on_mirroring_bucket` tasks.

src/backend/core/tasks/item.py
jmaupetit, line 1: `s/th e/the` in the second commit message body. Also a period is missing at the end of the second expression 😊

src/backend/core/tasks/storage.py
```diff
@@ -144,3 +144,37 @@ def rename_file_on_mirroring_bucket(item_id, from_file_key, to_file_key):
...
     )
+
+
+@app.task
+def duplicate_file_on_mirroring_bucket(duplicated_item_id, from_file_key, to_file_key):
+    """Duplicate a file on the mirror bucket by creating a copy of the original file."""
+
+    mirror_s3_client = get_mirror_s3_client()
+    if not mirror_s3_client:
+        logger.info(
```
jmaupetit, line 155: ditto (log level)

jmaupetit (commented):
I have the feeling that this (temporary) mirroring feature should not have been handled in the core project. Maybe we should consider removing it in favor of deploying a [rclone](https://rclone.org/)-based strategy?

PR #641, opened 2026-04-03
🐛(backend) sanitize slash in template-created filenames

neilcroft:
## Purpose

Titles containing '/' (e.g. "30/03/30 - liste à faire") produced a file_key with spurious path separators, crashing WOPI on open.

  Closes #626

## Proposal

Extract `format_template_filename()` to replace '/' with '-' when building the filename from a template title.

##  Alternatives considered                                                                                                                                                              

   ### Applying sanitize_filename() to template filenames

This existing function converts to ASCII, replaces spaces with underscores, and strips special characters. It would fix the /  issue but also mangle titles like "café résumé" into "cafe_resume", which diverges from what the user typed. Since the frontend displays title (not filename), the name would look fine in Drive but appear mangled in Collabora's editor (BaseFileName).                                                                                                               
                                                                                         
###  Aligning all filename paths to only replace / 
I explored replacing sanitize_filename() with the minimal `/ -> -` replacement in the rename task and update validation as well (not just template creation). This would make filenames consistent across all paths (template, rename, WOPI) and preserve unicode/accents everywhere. I decided against it for now as it  changes behavior on existing rename and update paths, and S3 keys would start containing characters they didn't before. The current fix is scoped to the template creation path where the bug occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to
 ### Changed
 
 - 🐛(backend) replace VersionId by Etag for WOPI
+- 🐛(backend) sanitize slash in template-created filenames
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


kernicPanel:
Could you please rebase this PR ?

kernicPanel (approved):

PR #644, opened 2026-04-06
purge deleted items on a daily basis using a cronjob

sampaccoud:
## Purpose

Over time, deleted items that remain in a soft or hard deleted state, can unnecessarily consume storage (S3 as well as database). These expired items can no longer be restored and should be permanently removed.

This PR introduces a mechanism to identify and purge such expired deleted items, ensuring that only relevant and recoverable data is retained while preventing unbounded growth of stale data.

## Proposal

- [x] Identify items that have been deleted and whose retention period has expired (soft delete cutoff period + grace period before purge).
- [x] Permanently remove these items from storage
- [x] Ensure the cleanup process is safe, idempotent, and does not affect active or recently deleted items
- [x] Integrate this cleanup into a scheduled cronjob to continuously maintain storage hygiene

This change improves storage efficiency, reduces unnecessary data retention, and aligns the system with expected lifecycle management of deleted resources.


src/backend/core/tasks/item.py
```diff
@@ -20,33 +23,51 @@
...
+            reason = f"soft-deleted but not past purge cutoff: {root.deleted_at.isoformat()}"
+
+        logger.info("Item %s is not eligible for purge: %s", item_id, reason)
+        return
 
-    if item.type == ItemTypeChoices.FOLDER:
-        for child in item.children():
-            process_item_deletion.delay(child.id)
+    # Get descendants, leaf first. Don't burst memory
+    for item in Item.objects.filter(path__startswith=root.path).order_by("-path").iterator():
```
lunika:
```suggestion
    for item in Item.objects.filter(path__descendants=root.path).order_by("-path").iterator():
```

src/backend/core/management/commands/purge_deleted_items.py
```diff
@@ -0,0 +1,39 @@
...
+        is_purgeable = Q(
+            deleted_at__lte=timezone.now()
+            - timedelta(days=settings.TRASHBIN_CUTOFF_DAYS + settings.PURGE_GRACE_DAYS)
+        )
+
+        count = 0
+        for item_id in Item.objects.filter(is_hard_deleted | is_purgeable).values_list(
+            "id", flat=True
+        ):
+            process_item_purge(item_id)
```
lunika:
```suggestion
            process_item_purge.delay(item_id)
```

lunika (approved):

PR #655, opened 2026-04-09
✨(frontend) preserve empty folders on drag & drop upload

### LINKED ISSUE #126: You can't make an empty folder tree
## Bug Report

**Problematic behavior**
You can't make an empty folder tree

PanchoutNathan:
## Summary

- Add a custom `getFilesFromEvent` for react-dropzone that walks the
  drop tree ourselves and emits a zero-byte marker file for every
  empty folder, so empty directories dropped into the explorer are
  now created instead of being silently dropped.
- Update `useUpload` to materialize the folder chain from markers via
  `getFolderByPath`, then filter them out before any size validation
  or backend upload. When a drop contains only markers, the file
  upload toast is suppressed and a dedicated success toast is shown
  once the folders have been created.
- Add unit tests covering marker creation, the batched
  `readEntries()` loop, recursive entry walking, and the
  `customGetFilesFromEvent` entry point.

Input change events (folder/file buttons) remain delegated to
`file-selector`: the browser does not expose empty folders through
`<input webkitdirectory>`, matching the Google Drive / Proton Drive
limitation.

## Test plan

- [ ] Drop an empty folder into the explorer and confirm it is
      created with the success toast.
- [ ] Drop a folder tree mixing empty and non-empty subfolders and
      confirm both the files and the empty folders land correctly.
- [ ] Drop a single file and confirm the regular upload toast still
      shows progress.
- [ ] Use the "Import folder" button with a non-empty folder and
      confirm the existing behavior is unchanged.
- [ ] Run the new unit tests in
      `dropTraversal.test.ts`.

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/utils/__tests__/dropTraversal.test.ts
+++ b/src/frontend/apps/drive/src/features/explorer/utils/__tests__/dropTraversal.test.ts
@@ -1,0 +1,302 @@
+import {
+  createEmptyFolderMarker,
+  EMPTY_FOLDER_MARKER_NAME,
+  getFileFromEntry,
+  isEmptyFolderMarker,
+  readAllEntries,
+  traverseEntry,
+} from "../dropTraversal";
+
+// ---------------------------------------------------------------------------
+// Fake FileSystemEntry helpers (no DataTransfer / no DragEvent involved)
+// ---------------------------------------------------------------------------
+
+const makeFileEntry = (
+  fullPath: string,
+  content = "",
+): FileSystemFileEntry => {
+  const name = fullPath.split("/").pop() ?? "";
+  const file = new File([content], name, { type: "text/plain" });
+  return {
+    isFile: true,
+    isDirectory: false,
+    name,
+    fullPath,
+    file: (success: FileCallback) => success(file),
+  } as unknown as FileSystemFileEntry;
+};
+
+const makeFailingFileEntry = (
+  fullPath: string,
+  error: unknown,
+): FileSystemFileEntry => {
+  const name = fullPath.split("/").pop() ?? "";
+  return {
+    isFile: true,
+    isDirectory: false,
+    name,
+    fullPath,
+    file: (_success: FileCallback, errorCb?: ErrorCallback) => {
+      errorCb?.(error as DOMException);
+    },
+  } as unknown as FileSystemFileEntry;
+};
+
+/** Single-batch directory: readEntries returns children, then []. */
+const makeDirEntry = (
+  fullPath: string,
+  children: FileSystemEntry[],
+): FileSystemDirectoryEntry => {
+  const name = fullPath.split("/").pop() ?? "";
+  return {
+    isFile: false,
+    isDirectory: true,
+    name,
+    fullPath,
+    createReader: () => makeReader([children, []]),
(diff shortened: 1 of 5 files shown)


src/frontend/apps/drive/src/features/explorer/utils/dropTraversal.ts
```diff
@@ -0,0 +1,210 @@
...
+  // 'drop' event the DataTransferItem objects become inert after the
+  // first microtask tick, so calling webkitGetAsEntry() on later items
+  // after an await returns null — which silently drops every item past
+  // the first. We therefore pull all entries up front, then walk them
+  // asynchronously with Promise.all (mirroring what file-selector does
+  // via `items.map(toFilePromises)`).
+  const entries: FileSystemEntry[] = [];
+  for (const item of items) {
+    const entry = item.webkitGetAsEntry();
+    if (entry) entries.push(entry);
```
NathanVss, line 205: In which cases entry is not truthy?

src/frontend/apps/drive/src/features/explorer/utils/dropTraversal.ts
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * Input change events (the "Import folder/files" buttons) are delegated
+ * to file-selector's default behavior. The browser does not expose empty
+ * folders through <input webkitdirectory>, so this is an accepted
+ * limitation matching Google Drive / Proton Drive.
+ */
+
+export const EMPTY_FOLDER_MARKER_NAME = ".empty-folder";
+
+const EMPTY_FOLDER_MARKER_TYPE = "application/x-empty-folder-marker";
```
NathanVss, line 26: Seems not used, then not needed

src/frontend/apps/drive/src/features/explorer/utils/__tests__/dropTraversal.test.ts
NathanVss, line 1: Please also add a e2e test 🙏

NathanVss (approved):
Neat !! :)

src/frontend/apps/drive/src/features/explorer/utils/dropTraversal.ts
PanchoutNathan: webkitGetAsEntry() returns FileSystemEntry | null per spec. Even though we check that the method exists on all items (supportsEntries above), the method can still return null in edge cases — for example, browser security restrictions on certain drag origins, or unsupported item types. The synchronous snapshot loop (lines 195-201) already mitigates the main risk (items becoming inert after the first microtask), but the null-check remains necessary as a defensive guard against these browser-level edge cases.

src/frontend/apps/drive/src/features/explorer/utils/__tests__/dropTraversal.test.ts
PanchoutNathan: Unfortunately this isn't feasible with Playwright. The feature relies on DataTransferItem.webkitGetAsEntry(), which is a native browser API that Playwright cannot simulate — its dispatchEvent() and dragAndDrop() don't create real DataTransferItem objects with working webkitGetAsEntry() support. The existing e2e upload tests use page.waitForEvent("filechooser") which only covers input-based uploads, not native drag-and-drop.

PR #662, opened 2026-04-13
🐛(frontend) Range selection freezes when there are many items in the list

### LINKED ISSUE #124: Range selection freezes when there are many items in the list
## Bug Report

**Problematic behavior**
I had the case in the trash, because I had deleted a lot of elements and I tried to select several

https://github.com/user-attachments/assets/e050bb4c-b645-4a30-a297-2898a1cad0dd

PanchoutNathan:
## Purpose

cf #124

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
@@ -1,19 +1,18 @@
 import { Item, ItemType, ItemUploadState } from "@/features/drivers/types";
 import {
   createContext,
-  Dispatch,
-  SetStateAction,
   useCallback,
   useContext,
   useMemo,
   useRef,
   useState,
 } from "react";
+import { useSelectionStore } from "@/features/explorer/stores/selectionStore";
 import { useTranslation } from "react-i18next";
 import {
   CellContext,
   createColumnHelper,
-  flexRender,
+  Row,
 } from "@tanstack/react-table";
 import { useReactTable } from "@tanstack/react-table";
 import { getCoreRowModel } from "@tanstack/react-table";
@@ -32,7 +31,6 @@ import { EmbeddedExplorerGridActionsCell } from "@/features/explorer/components/
 import { useTableKeyboardNavigation } from "@/features/explorer/hooks/useTableKeyboardNavigation";
 import clsx from "clsx";
 import { isTablet } from "@/features/ui/components/responsive/ResponsiveDivs";
-import { Droppable } from "@/features/explorer/components/Droppable";
 import { useDragItemContext } from "@/features/explorer/components/ExplorerDndProvider";
 import { useModal } from "@gouvfr-lasuite/cunningham-react";
 import { ExplorerMoveFolder } from "@/features/explorer/components/modals/move/ExplorerMoveFolderModal";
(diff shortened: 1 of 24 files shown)


PERF_SELECTION_REFACTOR.md
NathanVss, line 1: This doc was ultra useful for the review. So I think you should make a proper english version that resume our approach on this pattern as a doc inside an architecture folder at the root of the repo under `docs/...` so we could in the future tell why we implemented this approach.

src/frontend/apps/drive/src/features/explorer/stores/selectionStore.ts
```diff
@@ -0,0 +1,175 @@
...
+    nextMap.forEach((_, id) => {
+      if (!prevMap.has(id)) changedIds.add(id);
+    });
+
+    changedIds.forEach((id) => {
+      const listeners = this.idListeners.get(id);
+      if (listeners) listeners.forEach((l) => l());
+    });
+
+    this.globalListeners.forEach((l) => l());
```
NathanVss, line 83: Should'n' this call be bypassed if there are no changeIds?

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
```diff
@@ -78,11 +75,12 @@ export type EmbeddedExplorerGridProps = {
 const EMPTY_ARRAY: Item[] = [];
 const columnHelper = createColumnHelper<Item>();
 
-type EmbeddedExplorerGridContextType = EmbeddedExplorerGridProps & {
-  selectedItemsMap: Record<string, Item>;
-  openMoveModal: () => void;
-  closeMoveModal: () => void;
```
NathanVss, line 84: damn :)

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
```diff
@@ -220,23 +211,166 @@ export const EmbeddedExplorerGrid = (props: EmbeddedExplorerGridProps) => {
...
+        items: getItemActionMenuItems(row.original),
+      });
+    },
+    [
+      props.displayMode,
+      selectionStore,
+      contextMenu,
+      getItemActionMenuItems,
+    ],
+  );
```
NathanVss, line 362: Would be great to put that into the row component

src/frontend/apps/drive/src/features/explorer/components/modals/move/ExplorerMoveFolderModal.tsx
```diff
@@ -87,28 +87,35 @@ export const ExplorerMoveFolder = ({
   const moveConfirmationModal = useModal();
   const createFolderModal = useModal();
 
+  // Subscribe to the embedded explorer's local selection store so the modal
+  // reactively updates when the user picks a target folder inside it.
+  const localSelectedItems = useSyncExternalStore(
+    itemsExplorer.selectionStore.subscribe,
+    itemsExplorer.selectionStore.getSelectedItems,
+    itemsExplorer.selectionStore.getSelectedItems,
```
NathanVss, line 95: Is the third argument needed?

NathanVss (approved):
Nice improvement 🙏

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGrid.tsx
PanchoutNathan: Thought about it and decided to keep them in the parent for now. The handlers depend on a lot of parent-scoped things (selectionStore, table, canSelect, contextMenu, getItemActionMenuItems, and several grid props). Pushing them into the row means either drilling all of that through props (noisy and fragile for the memo) or adding a dedicated row context. They're already useCallback-stable, so there's no perf win — only a readability trade-off. Happy to revisit in a follow-up if we grow a second grid variant or want to unit-test the row

PR #665, opened 2026-04-14
🐛(front) show actual selection count in hard delete modal

### LINKED ISSUE #651: Trash hard-delete disclaimer wrong number
## Bug Report

**Problematic behavior**
In the trash when hard deleting this modal always display "2 elements" even if we selected more than 2.

Image

**Expected behavior/code**
It should show the amount of selected items.

PanchoutNathan:
## Summary
- The trash hard delete confirmation modal always read "2 items"
  regardless of the selection size, because the modal received a
  boolean `multiple` prop and hardcoded `count: 2` for the i18n
  interpolation.
- Replace the `multiple` boolean with a `count` number prop and pass
  `selectedItems.length` from the trash page so the plural message
  reflects the real selection.

Closes #651

## Test plan
- [ ] Open the trash view, select 3+ items, click hard delete and
      confirm the modal reads "delete these 3 items forever?".
- [ ] Select a single item and confirm the modal reads
      "delete this item forever?".
- [ ] Run frontend type-check / lint.

### DIFF
--- a/src/frontend/apps/drive/src/features/explorer/components/modals/HardDeleteConfirmationModal.tsx
+++ b/src/frontend/apps/drive/src/features/explorer/components/modals/HardDeleteConfirmationModal.tsx
@@ -9,10 +9,10 @@ import { useTranslation } from "react-i18next";
 
 export const HardDeleteConfirmationModal = ({
   onDecide,
-  multiple = false,
+  count = 1,
   ...props
 }: DecisionModalProps & {
-  multiple?: boolean;
+  count?: number;
 }) => {
   const { t } = useTranslation();
   return (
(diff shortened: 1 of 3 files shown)


src/frontend/apps/drive/src/features/explorer/components/modals/HardDeleteConfirmationModal.tsx
```diff
@@ -45,7 +45,7 @@ export const HardDeleteConfirmationModal = ({
     >
       <div className="c__modal__content__text">
         {t("explorer.trash.hard_delete.content", {
-          count: multiple ? 2 : 1,
```
NathanVss, line 48: wtf 😂

NathanVss (approved):
quick win

PR #666, opened 2026-04-14
🐛(front) set size and variant on trash navigate modal

PanchoutNathan:
## Summary
- Pass `messageType: VariantType.INFO` and `size: ModalSize.MEDIUM` to the
  trash navigate message modal so it renders with the expected styling.

## Test plan
- [ ] Navigate into a trashed file/folder and confirm the modal displays
      with the medium size and info variant.


NathanVss (approved):
Indeed it fixes the issue but imho this fix should also happen in ui-kit as there is an issue with modal without any specific messageType, the icon is still rendered as undefined, causing this ghost gap

Image

PR #668, opened 2026-04-14
⬆️(dependencies) pin pillow to v12.2.0 [SECURITY]

NathanVss:
## Summary

- Pillow 12.1.1 is flagged for CVE-2026-40192 — a FITS GZIP
  decompression bomb that can be triggered by a crafted image.
- We don't depend on pillow directly: it comes in through
  `easy_thumbnails`, which has no upper bound, so a plain version bump
  isn't an option.
- Added a `[tool.uv] constraint-dependencies` entry pinning
  `pillow>=12.2.0` so any transitive pull picks up the patched release.


src/backend/pyproject.toml
```diff
@@ -92,6 +92,14 @@ dev = [
     "types-requests==2.32.4.20260107",
 ]
 
+[tool.uv]
+constraint-dependencies = [
+    # Pin pillow above the version vulnerable to CVE-2026-40192
+    # (FITS GZIP decompression bomb). Pulled transitively via
+    # easy_thumbnails, which has no upper bound on pillow.
+    "pillow>=12.2.0",
```
qbey, line 100: Not sure we need this, since this is the purpose of the lock file.

qbey (approved):

src/backend/pyproject.toml
NathanVss: I was concerned by the fact that the next `uv lock|add|remove` would override the value in the lock file automatically 🤔 Do you think we are safe or not doing this only in the lock file?

qbey: I don't think uv would downgrade a dependency except if a lib is requiring it explicitly. I don't have a strong opinion on this, you are right in the fact that it's explicit and if some dependency tries to downgrade, it will raise directly, instead of being detected afterward. So it's a good idea ;)

PR #673, opened 2026-04-20
✨(backend) make invitation validity duration configurable via env

NathanVss:
People using Drive, from Mon Espace Collectivite were facing invitation expiration too early for their process. By making this setting configurable, we let consumers decide their expiration period.

- Expose `INVITATION_VALIDITY_DURATION` as an environment variable instead of a hardcoded 7-day constant, so operators can tune invitation lifetime per deployment.
- Add a test asserting `Invitation.is_expired` honors the configured duration.

Default behavior is unchanged (604800 seconds / 7 days).

### DIFF
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -1005,7 +1005,9 @@ class Base(Configuration):
     EMAIL_FROM = values.Value("from@example.com")
 
     AUTH_USER_MODEL = "core.User"
-    INVITATION_VALIDITY_DURATION = 604800  # 7 days, in seconds
+    INVITATION_VALIDITY_DURATION = values.PositiveIntegerValue(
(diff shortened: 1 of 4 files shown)


src/backend/drive/settings.py
```diff
@@ -1005,7 +1005,9 @@ class Base(Configuration):
     EMAIL_FROM = values.Value("from@example.com")
 
     AUTH_USER_MODEL = "core.User"
-    INVITATION_VALIDITY_DURATION = 604800  # 7 days, in seconds
+    INVITATION_VALIDITY_DURATION = values.Value(
```
```suggestion
    INVITATION_VALIDITY_DURATION = values.IntegerValue(
```

Or maybe `PositiveIntegerValue` but it excludes the `0` value, so I don't know if it's what we want.

lunika (approved):

Issue #638, opened 2026-04-02
Versioning in Drive


robin-lecomte:
<img width="2978" height="1022" alt="Image" src="https://github.com/user-attachments/assets/b0f0b6e0-12cc-4cfd-92fa-607307879c7d" />

Full versioning functionality for the Drive app, with the ability to view and manage different versions of a document. Applies to all file types, but not to folders.

Details:  
- New option in the “More” menu to open the history modal  
- History modal allowing users to browse all versions, preview, download, restore, and delete them  
- Selection modal displayed when uploading a document with the same name as an existing one, with the option to increment the version

[FIGMA](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=14674-34720&t=rWcFvxxcSfPVAzrO-1)

PR #689, opened 2026-04-24
✨(backend) manage wopi convert action

lunika:
## Purpose

The actual WOPI implementation only managed the `edit` action. Some WOPI applications can edit legacy binary file format (`.doc`, `.xls`, `.ppt`, etc) but others are respecting the WOPI protocol and must [convert](https://api.onlyoffice.com/docs/docs-api/using-wopi/editing-binary-documents/) them before edition.

To use the `convert` action, we need to implement the `PutRelativeFile` method to save the converted file.

Documentation used to implement this PR and needed to review the PR

https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#wopi-actions
https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/scenarios/conversion
https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile

-----

After team discussions, the target UX is "always duplicate, keep the original intact", but this PR is not exactly this.
See https://github.com/suitenumerique/drive/pull/689#issuecomment-4425567655

## Proposal

- [x] ✨(backend) implement wopi PutRelativeFile action
- [x] ✨(backend) create a helper to get base url app
- [x] ♻️(backend) manage view and convert action in the wopi view
- [x] ♻️(backend) save all available actions for the wopi configuration

-----
  
- [x] ✨(backend) expose available wopi actions on items
- [x] ♻️(backend) duplicate file when converting via PutRelativeFile
- [x] 🔧(dev) force convert legacy formats for OnlyOffice in dev


src/backend/wopi/tasks/configure_wopi.py
```diff
@@ -74,31 +74,42 @@ def _configure_wopi_client_from_discovery(client, discovery_url):
...
                 mimetype = app.get("name")
 
                 if mimetype in settings.WOPI_EXCLUDED_MIMETYPES:
                     continue
 
-                wopi_configuration["mimetypes"][mimetype] = {
-                    "url": action.get("urlsrc"),
-                    "client": client,
-                }
+                mimetypes = wopi_configuration["mimetypes"].get(mimetype, {})
```
NathanVss: Naming this `mimetypes` var `actions` wouldn't make more sense?

src/backend/wopi/tasks/configure_wopi.py
```diff
@@ -74,31 +74,42 @@ def _configure_wopi_client_from_discovery(client, discovery_url):
...
                 extension = action.get("ext")
 
                 if extension in settings.WOPI_EXCLUDED_EXTENSIONS:
                     continue
 
-                wopi_configuration["extensions"][extension] = {
-                    "url": action.get("urlsrc"),
-                    "client": client,
-                }
+                extensions = wopi_configuration["extensions"].get(extension, {})
```
NathanVss: Ditto

src/backend/core/api/viewsets.py
```diff
@@ -1559,9 +1560,24 @@ def wopi(self, request, *args, **kwargs):
...
+                {"detail": "This item does not support WOPI integration."}
             )
 
+        action = None
+        abilities = item.get_abilities(request.user)
+        can_update = abilities["update"]
+
+        if wopi_client.get(WopiActions.EDIT):
+            action = wopi_client[WopiActions.EDIT]
+        elif can_update and wopi_client.get(WopiActions.CONVERT):
```
NathanVss, line 1581: Are we sure that we want only people with write access to convert a doc?

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+
+        item.title = target_filename
+        item.filename = target_filename
+        item.mimetype = detect_mimetype(
+            file_content.read(2048 if file_content.size > 2048 else file_content.size),
+            filename=target_filename,
+        )
+        item.size = file_content.size
+
+        default_storage.save(item.file_key, file_content)
```
NathanVss: WDYT about keeping the original uploaded file?

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
             )
 
         return Response(status=200)
+
+    def _put_relative_file(self, request, pk=None):
```
NathanVss, line 432: According to this doc: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile

Keeping `UserCanNotWriteRelative` true means that `PutRelativeFile` is not supported. Which is what we always return in `retrieve`

NathanVss, line 432: According to the doc this route should handle lock and 409 on locked file

NathanVss (approved):
Great work, approving do not block you, open to discuss the comments. Especially the one to let read visitors to trigger conversion so they can consult the documents.

src/backend/core/api/viewsets.py
```diff
@@ -1559,9 +1560,24 @@ def wopi(self, request, *args, **kwargs):
...
+            action = wopi_client[WopiActions.CONVERT]
+
+        if not can_update and wopi_client.get(WopiActions.VIEW):
+            action = wopi_client[WopiActions.VIEW]
+
+        if not action:
+            raise drf.exceptions.ValidationError({"detail": "This item has no action compatible."})
```
jmaupetit:
```suggestion
            raise drf.exceptions.ValidationError({"detail": "This item has no compatible action."})
```

src/backend/core/tests/test_utils_get_url_app.py
```diff
@@ -0,0 +1,29 @@
...
+
+from core.utils import get_url_app
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture(autouse=True)
+def reset_get_url_app_cache():
+    """reset the get_url_app cache."""
+    get_url_app.cache_clear()
```
jmaupetit: Why not defining this fixture in the project's `conftest.py`?

jmaupetit: Also if you want to perform this action during test teardown, you can use the yield pattern:

```suggestion
    yield
    get_url_app.cache_clear()
```

Every instruction before the yield statement correspond to the test setup and after, the teardown.

In this case every test starts with a fresh cache.

src/backend/core/tests/test_utils_get_url_app.py
```diff
@@ -0,0 +1,29 @@
...
+        None,  # Test fallback to Site domain
+    ],
+)
+def test_get_url_app(url_app, settings):
+    """test the get_url_app expected output."""
+    settings.URL_APP = url_app
+
+    expected_url_app = url_app or "example.com"
+
+    assert get_url_app() == expected_url_app
```
jmaupetit: Maybe it would be more explicit to have a `(url_app, expected)` tuple as test parameter?

jmaupetit: Also, can you test that the utility properly uses the cache?

docs/env.md
```diff
@@ -114,6 +114,7 @@ This document lists all configurable environment variables for the Drive applica
 | `SPECTACULAR_SETTINGS_ENABLE_DJANGO_DEPLOY_CHECK` | Enable Django deploy check in Spectacular | `False` |
 | `STORAGES_STATICFILES_BACKEND` | Backend for static files storage | `whitenoise.storage.CompressedManifestStaticFilesStorage` |
 | `TRASHBIN_CUTOFF_DAYS` | Number of days before items are automatically removed from trash after their soft deletion | `30` |
+| `URL_APP` | The base url of the application | None |
```
jmaupetit: Shouldn't it be named `APP_URL` instead?

src/backend/core/utils/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Module for common utils in the project."""
+
+from functools import cache
+
+from django.conf import settings
+from django.contrib.sites.models import Site
+
+
+@cache
+def get_url_app():
```
jmaupetit: ditto?

```suggestion
def get_app_url():
```

src/backend/drive/settings.py
```diff
@@ -145,6 +145,12 @@ class Base(Configuration):
 
     SITE_ID = 1
 
+    URL_APP = values.Value(
+        default=values.Value(default=None, environ_name="EMAIL_URL_APP"),
```
jmaupetit, line 149: Should we understand that this is a fallback to the previous name of the setting?

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+
+        https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile
+        """
+        suggested_target = request.META.get(HTTP_X_WOPI_SUGGESTEDTARGET)
+        relative_target = request.META.get(HTTP_X_WOPI_RELATIVETARGET)
+        file_conversion = request.META.get(HTTP_X_WOPI_FILECONVERSION)
+
+        if not file_conversion:
+            return Response(
+                data={"error": {"details": "only file conversion is supported."}}, status=400
```
jmaupetit:
```suggestion
                data={"error": {"details": "Only file conversion is supported."}}, status=400
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+            )
+
+        # Exactly one header must be present; they are mutually exclusive
+        if bool(suggested_target) == bool(relative_target):
+            return Response(
+                data={
+                    "error": {
+                        "details": (
+                            f"both {HTTP_X_WOPI_SUGGESTEDTARGET} and {HTTP_X_WOPI_RELATIVETARGET}"
```
jmaupetit:
```suggestion
                            f"{HTTP_X_WOPI_SUGGESTEDTARGET} and {HTTP_X_WOPI_RELATIVETARGET}"
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+        relative_target = request.META.get(HTTP_X_WOPI_RELATIVETARGET)
+        file_conversion = request.META.get(HTTP_X_WOPI_FILECONVERSION)
+
+        if not file_conversion:
+            return Response(
+                data={"error": {"details": "only file conversion is supported."}}, status=400
+            )
+
+        # Exactly one header must be present; they are mutually exclusive
+        if bool(suggested_target) == bool(relative_target):
```
jmaupetit:
```suggestion
        if bool(suggested_target) and bool(relative_target):
```

or 

```suggestion
        if all(map(bool, [suggested_target, relative_target]):
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+                        )
+                    }
+                },
+                status=400,
+            )
+
+        try:
+            file_content = ContentFile(request.body)
+        except RequestDataTooBig:
+            return Response(status=413)
```
jmaupetit: When it comes to rarely used code, maybe it would be more explicit to use DRF statuses (_e.g._ `rest_framework.status`)?

```suggestion
            return Response(status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE)
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+
+        item = request.auth.item
+
+        if suggested_target:
+            logger.info("putRelativeFile in suggest_target mode")
```
jmaupetit: As there is not that much context in this log, maybe it would only pollute log files and should be decreased to the debug level?

```suggestion
            logger.debug("putRelativeFile in suggested_target mode")
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+            target = suggested_target.encode("ascii").decode("utf-7")
+            if target.startswith("."):
+                base, _ = splitext(item.filename)
+                target_filename = f"{base}{target}"
+            else:
+                target_filename = target
+
+        else:
+            logger.info("putRelativeFile in relative_target mode")
```
jmaupetit: ditto

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+        try:
+            file_content = ContentFile(request.body)
+        except RequestDataTooBig:
+            return Response(status=413)
+
+        item = request.auth.item
+
+        if suggested_target:
+            logger.info("putRelativeFile in suggest_target mode")
+            target = suggested_target.encode("ascii").decode("utf-7")
```
jmaupetit: Wow, first time I see UTF-7. 

- Is this related to WOPI? 
- Shouldn't it be configurable?

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+            if target.startswith("."):
+                base, _ = splitext(item.filename)
+                target_filename = f"{base}{target}"
+            else:
+                target_filename = target
+
+        else:
+            logger.info("putRelativeFile in relative_target mode")
+            # Specific mode: host must use the exact name provided
+            target_filename = relative_target.encode("ascii").decode("utf-7")
```
jmaupetit: ditto

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+
+        item.title = target_filename
+        item.filename = target_filename
+        item.mimetype = detect_mimetype(
+            file_content.read(2048 if file_content.size > 2048 else file_content.size),
```
jmaupetit: Also, 2048 should be defined or configured somewhere?

src/backend/core/tests/items/test_api_items_wopi.py
```diff
@@ -293,7 +313,104 @@ def test_api_items_wopi_authenticated_user_item_mimetype_not_supported():
...
+            }
+        ],
+        "type": "validation_error",
+    }
+
+
+def test_api_items_wopi_authenticated_user_item_extension_convert_action(
+    timestamp_now, valid_wopi_launch_url
+):
+    """Authenticated user can access item that extension is supported for convert action."""
```
jmaupetit:
```suggestion
    """Authenticated user can access item with a supported extension for convert action."""
```

src/backend/core/tests/items/test_api_items_wopi.py
```diff
@@ -293,7 +313,104 @@ def test_api_items_wopi_authenticated_user_item_mimetype_not_supported():
...
+        f"&closebutton=false&lang={user.language}"
+    )
+
+
+def test_api_items_wopi_authenticated_user_item_extension_view_action(
+    timestamp_now, valid_wopi_launch_url
+):
+    """
+    Authenticated user can access item that extension is supported without update permission
+    should fallback on view action.
```
jmaupetit:
```suggestion
    Authenticated user can access item with a supported extension but without update permission.
    
    In this case, we should fallback on the view action.
```

src/backend/core/api/viewsets.py
```diff
@@ -1559,9 +1560,24 @@ def wopi(self, request, *args, **kwargs):
...
+        action = None
+        abilities = item.get_abilities(request.user)
+        can_update = abilities["update"]
+
+        if wopi_client.get(WopiActions.EDIT):
+            action = wopi_client[WopiActions.EDIT]
+        elif can_update and wopi_client.get(WopiActions.CONVERT):
+            action = wopi_client[WopiActions.CONVERT]
+
+        if not can_update and wopi_client.get(WopiActions.VIEW):
```
jmaupetit, line 1585: Maybe add a comment here to explain that this is the fallback mode if the item is compatible with the view action (AFAIU)?

src/backend/core/tests/items/test_api_items_wopi.py
```diff
@@ -293,7 +313,104 @@ def test_api_items_wopi_authenticated_user_item_mimetype_not_supported():
...
+    assert data["access_token"] is not None
+    assert data["access_token_ttl"] > timestamp_now
+    assert data["launch_url"] == (
+        f"{valid_wopi_launch_url}/view/?WOPISrc={wopi_src}&closebutton=false&lang={user.language}"
+    )
+
+
+def test_api_items_wopi_authenticated_user_items_extension_no_action():
+    """
+    Authenticated user who can access to an item and is wopi compaitble but without available
```
jmaupetit:
```suggestion
    Authenticated user who can access to an item and is WOPI-compatible but without any
```

src/backend/wopi/viewsets.py
```diff
@@ -397,3 +404,85 @@ def _rename_file(self, request, pk=None):
...
+                base, _ = splitext(item.filename)
+                target_filename = f"{base}{target}"
+            else:
+                target_filename = target
+
+        else:
+            logger.info("putRelativeFile in relative_target mode")
+            # Specific mode: host must use the exact name provided
+            target_filename = relative_target.encode("ascii").decode("utf-7")
+
```
jmaupetit, line 494: So suggested and relative targets perform the exact same action right?

src/backend/wopi/tests/viewset/test_put_relative_file.py
```diff
@@ -0,0 +1,472 @@
...
+        HTTP_AUTHORIZATION=f"Bearer {access_token}",
+        headers={
+            "X-WOPI-Override": "PUT_RELATIVE",
+            "X-WOPI-SuggestedTarget": suggested,
+            "X-WOPI-FileConversion": "1",
+        },
+    )
+
+    assert response.status_code == 200
+    assert "access_token=" in response.json()["Url"]
```
jmaupetit: Maybe you can parse the URL to better get the access token?

src/backend/wopi/tests/viewset/test_put_relative_file.py
```diff
@@ -0,0 +1,472 @@
...
+        headers={
+            "X-WOPI-Override": "PUT_RELATIVE",
+            "X-WOPI-FileConversion": "1",
+        },
+    )
+
+    assert response.status_code == 400
+    assert response.json()["error"]["details"] == (
+        "both HTTP_X_WOPI_SUGGESTEDTARGET and HTTP_X_WOPI_RELATIVETARGET are mutually exclusive."
+    )
```
jmaupetit, line 435: This shows that you must test if any of those expected header is provided in your viewset and, if not, an explicit error message must be provided.

jmaupetit (commented):
Wow, that's a big improvement! 👏

src/backend/core/api/viewsets.py
lunika: the wopi edit action is not related to the drive user role. The edit action is using a read only view when the user is in read only

src/backend/wopi/viewsets.py
lunika: It was the first implementation I made, a kind of duplication creating a new item. And finally, I choose to revert it. 
We should talk about this with the team, it can be a UX decision.

lunika:
> conversion modulo modale d'information/consentement. dupliquer ou éditer/écraser.

- Expose to the frontend application all the WOPI actions available for the given file. (If an entry exists for the extension or mimetype, just return the dictionary keys).
- Modify the ItemViewset::wopi action to trigger the desired action. In this PR I try to guess the appropriate action to use between "edit" and "convert." This choice should be made by the front-end application now.
- Find a way to persist the user choice between duplicate or replace the original `.doc` from the modal to the `putRelativeFile` action.

robin-lecomte:
RL 2026-05-06 at 17 11 41@2x

After discussions with the team, I suggest this user flow:
a. Double-click a doc, xls, or ppt file  
b. A modal prompts to convert the file before opening  
c. The file is converted and pre-selected in the index  
d. The user must double-click the converted file to open it

The converted file does not inherit the sharing settings of the original file (similar to duplication).

For the record, I was against asking the user to decide whether to keep a copy, to avoid cognitive overload each time they open this type of file. Here, we offer a predictable behavior that preserves the original document and requires no decision.  
Ultimately, integrating this into a file’s history (https://github.com/suitenumerique/drive/issues/638) would be the best UX approach.

[FIGMA](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=15252-1478&t=pSEeVRorvJBjCNyw-1)

src/backend/drive/settings.py
kernicPanel: I believe this is kept for backward compatibility.

src/backend/wopi/viewsets.py
kernicPanel: `if suggested_target and relative_target:` should work I believe.

src/backend/wopi/viewsets.py
kernicPanel: It looks like utf7 is needed for those wopi headers. A comment has been added to clarify it.

src/backend/wopi/viewsets.py
kernicPanel: I've extracted it to a constant, as I don't think a setting is relevant here.

src/backend/wopi/viewsets.py
kernicPanel: Almost: suggested lets the host adjust the name, relative is used directly.

src/backend/drive/settings.py
lunika: Yes, it is kept for backward compatibility. You can remove it in version 1.0.0

src/backend/wopi/viewsets.py
kernicPanel: I'm not sure to get your point. Can you clarify it please ?

kernicPanel:
> > conversion modulo modale d'information/consentement. dupliquer ou éditer/écraser.
> 
>     * Expose to the frontend application all the WOPI actions available for the given file. (If an entry exists for the extension or mimetype, just return the dictionary keys).

done

>     * Modify the ItemViewset::wopi action to trigger the desired action. In this PR I try to guess the appropriate action to use between "edit" and "convert." This choice should be made by the front-end application now.
>  
>     * Find a way to persist the user choice between duplicate or replace the original `.doc` from the modal to the `putRelativeFile` action.
 
Since the UX changed, I think those are not relevant.

kernicPanel:
> RL 2026-05-06 at 17 11 41@2x
> 
> After discussions with the team, I suggest this user flow: a. Double-click a doc, xls, or ppt file   b. A modal prompts to convert the file before opening   c. The file is converted and pre-selected in the index   d. The user must double-click the converted file to open it
> 
> The converted file does not inherit the sharing settings of the original file (similar to duplication).
> 
> For the record, I was against asking the user to decide whether to keep a copy, to avoid cognitive overload each time they open this type of file. Here, we offer a predictable behavior that preserves the original document and requires no decision. Ultimately, integrating this into a file’s history (#638) would be the best UX approach.
> 
> [FIGMA](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=15252-1478&t=pSEeVRorvJBjCNyw-1)

This PR lays down the foundation for the conversion, but the UX is not exactly the one you asked for.
For now, when a user double-clicks on a .doc file, a new tab opens prompting the user to directly edit the file : 
image
Meanwhile, the conversion has been added to Drive: 
image
If we want to skip the new tab part, we should add a server to server call between django and onlyoffice to silently convert the file. But I think this should be done in a separate PR.

kernicPanel:
@lunika @NathanVss @jmaupetit  Can I rebase on main and squash, or do you prefer to keep the fixup commits for now ?

src/backend/wopi/viewsets.py
NathanVss: From here: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile#response-headers

```
[409 Conflict](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.10) – Target file already exists or the file is locked. If the target file is locked, you must include an X-WOPI-Lock response header containing the value of the current lock on the file.
```

kernicPanel: Thanks for your explanation. I added a squashed fixup for this.

robin-lecomte:
It’s unfortunate that we can’t follow the UX specifically designed to clearly show the user that their file was duplicated before being opened. Reversing this flow makes the entire experience more confusing. On top of that, if the user doesn’t want to convert the file, they’re forced to switch contexts twice (open the file in a new OnlyOffice tab, avoid clicking “convert,” then go back to Drive).

In short, the UX is weaker. But if the technical constraint is too strong, I understand. I’ll let @lterm decide on this new constrained behavior. I’ve archived [the UX proposal](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=15252-1478&t=PYfFx8kHawWqRNLq-1) accordingly.

src/backend/wopi/viewsets.py
```diff
@@ -475,31 +480,37 @@ def _put_relative_file(self, request, pk=None):
             else:
                 target_filename = target
 
+            target_filename = self._disambiguate_target(siblings, target_filename)
```
lunika: I think this is not needed, the `create_child` method handles it.

src/backend/wopi/viewsets.py
```diff
@@ -475,31 +480,37 @@ def _put_relative_file(self, request, pk=None):
...
+
         else:
             logger.debug("putRelativeFile in relative_target mode")
-            # Specific mode: host must use the exact name provided
+            # Specific mode: host must use the exact name provided.
             target_filename = relative_target.encode("ascii").decode("utf-7")
+            if conflict := self._find_relative_target_conflict(siblings, target_filename):
+                lock_value = LockService(conflict).get_lock(default="")
+                return Response(status=409, headers={X_WOPI_LOCK: lock_value})
```
lunika: the lock header can be removed IMO. 

This is what is present in the documentation : 

```
If the X-WOPI-OverwriteRelativeTarget request header is set to true and a file with the specified name already exists and is locked, the host must respond with a [409 Conflict](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.10) and include an X-WOPI-Lock response header containing the value of the current lock on the file.
```

We are not using the ` X-WOPI-OverwriteRelativeTarget` header, so we don't have to include the lock header, IMO.

src/backend/wopi/viewsets.py
```diff
@@ -508,10 +519,105 @@ def _put_relative_file(self, request, pk=None):
...
+        return target
+
+    @staticmethod
+    def _find_relative_target_conflict(siblings, target):
+        """
+        Return the non-deleted sibling whose title matches `target`, or None
+        when no conflict. Uses the same rule as `Item.objects.create_child`
+        so a sibling folder bearing that title is also a conflict.
+        """
+        return siblings.filter_non_deleted(title=target).order_by("created_at").first()
```
lunika:
```suggestion
        return siblings.filter_non_deleted(title=target).order_by("created_at").exists()
```

`exists` is faster and with my previous comment we don't need the item instance anymore.

env.d/development/common
```diff
@@ -82,6 +82,9 @@ WOPI_SRC_BASE_URL=http://app-dev:8000
 # Disable rename for Collabora by default because it does not post messages when renaming a file.
 # So the frontend cannot reflect renaming operations.
 WOPI_COLLABORA_OPTIONS={"SupportsRename": False}
+# OnlyOffice cannot edit legacy binary formats reliably; force conversion to
+# their OpenXML equivalent on these extensions/mimetypes.
```
lunika:
```suggestion
their Office Open XML (OOXML) equivalent on these extensions/mimetypes.
```

src/backend/wopi/viewsets.py
kernicPanel: create_child handles title uniqueness, but not the filename.

src/backend/wopi/viewsets.py
```diff
@@ -402,3 +417,227 @@ def _rename_file(self, request, pk=None):
...
+        siblings = self._resolve_put_relative_siblings(parent, request.user)
+
+        try:
+            if suggested_target:
+                target_filename = self._resolve_suggested_target_filename(
+                    item, siblings, suggested_target
+                )
+            else:
+                target_filename = self._resolve_relative_target_filename(siblings, relative_target)
+        except PutRelativeTargetConflict:
```
lunika: This exception is only raised by `_resolve_relative_target_filename` so the `try/except` block should be only around `target_filename = self._resolve_relative_target_filename(siblings, relative_target)`

src/backend/wopi/tasks/configure_wopi.py
```diff
@@ -55,10 +55,9 @@ def _configure_wopi_client_from_discovery(client, discovery_url):
             f"wopi client {client} is invalid"
         )
 
-    wopi_configuration = cache.get(
-        WOPI_CONFIGURATION_CACHE_KEY,
-        default=WOPI_DEFAULT_CONFIGURATION,
-    )
+    wopi_configuration = cache.get(WOPI_CONFIGURATION_CACHE_KEY)
+    if wopi_configuration is None:
+        wopi_configuration = {"mimetypes": {}, "extensions": {}}
```
jmaupetit: More condensed?

```suggestion
    wopi_configuration = cache.get(WOPI_CONFIGURATION_CACHE_KEY) or {"mimetypes": {}, "extensions": {}}
```

src/backend/wopi/tasks/configure_wopi.py
```diff
@@ -84,21 +88,41 @@ def _configure_wopi_client_from_discovery(client, discovery_url):
...
+                actions.update(
+                    {
+                        action_name: {
+                            "url": action.get("urlsrc"),
+                            "client": client,
+                        }
+                    }
+                )
+
+                wopi_configuration["extensions"][extension] = actions
```
jmaupetit: Maybe this whole block can be refactored as it's the same pattern whether we are dealing with an extension or a mime-type?

src/backend/core/tests/conftest.py
```diff
@@ -30,9 +30,13 @@ def clear_cache():
     from core.storage import (  # pylint:disable=import-outside-toplevel # noqa: PLC0415
         get_storage_compute_backend,
     )
+    from core.utils import (  # pylint:disable=import-outside-toplevel # noqa: PLC0415
+        get_app_url,
+    )
```
jmaupetit: Why importing from here and not from top-level? You have a circular import somewhere?

src/backend/wopi/viewsets.py
```diff
@@ -410,3 +420,106 @@ def _rename_file(self, request, pk=None):
...
+
+        # WOPI requires X-WOPI-SuggestedTarget and X-WOPI-RelativeTarget headers
+        # to be UTF-7 encoded. See the spec:
+        # https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile
+        if suggested_target:
+            logger.debug("putRelativeFile in suggest_target mode")
+            target = suggested_target.encode("ascii").decode("utf-7")
+            if target.startswith("."):
+                base, _ = splitext(item.filename)
+                target_filename = f"{base}{target}"
```
jmaupetit: Using `pathlib.PurePath` from stdlib:

```suggestion
                target_filename = PurePath(item.filename).with_suffix(target)
```

src/backend/wopi/tests/viewset/test_put_relative_file.py
```diff
@@ -39,6 +41,10 @@ def test_put_relative_file_suggested_target_extension_only():
     service = AccessUserItemService()
     access_token, _ = service.insert_new_access(item, user)
     default_storage.save(item.file_key, BytesIO(b"original content"))
+    original_file_key = item.file_key
+    original_title = item.title
+    original_mimetype = item.mimetype
+    original_size = item.size
```
jmaupetit: Why not keeping an explicit copy?

```suggestion
    original = copy.copy(item)
```

jmaupetit (commented):
I only have minor comments, else 👍 

---
Side note:

I find it hard to review such work. Some comments might be outdated as you refactor new code from one commit to another on the same PR. So we cannot review it globally nor on a per-commit basis.

IMO we should find a way to increase the production/review cycle by defining new collaboration flows. Let's discuss this IRL!

src/backend/core/tests/conftest.py
kernicPanel: I confess I did that to match existing code, without thinking about it.

jmaupetit (approved):
Image

PanchoutNathan (approved):

kernicPanel:
Superseded by #719

PR #696, opened 2026-04-29
✨ Track column type change and item duplicate posthog events

NathanVss:
Adds two new product analytics events and the e2e plumbing to verify them:

- `column_type_changed` (frontend) — fired when a user switches one of the customizable explorer columns.
- `item_duplicate` (backend) — fired on item duplication request.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1641,6 +1641,8 @@ class ItemViewSet(
             duplicated_item_id=duplicated_item.id,
         )
 
+        posthog_capture("item_duplicate", user, {}, item=duplicated_item)
+
         serializer = self.get_serializer(duplicated_item)
         return drf.response.Response(serializer.data, status=drf.status.HTTP_201_CREATED)
 
--- a/src/backend/core/tests/items/test_api_items_duplicate.py
+++ b/src/backend/core/tests/items/test_api_items_duplicate.py
@@ -407,3 +407,33 @@ def test_api_items_duplicate_deleted_item():
     response = client.post(f"/api/v1.0/items/{item.id!s}/duplicate/")
 
     assert response.status_code == 403
+
+
+# Posthog events
+
+
+def test_api_items_duplicate_posthog_event(settings):
+    """Duplicating an item should send an 'item_duplicate' event."""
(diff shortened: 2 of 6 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -1642,6 +1642,8 @@ def duplicate(self, request, *args, **kwargs):
             duplicated_item_id=duplicated_item.id,
         )
 
+        posthog_capture("item_duplicate", user, {}, item=duplicated_item)
```
kernicPanel, line 1644: IMO we should use "item_duplicated" to match other event names.
```suggestion
        posthog_capture("item_duplicated", user, {}, item=duplicated_item)
```

src/backend/core/tests/items/test_api_items_duplicate.py
```diff
@@ -407,3 +407,39 @@ def test_api_items_duplicate_deleted_item():
...
+    assert response.status_code == 201
+
+    mock_capture.assert_called_once()
+    args, kwargs = mock_capture.call_args
+    assert args[0] == "item_duplicate"
+    assert args[1] == user
+    assert args[2] == {}
+    duplicated_item = kwargs["item"]
+    assert duplicated_item.id != item.id
+    assert duplicated_item.creator == user
```
kernicPanel: Those assertions could be simplified with:
```suggestion
    duplicated_item = models.Item.objects.get(id=response.json()["id"])
    mock_capture.assert_called_once_with(
        "item_duplicate", user, {}, item=duplicated_item
    )
```

src/backend/core/api/viewsets.py
NathanVss: It is on purpose as it only triggers before the duplication, so at this point the item is not duplicated. Shortly: we want to record duplicate requests. If the duplication fails for any reasons, we stil want the event

PanchoutNathan (approved):

src/backend/core/api/viewsets.py
kernicPanel: Oh sorry, I didn't see the "delay".

kernicPanel (approved):

PR #697, opened 2026-04-29
🐛(backend) better transaction management on duplicate action

lunika:
## Purpose

On the duplicate action we are using the decorator @transaction.atomic to be sure that the duplicated item and its accesses are correctly created. The problem with that is that the celery task doing the file duplication on the object storage can be executed before the transaction is commited leading to an error saying that the duplicated_item does not exists.

## Proposal

Description...

- [x] 🐛(backend) better transaction management on duplicate action

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to
 ### Fixed
 
 - 🐛(backend) accept CDFV2 mimetype from newer libmagic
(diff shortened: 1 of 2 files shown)


jmaupetit (approved):

NathanVss (approved):
Good catch! It may be the bug we faced on the ANCT instance, I'll double check that

PR #699, opened 2026-04-30
🔊(backend) log wopi max expected file size in get_file_content action

lunika:
## Purpose

In the get_file_content action we want to log the received max expected file size when this one is lower than the file size the user wants to edit. This will help us to correctly configure the WOPI host.

## Proposal

- [x] 🔊(backend) log wopi max expected file size in get_file_content action

### DIFF
--- a/src/backend/wopi/viewsets.py
+++ b/src/backend/wopi/viewsets.py
@@ -126,6 +126,11 @@ class WopiViewSet(viewsets.ViewSet):
         head_object = get_item_file_head_object(item)
         if max_expected_size:
             if int(head_object["ContentLength"]) > int(max_expected_size):
+                logger.info(
+                    "get_file_content: file size %s exceeds X-WOPI-MaxExpectedSize header value %s",
+                    int(head_object["ContentLength"]),
+                    int(max_expected_size),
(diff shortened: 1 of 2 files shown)


src/backend/wopi/viewsets.py
```diff
@@ -126,6 +126,14 @@ def _get_file_content(self, request, pk=None):
...
         if max_expected_size:
             if int(head_object["ContentLength"]) > int(max_expected_size):
+                logger.info(
+                    "get_file_content: The file size is higher than the X-WOPI-MaxExpectedSize "
+                    "header value. %r",
+                    {
+                        "file_size": int(head_object["ContentLength"]),
+                        "max_expected_size": int(max_expected_size),
+                    },
+                )
```
kernicPanel, line 133: IMO, using %s would be more consistent with existing logs.

```suggestion
                logger.info(
                    "get_file_content: file size %s exceeds X-WOPI-MaxExpectedSize header value %s",
                    int(head_object["ContentLength"]),
                    int(max_expected_size),
                )
```

kernicPanel (approved):

PR #700, opened 2026-05-04
♻️(frontend) move the file preview to ui-kit

NathanVss:
The file preview (PDF, image, video, audio, WOPI, error/unsupported, controls, navigation) now lives in the ui-kit. We drop the in-app implementation under `features/ui/preview/` and keep only a thin `CustomFilesPreview` wrapper that maps `Item` to `FilePreviewType` and plugs in the share modal, info sidebar and posthog event.

The local file icon catalog (MIME_MAP, KNOWN_EXTENSIONS, etc.) is also gone, replaced by the ui-kit `FileIcon`. We keep only the drive-specific bits (folders, suspicious uploads) through `getItemExtendedIcon`.

Detailed preview behavior is now covered by Playwright CT in the ui-kit, so the drive e2e suite only keeps smoke checks: upload → open → render for PDF, image, video, audio, plus prev/next navigation.


src/frontend/apps/drive/src/features/explorer/utils/mimeTypes.ts
```diff
@@ -1,251 +1,18 @@
...
-  if (mimetype?.startsWith("audio/")) {
-    return MimeCategory.AUDIO;
-  }
-  if (mimetype?.startsWith("video/")) {
-    return MimeCategory.VIDEO;
-  }
-
-  return MimeCategory.OTHER;
-};
+export { getMimeCategory, ICONS, MimeCategory, removeFileExtension };
```
PanchoutNathan: For what ? If it's not going to change everywhere it's used, that's not a good reason for me haha

PanchoutNathan (approved):

PanchoutNathan:
Can you add a description for this PR plz ?

src/frontend/apps/drive/src/features/explorer/utils/mimeTypes.ts
NathanVss: Indeed, refactored automated

PR #709, opened 2026-05-14
(backend) ensure WOPI integration does not depend on versioning

RaitoBezarius:
## Purpose

Versioning in S3 is not implemented by all S3 implementations. Drive is mostly versioning independent except for its WOPI integration where it assumes `VersionId` to exist or be useable in a lot of places.

This has been tested with Garage.

## Proposal

Just ignore `VersionId` where it is not available.


lunika:
Hi,
Thank you for your contribution. This PR is a duplicate of #687 
We have to finish it and merge it.

PR #710, opened 2026-05-15
🐛(media-auth) strip X-Amz-* query params before SigV4 re-signing

etiquet:
## Purpose

Every `GET /media/<key>` returns **403 SignatureDoesNotMatch** when the Drive is deployed behind an nginx-ingress that uses the documented `auth-url` / `auth-response-headers` pattern against `/api/v1.0/items/media-auth/`.

## Root cause

`generate_s3_authorization_headers()` (`core/api/utils.py`) uses `boto3.generate_presigned_url("get_object")` to produce a URL, then re-signs that URL header-based with `S3SigV4Auth`.

The URL returned by `generate_presigned_url()` **always** contains the X-Amz-* query parameters (`X-Amz-Algorithm`, `X-Amz-Credential`, `X-Amz-Date`, `X-Amz-Expires`, `X-Amz-SignedHeaders`, `X-Amz-Signature`). These get hashed into the **canonical_query_string** component of the signed canonical request.

But nginx-ingress' `auth-response-headers` annotation only propagates the 3 SigV4 headers (`Authorization`, `X-Amz-Date`, `X-Amz-Content-SHA256`) to the upstream S3 request — the original X-Amz-* query params are **not** propagated. S3 thus computes a canonical request with empty `canonical_query_string`, gets a different hash, and rejects the request:

```xml
<Error><Code>SignatureDoesNotMatch</Code>...</Error>
```

## Verified on

- **Scaleway S3** (fr-par region), bucket configured via `AWS_S3_ENDPOINT_URL=https://s3.fr-par.scw.cloud` (path-style addressing). Tested against `lasuite/drive-backend:latest` shipped 2026-04-19.

## Fix

```python
url = url.split("?", 1)[0]
```

Stripped the query string before constructing the `AWSRequest`. The canonical_query_string is now empty, matching what nginx sends to S3.

## Test gap

The existing test suite (`test_api_items_media_auth.py`) only asserts the presence of `"AWS4-HMAC-SHA256 Credential="` on the `Authorization` header — it never actually exercises the SigV4 against a real S3 endpoint. End-to-end the bug was hidden.

## Out of scope but related

- `nginx.ingress.kubernetes.io/upstream-vhost` must be aligned with what boto3 produces (path-style: `s3.<region>.scw.cloud`, virtual-hosted: `<bucket>.s3.<region>.scw.cloud`). Misalignment between addressing styles is a separate cause of `SignatureDoesNotMatch`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


etiquet:
pushed pr on the wrong repos, sorry.

PR #711, opened 2026-05-15
Export entire folder recursively

### LINKED ISSUE #412: Export entire folder recursively
## Feature Request

Today in drive, it is only possible to download individual files
However, when a folder contains many files or multiple levels of subfolders and files, there is no simple way to export the full hierarchy in one action.

Users who organize their work with deep folder structures (projects, audits, reports, archives, etc.) often need to download the entire tree. Having to manually download each subfolder and file is time-consuming and error-prone.

## Possible solution

Add an option to export a folder recursively, including:
- all nested subfolders
- all contained files
- preserved directory structure

The export could be provided as a single .zip archive containing the full folder tree

kernicPanel:
## Purpose

Export entire folder recursively

Fixes #412.

## Proposal

- [x] Add an `export` ability for accessible folders
- [x] Add the `GET /api/v1.0/items/<id>/export/` endpoint
- [x] Stream the ZIP archive without loading all files in memory
- [x] Preserve the folder hierarchy through relative archive paths
- [x] Exclude soft-deleted descendants and files that are not ready

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -18,6 +18,7 @@ from django.db import IntegrityError, transaction
 from django.db import models as db
 from django.db.models.expressions import RawSQL
 from django.db.models.functions import Coalesce
+from django.http import StreamingHttpResponse
 from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.functional import cached_property
@@ -46,6 +47,7 @@ from rest_framework_api_key.permissions import HasAPIKey
 
 from core import enums, models
 from core.entitlements import get_entitlements_backend
+from core.services.item_exports import build_zip_stream, export_descendants
 from core.services.sdk_relay import SDKRelayManager
 from core.services.search_indexers import (
     get_file_indexer,
@@ -77,8 +79,6 @@ MEDIA_STORAGE_URL_PATTERN = re.compile(
     f"{settings.MEDIA_URL:s}(?P<preview>preview/)?"
     f"(?P<key>{ITEM_FOLDER:s}/(?P<pk>{UUID_REGEX:s})/.*{FILE_EXT_REGEX:s})$"
 )
-
-
 # pylint: disable=too-many-ancestors
 
 
@@ -1528,6 +1528,23 @@ class ItemViewSet(
         return drf.response.Response(
             status=status.HTTP_302_FOUND,
             headers={"Location": redirect_url},
+        )
+
+    @drf.decorators.action(detail=True, methods=["get"], url_path="export")
+    def export(self, request, *args, **kwargs):
+        """
+        Stream a recursive ZIP archive of a folder's content.
+        """
+        folder = self.get_object()
+
+        descendants = export_descendants(folder)
+        zip_stream = build_zip_stream(descendants)
+
+        encoded_name = quote(f"{folder.title}.zip", safe="")
+        return StreamingHttpResponse(
+            zip_stream,
+            content_type="application/zip",
+            headers={"Content-Disposition": f"attachment; filename*=UTF-8''{encoded_name}"},
         )
 
     @drf.decorators.action(detail=False, methods=["get"], url_path="media-auth")
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1539,10 +1539,12 @@ class ItemViewSet(
         descendants = export_descendants(folder)
         zip_stream = build_zip_stream(descendants)
 
-        response = StreamingHttpResponse(zip_stream, content_type="application/zip")
         encoded_name = quote(f"{folder.title}.zip", safe="")
-        response["Content-Disposition"] = f"attachment; filename*=UTF-8''{encoded_name}"
-        return response
+        return StreamingHttpResponse(
+            zip_stream,
+            content_type="application/zip",
+            headers={"Content-Disposition": f"attachment; filename*=UTF-8''{encoded_name}"},
+        )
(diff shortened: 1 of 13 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -1529,6 +1529,21 @@ def download(self, request, *args, **kwargs):
...
+        """
+        folder = self.get_object()
+
+        descendants = export_descendants(folder)
+        zip_stream = build_zip_stream(descendants)
+
+        response = StreamingHttpResponse(zip_stream, content_type="application/zip")
+        encoded_name = quote(f"{folder.title}.zip", safe="")
+        response["Content-Disposition"] = f"attachment; filename*=UTF-8''{encoded_name}"
+        return response
```
lunika:
```suggestion
        encoded_name = quote(f"{folder.title}.zip", safe="")
        return StreamingHttpResponse(
            zip_stream,
            content_type="application/zip",
            headers = {
                "Content-Disposition": f"attachment; filename*=UTF-8''{encoded_name}"
            }
        )
```

src/backend/core/services/item_exports.py
```diff
@@ -0,0 +1,59 @@
+"""Service for exporting item folders as streaming ZIP archives."""
+
+from django.core.files.storage import default_storage
+
+from zipstream import ZipStream
+
+from core import models
+
+DEFAULT_STORAGE_READ_CHUNK_SIZE = 64 * 1024
```
lunika: seems huge, the `StreqmingBody` function of botocore as for default value `1024`, we can probably follow what they are doing.

src/backend/core/services/item_exports.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def iter_storage_chunks(file_key, chunk_size=DEFAULT_STORAGE_READ_CHUNK_SIZE):
```
lunika, line 12: You can maybe take a look at this function https://github.com/boto/botocore/blob/develop/botocore/response.py#L165 and take the exit case they have if the chunk is an empty string.

src/backend/core/services/item_exports.py
```diff
@@ -0,0 +1,59 @@
...
+    """Yield bytes from object storage without buffering the whole file."""
+    with default_storage.open(file_key, "rb") as fh:
+        while chunk := fh.read(chunk_size):
+            yield chunk
+
+
+def export_descendants(folder):
```
lunika, line 19: This function is working while we keep the inheritance mechanism in "open mode" : You can't restrict role on descendants, you can only open them, so for now if you access a folder, you can access all of the items it contains. But in the road map we should remove this restriction. 
We can maybe anticipate it by checking it ? But be aware that calling the `get_role` on each item is subject to the `N+1` sql query and the annotation `annotate_user_roles` should probably be used to avoid it.

src/backend/core/tests/items/test_api_items_export.py
```diff
@@ -0,0 +1,258 @@
...
+
+    assert response.status_code == 200
+    assert _zip_names(response) == [keep.filename]
+
+
+@pytest.mark.parametrize(
+    "upload_state",
+    [
+        models.ItemUploadStateChoices.PENDING,
+        models.ItemUploadStateChoices.DUPLICATING,
```
lunika:
```suggestion
        state
        for state in models.ItemUploadStateChoices.values
        if state != models.ItemUploadStateChoices.READY
```

Something like this to test all upload state and removing ready ?

src/backend/core/tests/items/test_api_items_export.py
```diff
@@ -0,0 +1,258 @@
...
+        title="été 2026",
+        users=[(user, models.RoleChoices.OWNER)],
+    )
+
+    response = client.get(f"/api/v1.0/items/{folder.pk}/export/")
+
+    assert response.status_code == 200
+    disposition = response["Content-Disposition"]
+    assert disposition.startswith("attachment;")
+    assert "filename*=UTF-8''" in disposition
```
lunika: why not asserting the complete `disposition` value?



```suggestion
    assert disposition == "attachment; filename*=UTF-8'' ..."
```

src/backend/core/services/item_exports.py
kernicPanel: Empty chunk is already managed with the walrus operator. Otherwise the tests should fail.

src/backend/core/services/item_exports.py
kernicPanel: I got your point, but I think we should manage this with the permission update, in another PR.

kernicPanel: As discussed, a note has been added to remember your point.

lunika (approved):
Awesome feature, congrats.

src/frontend/apps/drive/src/features/explorer/hooks/useItemActionMenuItems.tsx
```diff
@@ -164,6 +165,14 @@ export const useItemActionMenuItems = ({
           handleDownloadItem(item);
         },
       },
+      {
+        icon: <Download />,
+        label: t("explorer.item.actions.download"),
```
PanchoutNathan: Maybe export is more explicit ?

PanchoutNathan (commented):
A really cool feature. One small bug before validating its functionality: the complete directory structure isn't displayed. If I have a folder A with a folder A.1 and a file X, if I export A, I only get file X by itself.

PanchoutNathan (approved):

Issue #597, opened 2026-03-10
Enhance topbar filters


lterm:
## Feature Request

Currently, a Type filter exists to distinguish between folders and files. We propose a broader overhaul of the filtering system in the topbar.
We want to add filters by Filetype categories, Contact, Date modified, and Source.

<img width="852" height="256" alt="Image" src="https://github.com/user-attachments/assets/c8685d47-e74a-4a1a-8e22-3ca5bfd0d1bc" />

**Proposal**

Proposed filters: 

- **File type** 

Replace the current binary folder/file toggle with file type categories:
Text documents: .doc, .docx, .odt, .txt...
Spreadsheets: .xls, .xlsx, .ods, .csv...
Presentations : .ppt, .pptx, .odp...
PDF 
Images : png, .jpg, .svg, .gif, tiff...
Videos  .mp4, .mov, .avi...
Archives .zip, .tar, .gz...
Audio .mp3, .wav, .ogg...
Others : any unmatched MIME type

- **Contact**

Filter files by people involved in sharing, regardless of direction (shared by or shared with). 
The filter returns files where the selected contact is involved in any sharing relationship with the current user.

UX question: Should the default state show a contact search input, or pre-populate with the most frequently interacted collaborators?

- **Date modified**

Filter by last modification date using preset ranges: Today / Yesterday / Last 7 days / Last 30 days / Custom range (date picker)

- **Source**

Filter files by their origin/integration source: Visio (recordings), Docs, Grist, Mail (email attachments) (extensible as new integrations are added)
This is especially relevant to highlight upcoming interoperability features with external tools.
It will only be visible once the corresponding interoperability features are available. It is documented here to inform the design system early and ensure future compatibility.

**Describe alternatives you've considered**

**Discovery, Documentation, Adoption, Migration Strategy**

- Filters should be combinable.
- Active filters should be clearly visible and individually dismissable.

robin-lecomte:
[Here is the mockup](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=9171-8091&t=HgYth0ZOO7FgLdHd-1)

**File type**
_Keep fewer than 10 categories to comply with Hick’s Law + grouping categories semantically_
Text document: .doc, .docx, .odt, .txt, .rtf, .md, .pages
Slides: .ppt, .pptx, .odp, .key
Spreadsheet: .xls, .xlsx, .ods, .csv, .tsv, .numbers
PDF: .pdf
[...]

PR #712, opened 2026-05-15
Enhance filters

kernicPanel:
## Purpose

Currently, a Type filter exists to distinguish between folders and files.
#597 proposes a broader overhaul of the filtering system in the topbar,
adding filters by file type categories, contact, date modified and source,
so users can find items faster.

## Proposal

Add three combinable topbar filters, available in every explorer listing
(list, children, recents and favorites):

- [x] filter by file type category (document, slides, spreadsheet, pdf, image, video, audio, archive, other), based on filename extension — folders are always kept so navigation is not broken
- [x] filter by shared contact, matching items shared with or by the selected user; a new `/users/contacts/` endpoint lists the current user's frequent contacts to pre-populate the picker
- [x] filter by modification date, with preset ranges and a custom date picker
- [x] enrich `create_demo` with file type and sharing fixtures to exercise the filters

The Source filter is out of scope: it depends on interoperability features
not yet available.

Part of #597


src/backend/core/tests/test_api_users_contacts.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def test_api_users_contacts_anonymous():
+    """Anonymous users should not be allowed to list their contacts."""
+    client = APIClient()
+    response = client.get("/api/v1.0/users/contacts/")
+    assert response.status_code == 401
+
+
+def test_api_users_contacts():
```
lunika, line 41: At lest one test should have an assertion against the complete response json to check what is return

src/backend/core/api/viewsets.py
```diff
@@ -252,6 +252,40 @@ def get_me(self, request):
...
+        frequency = db.Count("itemaccess", filter=shared_with)
+        contacts = (
+            models.User.objects.filter(is_active=True)
+            .filter(shared_with | shared_by)
+            .exclude(pk=user.pk)
+            .annotate(frequency=frequency)
+            .order_by("-frequency", "email")[: settings.API_USERS_LIST_LIMIT]
+        )
+
+        return drf.response.Response(self.get_serializer(contacts, many=True).data)
```
lunika, line 308: I don't know if all the user information should be returned or only the `full_name` and `short_name` to not leak PII

src/backend/core/api/filters.py
```diff
@@ -1,26 +1,92 @@
...
         model = models.Item
-        fields = ["title", "type"]
+        fields = ["title", "type", "category", "contact", "updated_at"]
+
+    @staticmethod
+    def _extensions_q(extensions):
+        """Build a Q matching filenames ending with one of the given extensions."""
+        matched = Q()
+        for extension in extensions:
+            matched |= Q(filename__iendswith=extension)
```
lunika: Not sure this can scale or if we should extract the extension in a dedicated column

kernicPanel: I think it will be ok. Let's put this on real world context before rethinking this.

src/backend/demo/tests/test_commands_create_demo.py
```diff
@@ -32,3 +33,50 @@ def test_commands_create_demo():
     # assert dev users have doc accesses
     user = models.User.objects.get(email="drive@drive.world")
     assert models.ItemAccess.objects.filter(user=user).exists()
+
+
+@override_settings(DEBUG=True)
+def test_commands_create_demo_shares_dev_users_items_with_demo_users():
+    """The create_demo command should share existing dev users items with demo users."""
+    call_command("create_demo")
```
NathanVss, line 42: I'd like to assert before the `call_command` call that there is not items in db.

src/backend/demo/management/commands/create_demo.py
```diff
@@ -128,7 +188,22 @@ def create_items(users):
         yield create_item(user)
 
 
-def create_demo(stdout):
+def create_file_type_items(user):
+    """Create one ready file for each file type category described in issue #597."""
+    for file_type_item in FILE_TYPE_ITEMS:
+        yield create_item(
```
NathanVss, line 204: What is the use of a generator for now?

src/backend/core/enums.py
```diff
@@ -10,3 +10,39 @@
...
+        ".tiff",
+        ".webp",
+        ".bmp",
+        ".heic",
+        ".ico",
+    ],
+    "video": [".mp4", ".mov", ".avi", ".mkv", ".webm", ".wmv"],
+    "audio": [".mp3", ".wav", ".ogg", ".aac", ".flac", ".m4a"],
+    "archive": [".zip", ".tar", ".gz", ".rar", ".7z", ".bz2"],
+}
```
NathanVss, line 25: Keeping the dot `.` is maybe not needed?

I am not against using extension, but by curiosity: is there a reason to not use mime type?

src/backend/core/api/viewsets.py
```diff
@@ -690,7 +690,7 @@ def list(self, request, *args, **kwargs):
         filter_data = filterset.form.cleaned_data
 
         # Filter as early as possible on fields that are available on the model
-        for field in ["is_creator_me", "title", "type"]:
+        for field in ["is_creator_me", "title", "type", "category"]:
```
NathanVss: I am starting to think  this loop with exhaustive filters needed to be defined is not necessarily a good pattern, we are doing this only because we want to apply the favorite filter after the annotation. WDYT?

We could do smth like: 

```
EARLY_FILTERS = [f for f in filterset.filters if f != "is_favorite"]
for field in EARLY_FILTERS:
    queryset = filterset.filters[field].filter(queryset, filter_data[field])
```

lunika (approved):

src/backend/core/api/viewsets.py
```diff
@@ -252,6 +252,40 @@ def get_me(self, request):
...
+        # yields at least one item.
+        shared_items = models.Item.objects.filter(
+            db.Q(accesses__user=user) | db.Q(accesses__team__in=user.teams),
+            hard_deleted_at__isnull=True,
+            ancestors_deleted_at__isnull=True,
+        )
+
+        shared_with = db.Q(itemaccess__item_id__in=shared_items.values("pk"))
+        shared_by = db.Q(pk__in=shared_items.values("creator_id"))
+        frequency = db.Count("itemaccess", filter=shared_with)
```
NathanVss: Maybe I am missing smth: why is the frequency not taking shared_by into account?

src/backend/core/api/viewsets.py
```diff
@@ -252,6 +252,40 @@ def get_me(self, request):
...
+        contacts = (
```
NathanVss, line 283: Aside: we should maybe be careful with the performance of this request on production. Makes me think that we should maybe think in the near future of some sort of performance testing with a well populated db and make sure on the ci the execution time of the critical query does not explodes after changes

NathanVss: I am thinking of duplicated exec of `shared_with`, as django does not handle CTE

src/frontend/apps/drive/src/features/i18n/translations.json
```diff
@@ -867,6 +881,20 @@
...
+              "powerpoint": "Présentation",
+              "calc": "Tableur",
+              "pdf": "PDF",
+              "image": "Image",
+              "video": "Vidéo",
+              "audio": "Audio",
+              "archive": "Archive",
+              "other": "Autre"
+            }
+          },
```
NathanVss, line 934: Seems you forgot nl translation. ( We should setup a ci job for this imo )

src/frontend/apps/drive/src/features/explorer/components/app-view/ExplorerFilters.tsx
```diff
@@ -54,10 +64,114 @@ export const ExplorerFilters = () => {
...
+      )}
     </div>
   );
 };
 
+const CONTACT_RESET = "__contact_reset__";
+
+type ContactItem = { id: string; label: string; user?: User };
+
+export const ExplorerFilterContact = (props: {
```
NathanVss: Two things about contacts:

- I noticed that a `/contacts` api call is made each time we navigate to a folder ( even when switching tabs ), even without opening the filter. Imo It should only be called when performing a contact search when the filter is opened. The contact call could be a bit costly moreover.

- The contacts route only returns 5 users and does not takes a search parameter, so if we have dozens of contacts, if we search for a contact not into the 5 top most ones, we will never find it, right? 

WDYT about having a contact route that takes a search parameters to apply the filter on the backend side?

src/frontend/apps/drive/src/features/explorer/components/app-view/ExplorerFilters.scss
NathanVss, line 1: Image

I am not sure that the date picker are intended to be displayed in the top bar zone. I do not see this on figma

https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=9171-8091&t=ZxmdKk58OkQE3L8g-0

maybe I missed a conversation between you and @robin-lecomte ?

src/frontend/apps/drive/src/features/explorer/components/app-view/ExplorerFilters.scss
```diff
@@ -2,6 +2,9 @@
   display: flex;
   align-items: center;
   gap: 0.5rem;
+
+  // Match the custom date range picker height with the filter buttons (32px).
+  --c--components--forms-datepicker--height: 32px;
```
NathanVss, line 7: Not sure that this is going to be needed ( depends on my previous comment about the top bar )

src/frontend/apps/drive/src/features/i18n/translations.json
```diff
@@ -907,6 +919,18 @@
             "empty": "Aucun contact trouvé",
             "reset": "Réinitialiser"
           },
+          "modified": {
```
NathanVss, line 941: + nl

src/frontend/apps/e2e/__tests__/app-drive/filters.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    filtersBar.getByRole("button", { name: /^Modified/ }),
+  ).toBeVisible();
+  // The contact filter requires authentication: its contacts request would
+  // get a 401 and redirect the anonymous visitor to the 401 page.
+  await expect(
+    filtersBar.getByRole("button", { name: /^Shared with/ }),
+  ).not.toBeVisible();
+  expect(anonPage.url()).toBe(folderUrl);
+
+  await anonContext.close();
```
NathanVss, line 192: Would be great to make sure those filters work with an anon user too

NathanVss (commented):
Nice improvements, well done 👏

I'm available to talk about those comment if you need 😊

src/backend/core/enums.py
kernicPanel: I removed the `.` in the enum.
About the mime type, I agree we should use it too, but I'm wondering if it's not overkill.

src/backend/demo/management/commands/create_demo.py
kernicPanel: I used it only to match `create_items`.

NathanVss (approved):

PR #713, opened 2026-05-18
🐛(backend) replace VersionId by Etag for WOPI

lunika:
## Purpose

When we try to get the VersionId from the s3 response but this one is not send by the s3 storage (garage for example), we get a 500 error.
To fix this issue we use the ETag sent by the object storage to track the version id. This ETag is present also when versionning is not implemented.

## Proposal

- [x] 🐛(backend) replace VersionId by Etag for WOPI


src/backend/wopi/utils/__init__.py
```diff
@@ -88,3 +88,18 @@ def compute_wopi_launch_url(launch_url, get_file_info_path, lang=None):
...
+    """Build a stable WOPI item version token from storage metadata."""
+    etag = head_object.get("ETag")
+    if etag:
+        return etag.strip('"')
+
+    last_modified = head_object.get("LastModified")
+    if last_modified:
+        if hasattr(last_modified, "isoformat"):
+            return last_modified.isoformat()
+        return str(last_modified)
```
kernicPanel:
```suggestion
    if etag := head_object.get("ETag") :
        return etag.strip('"')

    if last_modified := head_object.get("LastModified"):
        if hasattr(last_modified, "isoformat"):
            return last_modified.isoformat()
            
        return str(last_modified)
```

kernicPanel (approved):

PR #716, opened 2026-05-22
✨ Add user account reconciliation

kernicPanel:
## Purpose

Port of suitenumerique/docs#1878 (with its follow-up fix 3399734a). Kept as
close as possible to docs to allow future factorization into django-lasuite.

For now, reconciliation requests are imported through CSV in the Django admin,
which sends confirmation emails to both addresses. When both are checked, the
actual reconciliation is processed, and Drive user-related references are
updated. The inactive account is deactivated, not deleted.

## Proposal

- New UserReconciliationCsvImport model to manage the import of reconciliation requests through a task (user_reconciliation_csv_import_job)
- New UserReconciliation model to store the reconciliation requests themselves (a row = an active_user/inactive_user pair)
  - On save, confirmation emails are sent to the users
- A process_reconciliation admin action processes the requested entries, if both emails have been checked
  - Bulk update ItemAccess, keeping the higher role when both users have access
  - Bulk update LinkTrace
  - Bulk update ItemFavorite
  - Bulk update Item.creator
  - Bulk update Invitation.issuer
  - Bulk update the is_active status on both users
- New USER_RECONCILIATION_FORM_URL env variable for the "make a new request" URL in an email
- New send_email() method on User, similar to the one on the Item model
- Public frontend confirmation pages for email validation, including the nginx rule required by Drive static export
- Dev: create_reconciliation_demo management command (make reconciliation-demo) to exercise the flow
- Unit and e2e tests

## Drive-specific vs docs

- No comment system in Drive: drops the Thread/Comment/Reaction transfers
- Also reassigns Item.creator and Invitation.issuer
- Invalidates the nb_accesses cache on bulk-deleted accesses


src/backend/core/models.py
```diff
@@ -324,6 +324,41 @@ def email_user(self, subject, message, from_email=None, **kwargs):
             raise ValueError("User has no email address.")
         mail.send_mail(subject, message, from_email, [self.email], **kwargs)
 
+    def send_email(self, subject, context=None, language=None):
+        """Generate and send email to the user from a template."""
+
+        if not settings.EMAIL_HOST:
+            logger.debug("EMAIL_HOST host is not set, skipping email sending")
```
lunika: IMO debug is not appropriate. I would use at least `info`

src/backend/core/migrations/0022_alter_item_numchild_and_more.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 5.2.14 on 2026-05-21 12:03
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
lunika, line 6: Should be removed.

src/backend/core/migrations/0023_userreconciliationcsvimport_userreconciliation.py
```diff
@@ -0,0 +1,57 @@
...
+import django.db.models.deletion
+import uuid
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0022_alter_item_numchild_and_more'),
```
lunika: This migration does not exists anymore.

src/backend/demo/management/commands/create_reconciliation_demo.py
```diff
@@ -0,0 +1,183 @@
...
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from core import factories, models
+
+DEFAULT_OUTPUT = "demo/reconciliation_demo.csv"
+# Active user = the demo dev user, which exists in the Keycloak realm so it can
+# be used to log in on the frontend and check the merged content.
+ACTIVE_EMAIL = "drive@drive.world"
```
lunika, line 15: This user is already present in the keycloak configuration file and can be used for test purpose. When using this command it can lead to an error saying that the user with this sub already exists. You can create another user or catch this error to ignore it.

lunika (approved):

PR #719, opened 2026-05-27
Wopi/background conversion poc

kernicPanel:
## Purpose

  Convert `.doc` / `.xls` / `.ppt` to their modern equivalents in the
  background, without opening an editor. Double-clicking a legacy file
  shows a confirmation modal; on confirm, a placeholder item appears as
  "converting" in the listing and is replaced by the converted file when
  ready.

  ## Proposal

  Backend
  
  - `POST /api/v1.0/items/{id}/convert/` — creates a placeholder Item in
    the new `CONVERTING` upload state at the destination, queues
    `convert_file` on Celery, returns the placeholder `id`. Cleans up and
    returns `503` on broker failure.
  - `wopi/conversion/` — policy (`doc → docx`, `xls → xlsx`, `ppt → pptx`),
    `OnlyOfficeConversionBackend` driving `/converter` with separate
    connect/read timeouts and optional HS256 JWT.
  - Short-lived WOPI source URL signed with the existing
    `AccessUserItemService` (`WOPI_CONVERSION_SOURCE_TOKEN_TIMEOUT`,
    default `120s`).
  - New `convert` ability on `Item`, aligned with `update`.
  - Case-insensitive WOPI extension lookup.

  Frontend
  
  - Confirmation modal on double-click of a legacy file.
  - Placeholder rendered as a transient "converting" row.
  - Duplicating poller generalised to transient items.
  - API client for the convert endpoint.


src/backend/wopi/conversion/policy.py
```diff
@@ -0,0 +1,42 @@
...
+
+    The decision is product policy, not discovery-derived: a client is allowed
+    to advertise an "edit" action for a legacy format and still require the
+    Drive backend to convert it first.
+    """
+    if not client_options:
+        return False
+
+    extension = _normalize(item.extension)
+    forced_extensions = {_normalize(e) for e in client_options.get("ForceConvertExtensions") or []}
```
lunika, line 33: The `ForceConvertExtensions` is created in the next commits, it is confusing to use code you will create in a future commit.

src/backend/wopi/tests/conversion/test_backends_onlyoffice.py
```diff
@@ -0,0 +1,189 @@
+"""Tests for the OnlyOffice server-to-server conversion backend."""
```
lunika, line 1: I think it should be in a `conversion` module `wopi/tests/conversion/test_backends_onlyoffice.py`

src/backend/wopi/tests/test_conversion_backends_onlyoffice.py
```diff
@@ -0,0 +1,189 @@
...
+
+CONVERT_URL = "https://office.example/converter"
+FILE_URL = "https://office.example/result/converted.docx"
+SOURCE_URL = "https://drive.example/source"
+JWT_SECRET = "test-secret-with-enough-length-for-hs256"
+
+
+def _item(filename="document.doc", extension="doc"):
+    """Build a minimal Item-like namespace for backend tests."""
+    return SimpleNamespace(id=uuid4(), filename=filename, extension=extension)
```
lunika: Why using this instead of a real factory ? What will happen if tomorrow the `filename` property changes in the Item model?

src/backend/wopi/tests/test_conversion_source_url.py
```diff
@@ -0,0 +1,65 @@
...
+    """Start the built URL with WOPI_SRC_BASE_URL and embed the access token."""
+    settings.WOPI_SRC_BASE_URL = "https://drive.example"
+    item = SimpleNamespace(id=uuid4())
+    user = SimpleNamespace()
+
+    url = build_source_url(item, user)
+
+    assert url.startswith("https://drive.example/api/v1.0/wopi/files/")
+    assert f"files/{item.id}/contents/" in url
+    assert "access_token=tok-abc" in url
```
lunika: why not doing only one assert with what is exactly expected ? some other thing can be present and you will not detect it probably

src/backend/wopi/conversion/services.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+
+def _resolve_destination_parent(source_item, user):
+    """Return the source parent when the user can still update it, else None."""
+    if source_item.depth <= 1:
+        return None
+
+    parent = source_item.parent()
+    if not parent.get_abilities(user).get("update"):
```
lunika:
```suggestion
    if not parent.get_abilities(user).get("children_create"):
```

I think the good ability to check is `children_create`

src/backend/wopi/conversion/services.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    converted_file.seek(0)
+
+    default_storage.save(placeholder.file_key, converted_file)
+    try:
+        with transaction.atomic():
+            placeholder.mimetype = mimetype
+            placeholder.size = converted_file.size
+            placeholder.upload_state = models.ItemUploadStateChoices.READY
+            placeholder.save(update_fields=["mimetype", "size", "upload_state", "updated_at"])
```
lunika: Don't understand the usage of a transaction here. Only one operation is made, so it fails there is nothing to rollback.

src/backend/wopi/conversion/services.py
```diff
@@ -0,0 +1,160 @@
...
+    except Exception:
```
lunika: Not sure pylint allows you to except for an `Exception`, you are more looking for a `DatabaseError`

src/backend/core/models.py
```diff
@@ -1318,6 +1325,7 @@ def get_abilities(self, user):
             "update": can_update,
             "upload_ended": can_update and user.is_authenticated,
             "wopi": can_get,
+            "must_convert": must_convert,
```
lunika: The viewsets action name is `convert`, why not calling it `convert` ? The name of an ability should map an action name.

src/backend/core/api/permissions.py
```diff
@@ -11,6 +11,7 @@
 ACTION_FOR_METHOD_TO_PERMISSION = {
     "versions_detail": {"DELETE": "versions_destroy", "GET": "versions_retrieve"},
     "children": {"GET": "children_list", "POST": "children_create"},
+    "convert": {"POST": "update"},
```
lunika: Don't understand this mapping. You created in the previous PR an ability `must_update` that is checking must more than only the `update` ability. And this ability `must_convert` seems used only in the tests.

src/backend/core/api/viewsets.py
```diff
@@ -644,6 +649,41 @@ def hard_delete(self, request, *args, **kwargs):
...
+                converted_item_id=str(placeholder.id),
+                user_id=str(request.user.id),
+            )
+        except (CeleryError, KombuError):
+            placeholder.soft_delete()
+            placeholder.delete()
+            return drf.response.Response(
+                {"detail": "Conversion could not be queued."},
+                status=status.HTTP_503_SERVICE_UNAVAILABLE,
+            )
```
lunika: IMO nothing should be catch. A 500 must be raised to warn us. Here it is silently failing without any feedback.

src/backend/wopi/tasks/conversion.py
```diff
@@ -0,0 +1,61 @@
...
+        placeholder.delete()
+        return
+
+    try:
+        perform_conversion(source, placeholder, user)
+    except Exception:
+        logger.exception("convert_file: conversion failed for placeholder %s", placeholder.id)
+        placeholder.soft_delete()
+        placeholder.delete()
+        raise
```
lunika, line 78: This is a little bit abrupt. With a retry strategy, you can try the conversion multiple times before deleting the item.

src/backend/wopi/tests/tasks/test_conversion.py
```diff
@@ -0,0 +1,121 @@
...
+
+    monkeypatch.setattr(conversion, "perform_conversion", fake_perform)
+
+    conversion.convert_file(
+        source_item_id=str(source.id),
+        converted_item_id=str(placeholder.id),
+        user_id=str(user.id),
+    )
+
+    assert calls["args"] == (source.id, placeholder.id, user.id)
```
lunika: IMO you should use mock.path instead of monkeypath. The mock will allow you to change the implementation and have built assertion to check how the patch has been called.

src/backend/wopi/apps.py
```diff
@@ -7,3 +7,8 @@ class WopiConfig(AppConfig):
     """Configuration class for the wopi app."""
 
     name = "wopi"
+
+    def ready(self):
+        """Import celery tasks."""
+        # pylint: disable=import-outside-toplevel, unused-import
+        from .tasks import conversion  # noqa: PLC0415,F401
```
lunika: With what is made in the `drive.celery_app` module, this code should be useless.

lunika (commented):
As a global comment, I think you can replace all the usage of `mokeypatch` by mock.patch.

src/backend/wopi/tasks/conversion.py
kernicPanel: This as been reworked.

src/frontend/apps/drive/src/features/drivers/types.ts
```diff
@@ -93,6 +94,7 @@ export type Item = {
     tree: boolean;
     update: boolean;
     upload_ended: boolean;
+    convert?: boolean;
```
PanchoutNathan, line 102: Does that mean it's a converted file? If so, maybe `isConverted` or something like that would be better, WDYT?

src/frontend/apps/drive/src/features/explorer/components/modals/ConvertLegacyFileModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (isSubmitting) return;
+    setError(null);
+    onClose();
+  };
+
+  const handleConvert = async () => {
+    setIsSubmitting(true);
+    setError(null);
+    try {
+      await getDriver().convertItem(item.id);
```
PanchoutNathan: I don’t think this modal should call `getDriver().convertItem()` directly. Conversion is a server mutation with cache side effects, so it should go through a dedicated hook such as `useMutationConvertItem`, consistently with the other item actions. That would centralize `isPending`, error handling, and especially React Query refresh/invalidation in one place instead of delegating it to the parent through `onSuccess`. Since we will want to refresh lists after every conversion, that behavior should live in the mutation rather than in each calling component. Feel free to look at other mutation hooks; they're pretty much everywhere.

src/frontend/apps/drive/src/features/explorer/components/app-view/AppExplorerGrid.tsx
```diff
@@ -171,6 +183,14 @@ export const AppExplorerGrid = () => {
           <Spinner size="xl" />
         </div>
       )}
+      {convertModal.isOpen && itemToConvert && (
+        <ConvertLegacyFileModal
+          item={itemToConvert}
+          isOpen={convertModal.isOpen}
+          onClose={convertModal.onClose}
+          onSuccess={() => refresh(item?.originalId ?? item?.id)}
```
PanchoutNathan: cf ConvertLegacyFileModal comment

src/frontend/apps/drive/src/features/explorer/components/app-view/AppExplorerGrid.tsx
```diff
@@ -45,9 +48,18 @@ export const AppExplorerGrid = () => {
 
   const effectiveOnNavigate = appExplorer.onNavigate ?? onNavigate;
 
+  const convertModal = useModal();
+  const [itemToConvert, setItemToConvert] = useState<Item | null>(null);
+  const refresh = useRefreshQueryCacheAfterMutation();
```
PanchoutNathan: cf ConvertLegacyFileModal comment

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGridMobileCell.tsx
```diff
@@ -15,11 +19,15 @@ const EmbeddedExplorerGridMobileCellComponent = (
 ) => {
   const item = params.row.original;
   const { t } = useTranslation();
-  const isDuplicating = item.upload_state === ItemUploadState.DUPLICATING;
+  const isConverting = item.upload_state === ItemUploadState.CONVERTING;
+  const isTransient = TRANSIENT_UPLOAD_STATES.includes(item.upload_state);
+  const transientLabel = isConverting
+    ? t("explorer.item.converting")
+    : t("explorer.item.duplicating");
```
PanchoutNathan: Now that there are several cases, I would do it like a man, take an item as a parameter and it will return what you need. This way, if we add a case, it will be centralized and we can reuse it in all components.

src/frontend/apps/drive/src/features/explorer/components/embedded-explorer/EmbeddedExplorerGridNameCell.tsx
```diff
@@ -24,7 +29,11 @@ const EmbeddedExplorerGridNameCellComponent = (
   const [isOverflown, setIsOverflown] = useState(false);
   const { disableItemDragAndDrop } = useEmbeddedExplorerGirdContext();
   const isSelected = useIsItemSelected(item.id);
-  const isDuplicating = item.upload_state === ItemUploadState.DUPLICATING;
+  const isConverting = item.upload_state === ItemUploadState.CONVERTING;
+  const isTransient = TRANSIENT_UPLOAD_STATES.includes(item.upload_state);
+  const transientLabel = isConverting
+    ? t("explorer.item.converting")
+    : t("explorer.item.duplicating");
```
PanchoutNathan: ditto  : https://github.com/suitenumerique/drive/pull/719/changes/100b3a5a7c0665cbde64448bebad5362ed66e919#r3347528062

src/frontend/apps/drive/src/features/drivers/types.ts
kernicPanel: It means that the file is convertible and should be converted before edition.
Previously it was `must_convert` which is more explicit, but obviously defeats the `ability` purpose.
With lunika, we choose to match with the other actions names.

PanchoutNathan (approved):

lunika (approved):

PR #728, opened 2026-06-05
Fix OnlyOffice conversion JWT signing

kernicPanel:
## Purpose

OnlyOffice rejects unsigned `/converter` requests with error code -8
when JWT is enabled. The conversion feature (PR #719) supported JWT
signing but was never wired to a secret, and the token format did not
match the OnlyOffice API specification.

## Proposal

- Remove dead `ConvertJwtSecret` / `ConvertJwtRequired` options plumbing
- Add a dedicated `WOPI_ONLYOFFICE_CONVERT_JWT_SECRET` Django setting
- Fix JWT body token format: encode parameters directly instead of
  wrapping in `{"payload": …}`, send only `{"token": "…"}` in the body
- Prefer the WOPI `access_token` query param over the `Authorization`
  header so OnlyOffice's own JWT header does not shadow the WOPI token
- Reject conversion early (before creating the placeholder) when the
  secret is missing, and hide the `convert` ability from the API
- Enable JWT on the development OnlyOffice container
- Document the new conversion environment variables

### DIFF
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -1420,6 +1420,9 @@ class Base(Configuration):
     WOPI_ONLYOFFICE_CONVERT_DOWNLOAD_READ_TIMEOUT = values.IntegerValue(
         30, environ_name="WOPI_ONLYOFFICE_CONVERT_DOWNLOAD_READ_TIMEOUT", environ_prefix=None
     )
+    WOPI_ONLYOFFICE_CONVERT_JWT_SECRET = SecretFileValue(
+        None, environ_name="WOPI_ONLYOFFICE_CONVERT_JWT_SECRET", environ_prefix=None
+    )
     WOPI_DISABLE_CHAT = values.IntegerValue(
         0, environ_name="WOPI_DISABLE_CHAT", environ_prefix=None
     )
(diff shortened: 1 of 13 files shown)


src/backend/wopi/tests/conversion/test_backends_onlyoffice.py
```diff
@@ -59,20 +59,24 @@ def test_convert_sends_synchronous_payload_with_shardkey():
...
-    assert decoded["payload"] == payload
-    assert request.headers.get("Authorization") == f"Bearer {token}"
+    body = json.loads(request.body)
+    assert list(body.keys()) == ["token"]
+
+    decoded = jwt.decode(body["token"], JWT_SECRET, algorithms=["HS256"])
+    assert decoded["filetype"] == "doc"
+    assert decoded["outputtype"] == "docx"
+    assert decoded["url"] == SOURCE_URL
+    assert decoded["title"] == item.filename
```
lunika: Why not asserting `decoded` at once with 

````
assert decoded == { .. }
```

This way you test that there is no extra keys not tested.

src/backend/drive/settings.py
```diff
@@ -1420,6 +1420,9 @@ class Base(Configuration):
     WOPI_ONLYOFFICE_CONVERT_DOWNLOAD_READ_TIMEOUT = values.IntegerValue(
         30, environ_name="WOPI_ONLYOFFICE_CONVERT_DOWNLOAD_READ_TIMEOUT", environ_prefix=None
     )
+    WOPI_ONLYOFFICE_CONVERT_JWT_SECRET = values.Value(
```
lunika: you should use `SecretFileValue` instead of `values.Value`

lunika (approved):

PR #738, opened 2026-06-11
🐛(backend) stream export files from S3 without buffering

kernicPanel:
## Purpose

Exporting a folder holding large files can exhaust the worker memory:
`default_storage.open()` downloads each whole object in memory before
the first read.

## Proposal

- Stream files straight from the boto3 response body
- Raise the read chunk size from 1 KB to 64 KB

### DIFF
--- a/src/backend/core/services/item_exports.py
+++ b/src/backend/core/services/item_exports.py
@@ -1,19 +1,33 @@
 """Service for exporting item folders as streaming ZIP archives."""
 
+import logging
+
 from django.core.files.storage import default_storage
 
 from zipstream import ZipStream
 
 from core import models
 
-DEFAULT_STORAGE_READ_CHUNK_SIZE = 1024
+logger = logging.getLogger(__name__)
+
+DEFAULT_STORAGE_READ_CHUNK_SIZE = 64 * 1024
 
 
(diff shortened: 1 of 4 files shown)


src/backend/core/services/item_exports.py
```diff
@@ -6,14 +6,17 @@
...
 def iter_storage_chunks(file_key, chunk_size=DEFAULT_STORAGE_READ_CHUNK_SIZE):
     """Yield bytes from object storage without buffering the whole file."""
-    with default_storage.open(file_key, "rb") as fh:
-        while chunk := fh.read(chunk_size):
-            yield chunk
+    # default_storage.open() would download the whole object in memory before
+    # the first read, so stream straight from the boto3 response body instead.
+    s3_client = default_storage.connection.meta.client
+    bucket_name = default_storage.bucket_name
+    response = s3_client.get_object(Bucket=bucket_name, Key=file_key)
```
lunika: What should happen if the `get_object` fails ? Should we return the zip without this file or should we abort the export action ?

kernicPanel: You're right, the whole export shouldn't fail.
I added an except.

lunika (approved):

PR #740, opened 2026-06-12
✨(helm) add feature environment for review apps

kernicPanel:
## Purpose

Allow ArgoCD to deploy an autonomous review app per feature
branch, following the docs repository model. The dev-backend
values move to per-environment files; the dev rendering is
unchanged. WOPI points to an OnlyOffice instance shared by all
review apps, to be deployed separately.

### DIFF
--- a/src/helm/env.d/dev/values.dev-backend.yaml.gotmpl
+++ b/src/helm/env.d/dev/values.dev-backend.yaml.gotmpl
@@ -1,0 +1,105 @@
+postgres:
+  enabled: true
+  name: postgres
+  #serviceNameOverride: postgres
+  image: postgres:16-alpine
+  username: dinum
+  password: pass
+  database: dinum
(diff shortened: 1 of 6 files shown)


src/helm/env.d/dev/values.dev-backend.yaml.gotmpl
```diff
@@ -0,0 +1,105 @@
...
+    password: pass
+    database: keycloak
+    size: 1Gi
+    image: postgres:16-alpine
+  realm:
+    name: drive
+    username: drive
+    password: drive
+    email: drive@example.com
+dsproxy:
```
lunika: Can be removed

lunika:
> WOPI points to an OnlyOffice instance shared by all review apps, to be deployed separately.

So the settings will be overridden when deployed?

PR #743, opened 2026-06-16
✨(backend) accept configurable media-auth forward headers

mosa-riel:
## Summary

`ItemViewSet._authorize_subrequest()` only read `HTTP_X_ORIGINAL_URL`, an nginx-ingress convention. Behind a proxy that carries the original URL in a different header (for instance `X-Forwarded-Uri`) the lookup returns `None` and every `/media/...` subrequest is denied with a 403 — uploads succeed but downloads/previews never resolve.

This reads the original URL from the **first present** header in a new, configurable `MEDIA_AUTH_FORWARD_HEADERS` setting (default `["X-Original-Url", "X-Forwarded-Uri"]`, tried in order). Existing deployments are unaffected because `X-Original-Url` stays first; other proxies can be supported purely by overriding the env var, with no code change.

## Changes

- `core/api/viewsets.py` — iterate the configured headers, convert each to its `META` key (`HTTP_` prefix, uppercased, dashes→underscores) and use the first one present.
- `drive/settings.py` — new `MEDIA_AUTH_FORWARD_HEADERS` list setting (env `MEDIA_AUTH_FORWARD_HEADERS`).
- `docs/env.md` — document the new variable.
- `core/tests/items/test_api_items_media_auth.py` — mirror tests exercising `X-Forwarded-Uri` for the 200 (public) and 403 (authenticated/restricted) paths, plus a no-header 403 case.
- `CHANGELOG.md` — `Added` entry.

## Test plan

- [x] Existing `test_api_items_media_auth.py` suite still green (`X-Original-Url` path unchanged).
- [x] New `X-Forwarded-Uri` cases reach the same authorization decisions (200/403).
- [x] No-header subrequest still denied (403).
- [x] `ruff format --check .`, `ruff check .` clean; `pylint` 10.00/10 on the touched modules.
- [ ] Manual: behind a ForwardAuth-style middleware, `GET /media/<key>` returns 200 for an authorized user (previously 403).

`22 passed` locally (`DJANGO_CONFIGURATION=Test pytest core/tests/items/test_api_items_media_auth.py`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(backend) accept configurable media-auth forward headers
+
 ## [v0.21.0] - 2026-08-07
 
 ### Added
--- a/src/backend/drive/settings.py
+++ b/src/backend/drive/settings.py
@@ -143,6 +143,15 @@ class Base(Configuration):
     MEDIA_ROOT = os.path.join(DATA_DIR, "media")
     MEDIA_BASE_URL = values.Value(None, environ_name="MEDIA_BASE_URL", environ_prefix=None)
 
+    # Request headers carrying the original URL of a media auth subrequest, tried in order.
+    # nginx ingress uses "X-Original-Url"; Traefik's ForwardAuth uses "X-Forwarded-Uri".
+    # Add other proxy header names here (or via the env var) as needed.
+    MEDIA_AUTH_FORWARD_HEADERS = values.ListValue(
+        ["X-Original-Url", "X-Forwarded-Uri"],
+        environ_name="MEDIA_AUTH_FORWARD_HEADERS",
+        environ_prefix=None,
+    )
+
     SITE_ID = 1
 
     STORAGES = {
--- a/docs/env.md
+++ b/docs/env.md
@@ -72,6 +72,7 @@ This document lists all configurable environment variables for the Drive applica
 | `LOGGING_LEVEL_LOGGERS_APP` | Logging level for application loggers | `INFO` |
 | `LOGGING_LEVEL_LOGGERS_ROOT` | Logging level for root logger | `INFO` |
 | `MAX_PAGE_SIZE` | Limit the maximum page size the client may request | `200` |
+| `MEDIA_AUTH_FORWARD_HEADERS` | Headers carrying the original URL of a media auth subrequest, tried in order | `["X-Original-Url", "X-Forwarded-Uri"]` |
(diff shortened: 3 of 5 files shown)


src/backend/drive/settings.py
```diff
@@ -143,6 +143,15 @@ class Base(Configuration):
     MEDIA_ROOT = os.path.join(DATA_DIR, "media")
     MEDIA_BASE_URL = values.Value(None, environ_name="MEDIA_BASE_URL", environ_prefix=None)
 
+    # Request headers carrying the original URL of a media auth subrequest, tried in order.
+    # nginx ingress uses "X-Original-Url"; Traefik's ForwardAuth uses "X-Forwarded-Uri".
+    # Add other proxy header names here (or via the env var) as needed.
+    MEDIA_AUTH_FORWARD_HEADERS = values.ListValue(
```
kernicPanel, line 149: Please also document MEDIA_AUTH_FORWARD_HEADERS in docs/env.md.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(backend) accept configurable media-auth forward headers (Traefik X-Forwarded-Uri)
```
kernicPanel: Drop the "(Traefik X-Forwarded-Uri)" suffix.

kernicPanel (commented):
Sorry for the late reply.
The change looks good and safe.

Since the mechanism is now generic, please remove the Traefik/nginx mentions from the docstring and the comments in `viewset.py`.

This is a new feature rather than a fix, so the commit should be ✨ like the changelog entry.

The branch also conflicts with main, please rebase.

mosa-riel:
Thanks for the review! Addressed:

- **Generic wording** — dropped the nginx/Traefik mentions from the `_authorize_subrequest` docstring and the inline comment in `viewsets.py`; they now talk about "a reverse proxy auth subrequest" and point at `MEDIA_AUTH_FORWARD_HEADERS` for the header list.
- **✨ instead of 🐛** — commit retitled (and the PR title too) to match the changelog entry.
- **Changelog** — dropped the `(Traefik X-Forwarded-Uri)` suffix.
- **docs/env.md** — `MEDIA_AUTH_FORWARD_HEADERS` documented in the table.
- **Rebase** — rebased onto current `main`, conflict was just the changelog `[Unreleased]` section. No longer conflicting.

`22 passed` on `core/tests/items/test_api_items_media_auth.py`, `ruff format --check .` / `ruff check .` clean, `pylint` 10.00/10 on the touched modules.

kernicPanel (approved):

PR #748, opened 2026-06-19
Keep files accessible and convertible during malware analysis

kernicPanel:
## Purpose

After upload, a file is scanned for malware in the background (`analyzing`
state). It belongs to the user who just uploaded it and must stay fully
usable meanwhile. Instead, such items were locked in the UI, and converting
a legacy Office file was only offered once the analysis had finished.

## Proposal

- Keep `analyzing` items accessible: they are polled to refresh their
  abilities but no longer treated as transient/locked; the "analysis in
  progress" text becomes a non-blocking badge.
- Allow converting a file while it is still being analyzed: the convert
  ability and the conversion request are accepted during `analyzing`, but
  the real conversion is deferred until analysis confirms the source is safe
  (`ready`), and aborts if the file turns suspicious.
- Dev: the dummy malware backend now delays its result so the `analyzing`
  state is observable locally.

### DIFF
--- a/src/backend/wopi/tasks/conversion.py
+++ b/src/backend/wopi/tasks/conversion.py
@@ -12,6 +12,12 @@ from drive.celery_app import app
 
 logger = logging.getLogger(__name__)
 
+# Seconds to wait before re-checking a source still under malware analysis,
+# and how many times to wait before giving up (~10 minutes total, matching the
+# frontend polling timeout).
+ANALYSIS_RETRY_COUNTDOWN = 30
+ANALYSIS_MAX_RETRIES = 20
+
 
 class ConvertFileTask(app.Task):
     """Celery task base deleting the placeholder when conversion ultimately fails."""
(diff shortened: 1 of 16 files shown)


src/backend/wopi/tasks/conversion.py
```diff
@@ -71,6 +77,13 @@ def convert_file(source_item_id, converted_item_id, user_id):
         logger.error("convert_file: user %s does not exist, aborting", user_id)
         return
 
+    if source.upload_state == models.ItemUploadStateChoices.ANALYZING:
```
NathanVss, line 81: So it means that if the analysis takes more than 5min the conversion will never occur? It seems a bit short nope? WDYT?

NathanVss (approved):
LGTM ! 

Two minor points:
- Could you add in the convert modal a specific text if the analyzing is pending that the conversion will occur after
- Check with @robin-lecomte to see if we should keep this (analyse en cours) state. not sure imo.

Image

robin-lecomte:
I would remove the "(analyse en cours)" since it doesn’t seem to add any value, as the user can still open it anyway.

src/backend/wopi/tasks/conversion.py
kernicPanel: I raised the retries to 20, which gives 10 minutes total.

PR #756, opened 2026-07-01
Restricted access

### LINKED ISSUE #731: Restricted folders
Image

As part of prioritizing features related to the Resana migration, the restricted-access folders feature, is now a priority.

The restricted-access folder allows the user to break a folder’s incoming inheritance to enforce a more restrictive configuration than its parent. Inherited roles no longer grant any access, regardless of their level, including Owner.  

When deactivated, the folder reverts to the normal inheritance model: any explicit value lower than or equal to the inherited value is removed, while a higher explicit value is retained.

[SPECS](https://docs.numerique.gouv.fr/docs/8cfe8d2e-83b7-4750-8a13-8b08fa26d466/) | [FIGMA](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=15597-7601&t=3WaMvjiVXPAVUs9P-1)

kernicPanel:
## Purpose

Closes #731.

Let a folder owner restrict a folder so that only its explicit members
keep access, cutting the roles and link reach inherited from the tree.

This replaces the previous approach of this PR (cutting inheritance with
query-level exclusions) following review: the boundary had to be
reimplemented in every access-related query, and any forgotten spot
leaked content — link reach was already missing from the visibility
filters. The branch was rebuilt from main.

## Proposal

A restricted folder physically becomes a root of the tree, so
inheritance stops by construction and every existing and future query
is correct without knowing about restriction. A new `shortcut` item
type materializes its original location.

- [x] `is_restricted` field and `shortcut` item type targeting a restricted root
- [x] `restrict` ability: explicit owners only, activation needs a parent, deactivation always possible
- [x] activation: explicit owner access, move to the root, shortcut left in place, explicit link reach kept (defaulted to restricted when inherited)
- [x] deactivation: reattach at the shortcut's current parent, then normalize explicit accesses and link reach against restored inheritance
- [x] API: `PATCH {"is_restricted": bool}`, `target` payload on shortcuts (`can_access` drives the greyed entry), shortcuts in the tree endpoint
- [x] top-level listing hides a restricted root from users who already reach its live shortcut
- [x] deleting a shortcut detaches the folder (the excluded owner acts on the container only); trashing an ancestor or the restricted folder itself detaches too
- [x] shortcuts excluded from search, export and indexing
- [x] a folder can be created restricted (parent owners only)
- [x] permissions refactor: role resolution and abilities behind a swappable backend
- [x] fix: `parent()` resolved by exact path after a move

Frontend follow-up needed: render the `shortcut` type (open `target.id`
when `can_access`, greyed otherwise) and the activation/deactivation
warnings.


src/backend/core/api/serializers.py
```diff
@@ -691,6 +707,12 @@ def validate(self, attrs):
                 code="item_create_folder_title_required",
             )
 
+        if attrs.get("is_restricted") and attrs["type"] != models.ItemTypeChoices.FOLDER:
+            raise serializers.ValidationError(
```
sampaccoud, line 746: I think this constraint should be defined on the model directly

src/backend/core/permissions/backends/role.py
```diff
@@ -0,0 +1,27 @@
...
+        )
+
+    def roles_at(self, user, path):
+        """Return the roles the user holds at the given path, direct or inherited."""
+        return models.ItemAccess.objects.filter(
+            Q(user=user) | Q(team__in=user.teams),
+            item__path__ancestors=path,
+        ).values_list("role", flat=True)
+
+    def roles_for(self, user, item):
```
NathanVss: Could be put in the parent class?

src/backend/core/models.py
```diff
@@ -1199,10 +1198,7 @@ def get_role(self, user):
         try:
             roles = self.user_roles or []
         except AttributeError:
-            roles = ItemAccess.objects.filter(
-                models.Q(user=user) | models.Q(team__in=user.teams),
-                item__path__ancestors=self.path,
-            ).values_list("role", flat=True)
+            roles = get_permissions_backend().roles_for(user, self)
 
         return RoleChoices.max(*roles)
```
NathanVss, line 1277: `role_at` would work here?

src/backend/core/permissions/backends/role.py
```diff
@@ -1,27 +1,48 @@
...
 
     def effective_accesses(self, item):
-        """Return the accesses applying to the item, direct or inherited."""
+        """Return the accesses applying to the item, down to its restriction boundary."""
         return models.ItemAccess.objects.filter(
             item__path__ancestors=item.path,
-        )
+        ).exclude(_cut_by_restriction(item.path))
 
     def roles_at(self, user, path):
```
NathanVss: Aren't `roles_at` and `roles_for` redundant in some way?

src/backend/core/tests/test_models_items_restricted.py
```diff
@@ -113,3 +114,111 @@ def test_models_items_restricted_get_role_descendant_inherits_from_restricted_fo
...
+    child.is_restricted = True
+    child._computed_link_definition = None  # pylint: disable=protected-access
+
+    assert child.computed_link_definition == {
+        "link_reach": LinkReachChoices.RESTRICTED,
+        "link_role": "reader",
+    }
+
+
+def test_models_items_restricted_computed_link_definition_uses_explicit_value():
```
NathanVss: isn't redundant with `test_models_items_restricted_computed_link_definition_ignores_ancestors`?

src/backend/core/models.py
```diff
@@ -1020,11 +1021,20 @@ class Meta:
...
+            ),
         ]
         indexes = [
             GistIndex(fields=["path"]),
             models.Index(NLevel(models.F("path")), name="drive_item_path_nlevel_idx"),
+            GistIndex(
+                fields=["path"],
+                name="drive_item_restricted_path_ix",
+                condition=models.Q(is_restricted=True),
+            ),
```
sampaccoud: The "why" of this is not announced in the commit and no tests confirm the sought behavior.

src/backend/core/api/viewsets.py
```diff
@@ -1124,11 +1124,15 @@ def children(self, request, *args, **kwargs):
         # Apply ordering only now that everything is filtered and annotated
         queryset = ItemOrdering().filter_queryset(self.request, queryset, self)
 
-        # Pre-compute number of accesses
+        # Pre-compute number of accesses; the parent's count does not apply
+        # to restricted children which cut inheritance
         item_nb_accesses = item.nb_accesses
```
sampaccoud, line 1215: I think this should be renamed to `nb_effective_accesses`.
The current name is a bit confusing. I had to reread the `nb_accesses` method on the model to remember what past me wrote. 😅

src/backend/core/permissions/backends/role.py
```diff
@@ -1,27 +1,48 @@
...
+        )
 
     def roles_for(self, user, item):
         """Return the roles the user holds on the item, direct or inherited."""
+        if item.is_restricted:
+            # Inheritance is cut on the item itself: only direct accesses apply
+            return models.ItemAccess.objects.filter(
+                Q(user=user) | Q(team__in=user.teams),
+                item=item,
+            ).values_list("role", flat=True)
```
sampaccoud: smells a bit like AI slope.

Here you are shortcircuiting your own `roles_at` method instead of passing it an argument "should_inherit=True" and let it do the job.
And above `_cut_by_restriction` is worse because you went down to the path from the item and go back to check `is_restricted` on the item via a query 😲 

I think you should trash the `_cut_by_restriction` method and modify each existing filter clause directly looking at the `should_inherit` argument.

src/backend/core/permissions/backends/base.py
```diff
@@ -29,6 +29,14 @@ def roles_at(self, user: models.User | AnonymousUser, path: str) -> QuerySet[str
...
         """Return the roles the user holds on the item, direct or inherited."""
 
+    @abstractmethod
+    def ancestors_links_paths_mapping(self, item: models.Item) -> dict[str, list[dict]]:
+        """Return the link definitions applying to each ancestor path of the item."""
+
+    @abstractmethod
+    def link_definition_for(self, item: models.Item) -> dict:
+        """Return the effective link definition of the item, own and inherited combined."""
+
```
sampaccoud, line 31: It would have been nice to move it in the first commit creating the backend since they were already existing.

src/backend/core/models.py
```diff
@@ -1262,9 +1246,7 @@ def computed_link_definition(self):
         definitions and the document's own link definition.
         """
         if getattr(self, "_computed_link_definition", None) is None:
-            self._computed_link_definition = get_equivalent_link_definition(
-                [self.ancestors_link_definition, self.link_definition]
-            )
+            self._computed_link_definition = get_permissions_backend().link_definition_for(self)
```
sampaccoud: I liked that the `get` informed us about the fact that this requires computation
I liked that the `equivalent` informed is about the inheritance vs the direct link definition of an item. Maybe reword it to `effective` to follow your logic on accesses? But removing it from the name makes it harder to understand the code imo.

src/backend/core/permissions/backends/base.py
```diff
@@ -37,6 +37,10 @@ def ancestors_links_paths_mapping(self, item: models.Item) -> dict[str, list[dic
     def link_definition_for(self, item: models.Item) -> dict:
         """Return the effective link definition of the item, own and inherited combined."""
 
+    @abstractmethod
+    def abilities(self, user: models.User | AnonymousUser, item: models.Item) -> dict:
```
sampaccoud, line 29: It would have been nice to move it in the first commit creating the backend since the get_abilities method was already existing.

src/backend/core/permissions/backends/role.py
```diff
@@ -143,6 +161,7 @@ def abilities(self, user, item):  # pylint: disable=too-many-locals
             and bool(target_extension_for(item.extension))
             and bool(settings.WOPI_ONLYOFFICE_CONVERT_JWT_SECRET)
         )
+        can_restrict = is_owner and not is_deleted and item.type == models.ItemTypeChoices.FOLDER
```
sampaccoud: why not allow restricting a file? e.g I add an instruction file at the root of a public folder and want to restrict it to my colleagues.

src/backend/core/permissions/backends/role.py
```diff
@@ -119,12 +119,30 @@ def abilities(self, user, item):  # pylint: disable=too-many-locals
...
+            and item.is_restricted
+            and item.depth > 1
+        )
+        if needs_container_owner_check:
+            parent = (
+                models.Item.objects.annotate_user_roles(user)
+                .filter(path=str(item.path[:-1]))
+                .first()
+            )
+            is_container_owner = parent is not None and parent.get_role(user) == RoleChoices.OWNER
```
sampaccoud: parent should not be None since you checked depth > 1

src/backend/core/models.py
```diff
@@ -1463,6 +1465,18 @@ def move(self, target):
...
+    def activate_restriction(self, user: models.User) -> None:
+        """Activate restricted access on this folder and ensure the user has an explicit owner."""
+        ItemAccess.objects.update_or_create(
+            item=self, user=user, defaults={"role": RoleChoices.OWNER}
+        )
+        update_fields = ["is_restricted"]
+        self.is_restricted = True
+        if self.link_reach is None:
+            self.link_reach = LinkReachChoices.RESTRICTED
+            update_fields.append("link_reach")
```
sampaccoud: I think that if self.link_reach is not None you should also force it to RESTRICTED.
You a restricting access to an item, not just the explicit accesses to it but also public access! 
If I want to make this item public as well I should have to do it explicitly on the restricted item after restricting it.

src/backend/core/models.py
```diff
@@ -1477,6 +1477,42 @@ def activate_restriction(self, user: models.User) -> None:
             update_fields.append("link_reach")
         self.save(update_fields=update_fields)
 
+    def deactivate_restriction(self) -> None:
```
sampaccoud: this whole method must be atomic!? Same for activation

src/backend/core/models.py
```diff
@@ -1477,6 +1477,42 @@ def activate_restriction(self, user: models.User) -> None:
...
+            .values_list("user_id", "team", "role")
+        )
+        inherited_roles = {}
+        for user_id, team, role in inherited_accesses:
+            key = (user_id, team)
+            inherited_roles[key] = RoleChoices.max(inherited_roles.get(key), role)
+
+        redundant_ids = [
+            access.id
+            for access in ItemAccess.objects.filter(item=self)
```
sampaccoud, line 1714: You have the same problem on all its descendants... Maybe you should test that these redundant accesses are not causing any functional problem and if this is confirmed, you can clean them in an asynchronous job?

But actually I'm having second thoughts : IF after removing restriction, an access is more open than what is defined on the parent, I think the user should be warned! Because it is an open access that was not effective because of restriction and will now be effective for all users of the tree above! 
So now I think everything should be reset on the whole subtree when a restriction is removed.

src/backend/core/tests/test_models_items_restricted.py
```diff
@@ -309,6 +309,46 @@ def test_models_items_restricted_deactivate_restriction_keeps_link_reach_without
     assert folder.link_role == "reader"
 
 
+def test_models_items_restricted_uproot_moves_to_root():
```
sampaccoud: your commit message reads: "When an ancestor owner removes a restricted folder they don't
have access to"
How could I remove it if I don't see it?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to
 - ✨(frontend) add file type, contact and modification date topbar filters
 - ✨(frontend) add location, file type, contact and date search filters
 - ♻️(backend) route permission decisions through a swappable backend
+- ✨(backend) preserve restricted folders when an ancestor is soft-deleted
```
sampaccoud: This is going to create confusion and break things. Instead you should forbid deletion by a user if there is a restricted folder on which s.he does not have deletion rights...
Uprooting the restricted folder can be suggested but should be done explicitly... Maybe it means that owners should see restricted folders (greyed) and be able to move them out of their folders even if they have no rights on them...

src/backend/core/permissions/backends/base.py
```diff
@@ -62,6 +62,15 @@ def propagation_scope(self, item: models.Item) -> QuerySet[models.Item]:
...
         """Return the ancestors of the item, itself included, down to its restriction boundary."""
 
+    def visible(
+        self,
+        queryset: QuerySet[models.Item],
+        user: models.User | AnonymousUser,
+        path_field: str = "path",
+    ) -> QuerySet[models.Item]:
+        """Filter the queryset to the rows on which the user holds a role."""
+        return self.annotate_roles(queryset, user, path_field=path_field).exclude(user_roles=[])
```
sampaccoud: This is not a correct : 
- the semantic "visible" because an item may be visible through a link_reach
- the filtering because you will exclude the items that the user can see thanks to link_reach

src/backend/core/services/item_exports.py
```diff
@@ -39,7 +43,10 @@ def export_descendants(folder):
     a directory entry (`file_key=None`, trailing slash) for folders or a file
     entry for `FILE` items in the `READY` upload state.
     """
-    descendants = folder.descendants().filter(ancestors_deleted_at__isnull=True).order_by("path")
+    descendants = folder.descendants().filter(ancestors_deleted_at__isnull=True)
+    if user and user.is_authenticated:
+        descendants = get_permissions_backend().visible(descendants, user)
```
sampaccoud: Same. This is not correct because it forgets about link_reach.
This time it seems that you painted yourself in a corner as fixing this would lead to awful code and performance issues.

In the end, I think the correct way to do restriction is probably to move items that are getting restricted to the root of the tree and materialize their appearance in the origin folder by an item of type "restriction_link". By doing this, the whole access right chain will work without any modification and the item can be moved back to its location when the restriction is lifted.

#### before restriction

└── containing_folder (0001)
    ├── folder1 (00010001)
    │    ├── file1 (000100010001)
    │    └── file2 (000100010002)
    ├── folder2 (00010002)
    └── file3 (00010003)

#### After restriction

├── folder1 (0002)
│    ├── file1 (00020001)
│    └── file2 (00020002)
└── containing_folder (0001)
    ├── folder1 [restriction shortcut → 0002] (00010001)
    ├── folder2 (00010002)
    └── file3 (00010003)

Instead of modifying all the access rules and making the code buggy and complicated, you just need to handle : 
- hide it from the real root : easy since they are flagged as restricted items... 
- show them in their folder of origin : easy since you can add a shortcut of type "restriction" in the said folder

sampaccoud:
This review was made commit by commit. The first commits were fully commented as everything looked fine but at some point, the commits starting fixing caveats and everything goes wrong. The conclusion is that the method is a deadend and leads to code that is hard to understand and I'm afraid will never work (link reach was forgotten in restricted items accesses).

I ended-up suggesting another approach that should be simpler and work well : move an item to the root when it is being restricted. The review was aborted after this point.

lterm:
Thanks @sampaccoud for the review. So the alternative is to move a restricted folder to root, with a "restriction link" placeholder marking its original spot.

A few UX/UI questions this raises:
- This restriction link ties directly into the shortcuts feature : should we plan a specific representation for links to restricted folders, distinct from regular shortcuts?
- What does the flow look like when a user activates/deactivates the restriction ? do we walk them through what's about to happen (e.g. an info message or confirmation step explaining the folder will be moved to root)?
- What happens when someone tries to delete a restricted folder they don't have access to?

@robin-lecomte, what do you think?

sampaccoud:
> * This restriction link ties directly into the shortcuts feature : should we plan a specific representation for links to restricted folders, distinct from regular shortcuts?

I could be a separate object or a shortcut flagged as "restriction" :thinking: 

> * What does the flow look like when a user activates/deactivates the restriction ? do we walk them through what's about to happen (e.g. an info message or confirmation step explaining the folder will be moved to root)?

You don't need to tell them about moving the item to the root (plus it will be hidden from the root..)
You probably need to explain something when a restriction is activated: that the item's permissions are going to be detached from the above tree structure and that access rights inheritance will not apply anymore. Maybe let them know that this can be reverted?
Same when deactivating, warn them that access rights will now inherit from the tree structure in which the item is located...

The reason for informing the user is that access rights on the item are changing, not that we move the item to/from the root (not his business and only a technical subject)

> * What happens when someone tries to delete a restricted folder they don't have access to?

This was discussed elsewhere in my review with several cases:
- the user is NOT owner of the folder containing the restricted item : 
    - he can only see the restricted item if he has read access to it.
    - he can only delete the restricted item if he is a owner on it. 
- the user is owner of the containing folder :
    - he can see the restricted item even without read access to it (in this case it is grey and can't be entered into)
    - he can only delete the restricted item if he is a owner on it. Eventually, he can manually detach it for good before being allowed to delete the containing folder... In the current version of the PR, this issue was solved by moving the restricted folder in order to be able to delete the containing folder... With the pattern I propose, detachment is already done and we only have to delete the restriction shortcut and reset the `is_restricted` flag on the now root folder ... easy!

kernicPanel:
@sampaccoud @lterm 
I reworked this PR with the suggested approach.
I updated the description accordingly.

 As most inline threads target code that no longer exists, a few comments need answers:
- Restriction stays folder-only: This is part of the rules found in specs, but it can be changed if needed.
- Explicit link reach is preserved on activation: Also a spec rule
- Deactivation only cleans up the folder's own explicit accesses. Values set on its subfolders and files during restriction are kept, they were set on purpose. Warning the user about restored inheritance belongs to the frontend flow.

src/backend/core/models.py
```diff
@@ -1428,6 +1428,10 @@ def soft_delete(self):
                 ancestors_deleted_at__isnull=True,
             ).delete()
 
+        # No shortcut may keep pointing into the trash
+        if self.is_restricted:
+            self._meta.model.objects.filter(target=self).delete()
+
```
sampaccoud: This is not the behavior we want. Restoring from trash should revert to previous situation. You should mark the shortcut as deleted_at so you can revert it

src/backend/core/models.py
```diff
@@ -1420,6 +1420,14 @@ def soft_delete(self):
...
             )
 
+        # Restricted folders survive their container: delete (detach) the
+        # shortcuts of the subtree instead of trashing them
+        if self.type == ItemTypeChoices.FOLDER:
+            self.descendants().filter(
+                type=ItemTypeChoices.SHORTCUT,
+                ancestors_deleted_at__isnull=True,
+            ).delete()
+
```
sampaccoud: Same. Restoring should revert to the previous situation and the folder in the trash should still show its content. I agree that the restricted folder should now appear in the list view but this should be handled by not excluding from list viw a restricted folder targeted by a soft deleted shortcut.... you can clean up everything when the trashed item is really deleted from db.

sampaccoud (commented):
This review was made by reading just the commit titles and looking at the code only if I had doubts on what was mentioned.

The list of commits makes me very confident that this PR is now ok. Since I didn't bring my laptop on holidays I would suggest to request other reviews without waiting !

src/backend/core/permissions/backends/role.py
NathanVss: Could be put in the parent class?

src/backend/core/models.py
```diff
@@ -1314,93 +1313,8 @@ def computed_link_role(self):
...
-            "tree": can_get,
-            "media_auth": can_get,
-            "partial_update": can_update,
-            "update": can_update,
-            "upload_ended": can_update and user.is_authenticated,
-            "wopi": can_get,
-            "convert": can_convert,
-        }
+        """Compute and return abilities for a given user on the item."""
+        return get_permissions_backend().abilities(user, self)
```
NathanVss, line 1356: Yay <3

src/backend/core/permissions/backends/role.py
```diff
@@ -10,109 +15,200 @@
 from wopi.conversion.policy import target_extension_for
 
 
+class ItemAbilities:  # pylint: disable=too-many-public-methods
```
NathanVss, line 18: Love it

src/backend/core/models.py
```diff
@@ -1020,6 +1021,13 @@ class Item(TreeModel, BaseModel):
     )
     mimetype = models.CharField(max_length=255, null=True, blank=True)
     is_restricted = models.BooleanField(default=False)
+    target = models.OneToOneField(
```
NathanVss: In the future we might want multiple shortcut to the same folder, nope?

src/backend/core/models.py
```diff
@@ -1647,6 +1647,44 @@ def restrict(self, user):
...
+            raise ValidationError(
+                {
+                    "is_restricted": ValidationError(
+                        _("This folder is not restricted"),
+                        code="item_unrestrict_not_restricted",
+                    )
+                }
+            )
+
+        shortcut = self._meta.model.objects.select_for_update().filter(target=self).first()
```
NathanVss: Is `select_for_update` mandatory?

src/backend/core/tests/items/test_api_items_restrict.py
NathanVss, line 1: How do you make sure restricted folders are not visibles at the root?

src/backend/core/api/viewsets.py
```diff
@@ -763,6 +763,32 @@ def list(self, request, *args, **kwargs):
...
+                | db.Exists(
+                    models.Item.objects.filter(
+                        path__ancestors=db.OuterRef("path"),
+                        link_reach__in=[
+                            LinkReachChoices.PUBLIC,
+                            LinkReachChoices.AUTHENTICATED,
+                        ],
+                    )
+                )
+            )
```
NathanVss, line 803: I hope this scales well in a huge db

src/backend/core/models.py
```diff
@@ -1420,6 +1420,14 @@ def soft_delete(self):
                 "Cannot delete this item because one or more ancestors are already deleted."
             )
 
+        # Restricted folders survive their container: delete (detach) the
+        # shortcuts of the subtree instead of trashing them
+        if self.type == ItemTypeChoices.FOLDER:
+            self.descendants().filter(
+                type=ItemTypeChoices.SHORTCUT,
+                ancestors_deleted_at__isnull=True,
+            ).delete()
```
NathanVss: Why don't we keep them? If we restore we will loose them. 

Soft delete and restore should return to the original state? Maybe I'm missing a point

src/backend/core/services/item_exports.py
```diff
@@ -43,6 +43,10 @@ def export_descendants(folder):
 
     relative_paths = {str(folder.path): ""}
     for descendant in descendants:
+        # Shortcuts are tree entries: their target lives in another subtree
+        if descendant.type == models.ItemTypeChoices.SHORTCUT:
+            continue
+
```
NathanVss, line 49: In the case of an extract, it looks more instinctive to me that, if you can access a restricted folder, then the export should contain it. (its target to be more specific)

src/backend/core/api/viewsets.py
```diff
@@ -1145,11 +1151,22 @@ def children(self, request, *args, **kwargs):
...
-                creator=request.user,
-                parent=item,
-                **serializer.validated_data,
-            )
+            # The child must be created unrestricted first: activation moves
+            # it to the root and refuses an already restricted folder
+            is_restricted = serializer.validated_data.pop("is_restricted", False)
+            if is_restricted and item.get_role(request.user) != models.RoleChoices.OWNER:
+                raise drf.exceptions.PermissionDenied(
+                    "Only owners of the parent folder can create a restricted folder."
```
NathanVss, line 1165: Why no allowing administrators?

NathanVss (commented):
I really like the approach, feels like really simple and KISS. 

Good job @kernicPanel

src/backend/core/models.py
NathanVss: I understand that for now we have no choice in order to handle some operation related to restricted folders, but to be future proof it might be useful to think now about this to avoid locking the feature in the future.

src/backend/core/models.py
kernicPanel: The lock on the folder row serializes concurrent unrestrict/restrict calls.
The lock on the shortcut row pins its parent between the moment we read it and the moment we delete the row: without it, a concurrent move() of the shortcut could reattach the folder at a stale location.

src/backend/core/tests/items/test_api_items_restrict.py
kernicPanel: Covered in core/tests/items/test_api_items_list_restricted.py: the restricted root is hidden from the top-level listing whenever the user reaches a live shortcut (direct access, team access, direct or inherited link reach), and shows up again when the shortcut is gone or trashed.

src/backend/core/api/viewsets.py
kernicPanel: The exclusion is two correlated EXISTS backed by the GiST index on path, and it only runs against the restricted roots of the page being listed. The query count is constant regardless of the number of restricted roots (test_api_items_list_restricted_constant_queries), and I can share local bench numbers if useful.

src/backend/core/services/item_exports.py
kernicPanel: Spec choice for now: a restricted folder lives in a single location and is exported from there by its members, so the container export mirrors what the container actually holds.

src/backend/core/api/viewsets.py
kernicPanel: Consistency with the restrict ability: per the specs restriction is owner-only (the toggle on an existing folder also requires an owner), and creating a folder restricted from the start is just a shorthand for create + restrict.

src/backend/core/models.py
lunika: Yes but it will not take the advantage of the cache present in `self.user_roles`

src/backend/core/permissions/backends/role.py
```diff
@@ -10,105 +15,196 @@
...
+    def can_update(self) -> bool:
+        """Return whether the user can modify the item."""
+        return (self.is_owner_or_admin or self.role == RoleChoices.EDITOR) and not self.is_deleted
+
+    @property
+    def can_create_children(self) -> bool:
+        """Return whether the user can create children in the item."""
+        return self.can_update and self.user.is_authenticated
+
+    @cached_property
```
lunika, line 97: Why not using `@cached_property` for each ? call `can_update` twice should always return the same result for the same instance.

src/backend/core/models.py
```diff
@@ -1055,6 +1063,17 @@ class Meta:
...
+                condition=(
+                    (models.Q(type=ItemTypeChoices.SHORTCUT) & models.Q(target__isnull=False))
+                    | (~models.Q(type=ItemTypeChoices.SHORTCUT) & models.Q(target__isnull=True))
+                ),
+                name="check_target_only_on_shortcuts",
+            ),
+            models.UniqueConstraint(
+                fields=["target"],
+                name="unique_shortcut_per_target",
+            ),
```
lunika, line 1076: A shortcut can be targeted by something else than a folder ?

src/backend/core/models.py
```diff
@@ -1568,6 +1592,74 @@ def move(self, target):
...
+                        code="item_restrict_deleted",
+                    )
+                }
+            )
+
+        parent = item.parent()
+
+        ItemAccess.objects.update_or_create(
+            item=item, user=user, defaults={"role": RoleChoices.OWNER}
+        )
```
lunika, line 1678: Maybe it is overkill, but in the abilities, you check that the user must be an explicit owner. You should probably check here also ? All other checks made before are also checks on the abilities, if I'm not mistaken.

src/backend/core/models.py
```diff
@@ -1568,6 +1592,74 @@ def move(self, target):
...
+
+        # The move must run while the flag is still unset: move() refuses
+        # restricted folders
+        item.move(None)
+
+        item.is_restricted = True
+        if item.link_reach is None:
+            item.link_reach = LinkReachChoices.RESTRICTED
+        item.save(update_fields=["is_restricted", "link_reach"])
```
```suggestion
        item.save(update_fields=["is_restricted", "link_reach", "updated_at"])
```

src/backend/core/models.py
```diff
@@ -1660,6 +1660,44 @@ def restrict(self, user):
 
         return item
 
+    @transaction.atomic
+    def unrestrict(self):
+        """Lift restriction and reattach the folder at its shortcut location."""
+        self._meta.model.objects.select_for_update().get(pk=self.pk)
```
lunika: is it working to not use the return value ?

src/backend/core/models.py
```diff
@@ -1660,6 +1660,44 @@ def restrict(self, user):
...
+        self.refresh_from_db()
```
lunika: Using the previous return value can avoid this `refresh_from_db()` ?

src/backend/core/models.py
```diff
@@ -1660,6 +1660,44 @@ def restrict(self, user):
...
+        if parent:
+            self.title = manage_unique_title_utils(
+                self._meta.model.objects.children(parent.path), self.title
+            )
+            self.save(update_fields=["title"])
+            self.move(parent)
+
+        self.invalidate_nb_accesses_cache()
+
+        return self
```
lunika: IMO it is useless to return self unless you want to allow chaining other calls on the model

src/backend/core/models.py
```diff
@@ -1660,6 +1660,44 @@ def restrict(self, user):
...
+        shortcut = self._meta.model.objects.select_for_update().filter(target=self).first()
+        parent = None
+        if shortcut:
+            if shortcut.ancestors_deleted_at is None:
+                parent = shortcut.parent()
+            self._meta.model.objects.filter(pk=shortcut.pk).delete()
+
+        # The flag must fall before the move: move() refuses restricted folders
+        self.is_restricted = False
+        self.save(update_fields=["is_restricted"])
```
```suggestion
        self.save(update_fields=["is_restricted", "updated_at"])
```

src/backend/core/models.py
```diff
@@ -1660,6 +1660,44 @@ def restrict(self, user):
...
+
+        if parent:
+            self.title = manage_unique_title_utils(
+                self._meta.model.objects.children(parent.path), self.title
+            )
+            self.save(update_fields=["title"])
```
```suggestion
            self.save(update_fields=["title", "updated_at"])
```

src/backend/core/tests/test_models_items_restricted.py
```diff
@@ -353,6 +353,82 @@ def test_models_items_restricted_unrestrict_without_live_shortcut():
...
+        type=models.ItemTypeChoices.FOLDER,
+        users=[(parent_owner, models.RoleChoices.OWNER)],
+    )
+    folder = factories.ItemFactory(parent=parent, type=models.ItemTypeChoices.FOLDER)
+    folder = folder.restrict(user)
+    factories.UserItemAccessFactory(item=folder, user=parent_owner, role="editor")
+
+    folder = folder.unrestrict()
+
+    assert not models.ItemAccess.objects.filter(item=folder, user=parent_owner).exists()
```
lunika, line 385: Does `parent_owner` is `owner` again ?

src/backend/core/models.py
```diff
@@ -1681,6 +1681,17 @@ def _normalize_explicit_accesses(self):
...
+    def _normalize_explicit_link_reach(self):
+        """Reset the link reach to inherit when inferior or equal to the inherited one."""
+        # The cached ancestors definition predates the move, recompute it
+        self._ancestors_link_definition = None
+        inherited_reach = self.ancestors_link_definition["link_reach"]
+        if LinkReachChoices.get_priority(self.link_reach) <= LinkReachChoices.get_priority(
+            inherited_reach
+        ):
+            self.link_reach = None
+            self.save(update_fields=["link_reach"])
```
```suggestion
            self.save(update_fields=["link_reach", "updated_at"])
```

src/backend/core/api/serializers.py
```diff
@@ -536,7 +539,19 @@ def create(self, validated_data):
...
+            user = self.context["request"].user
+            if not instance.get_abilities(user).get("restrict"):
+                raise exceptions.PermissionDenied()
+            # Toggling restriction moves the item: keep working on the
+            # returned instance so its refreshed path is not overwritten
+            if is_restricted:
+                instance = instance.restrict(user)
+            else:
+                instance = instance.unrestrict()
+
```
lunika: Why doing this way instead of having the traditional pattern 1 viewset action by ability ? With an action `POST restrict` and `DELETE restrict` ?

Here you are mixing multiple action in an update of the item.

lunika: One more argument: by modifying this property, you are modifying other properties on the model (the path is changed, the link_reach can change, etc), I'm not sure if this can have a side effect on the frontend application.

src/backend/core/permissions/backends/role.py
kernicPanel: As all `@property` here depend on other  `@cached_property`, it was already the case.
Though, I moved them all to `@cached_property` for easier understanding.

src/backend/core/models.py
kernicPanel: For now, it can't. Shortcuts are only created through `restrict()`, which checks for a folder.
But I don't think we should add a db constraint unless needed.

src/backend/core/models.py
kernicPanel: Good point, and not overkill: `restrict()` grants an explicit owner access to the calling user, so it must not rely on callers for the permission check.
It now validates the user holds an owner role (direct or inherited) like the ability does.

src/backend/core/models.py
kernicPanel: Yes it was working, but I replaced with the same approach than `restrict()` for consistency, and for removing `refresh_from_db()` us you pointed out.

src/backend/core/models.py
kernicPanel: Indeed, done.

src/backend/core/models.py
kernicPanel: The return value is needed: the serializer keeps working on the returned instance because the path and title may change on deactivation (same pattern as `restrict()`).
After the rework it also carries the fresh locked instance instead of the stale one.

src/backend/core/tests/test_models_items_restricted.py
kernicPanel: Yes, I added an assert.

docker/auth/realm.json
```diff
@@ -64,6 +64,62 @@
         ],
         "realmRoles": ["user"]
       },
+      {
+        "username": "paige",
+        "email": "page.turner@library.book",
```
sampaccoud: paige?

src/backend/core/permissions/backends/role.py
```diff
@@ -1,7 +1,12 @@
 """Role-based permissions backend."""
 
+from __future__ import annotations
```
sampaccoud, line 3: I agree that this get_abilities method has become very big but what this commit proposes looks like object-maxing to me.

I’m concerned that `cached_property` hides the cost of permission checks. Attribute access can trigger DB work, making performance and N+1s much harder to reason about. This will lead to convoluted and inefficient code in the future.

I’d keep `ItemAbilities` as the API, but make the expensive computes explicit and compute them once when the object is created or when a called property requires it if you really want to optimize... but in the case of API calls you need to build the whole dict anyway so KISS. 

The individual abilities can then remain small, pure methods and you avoid using `cached_property` systematically without explicit reason (the next developer will forget it):

```python
class ItemAbilities:
    def __init__(self, user, item):
        self.user = user
        self.item = item

        # Explicitly compute/load anything that may hit the DB.
        self.role = item.get_role(user)
        self.link_definition = item.computed_link_definition
        self.is_deleted = bool(item.ancestors_deleted_at)

    def can_update(self):
        return (
            self.role in {RoleChoices.OWNER, RoleChoices.ADMIN}
            and not self.is_deleted
        )

    def can_delete(self):
        return (
            self.role == RoleChoices.OWNER
            and not self.is_deleted
        )

    def can_get(self):
        return self.role is not None and not self.is_deleted

    def as_dict(self):
        return {
            "can_update": self.can_update(),
            "can_delete": self.can_delete(),
            "can_get": self.can_get(),
        }
```

This preserves the readability of one rule per ability, but avoids making each attribute access a potentially expensive, hidden computation. It also gives us one obvious place to optimize or batch the DB work later.r

Add tests to secure number of db calls if missing and make use of these new properties in python code now that we have them?

src/backend/core/tests/test_models_items.py
```diff
@@ -101,7 +101,12 @@ def test_models_items_soft_delete(depth):
 
     with pytest.raises(RuntimeError) as exc_info:
         items[-1].soft_delete()
-        assert str(exc_info) == "This item is already deleted or has deleted ancestors."
```
sampaccoud, line 104: I didn't get why it's never caught. It never ran because the error was not a RuntimeError? Then it's the error the being caught that we should change?

src/backend/core/tests/test_models_items.py
```diff
@@ -55,6 +55,18 @@ def test_models_items_path_for_children_contains_parent_path():
...
+def test_models_items_parent_resolves_by_path_after_move():
+    """The parent should stay the direct one after the subtree was moved."""
+    parent = factories.ItemFactory(type=models.ItemTypeChoices.FOLDER)
+    child = factories.ItemFactory(parent=parent, type=models.ItemTypeChoices.FOLDER)
+    new_grandparent = factories.ItemFactory(type=models.ItemTypeChoices.FOLDER)
+
+    parent.move(new_grandparent)
+    child.refresh_from_db()
+
+    assert child.parent() == parent
```
sampaccoud, line 67: So this is a bug in django-ltree right? Should we make a pull request on the lib as well?l

src/backend/core/tests/test_models_items_shortcuts.py
```diff
@@ -0,0 +1,58 @@
...
+def test_models_items_shortcuts_target_forbidden_on_other_types(item_type):
+    """Only shortcuts can carry a target."""
+    target = factories.ItemFactory(type=models.ItemTypeChoices.FOLDER, is_restricted=True)
+
+    with pytest.raises(ValidationError):
+        factories.ItemFactory(type=item_type, target=target)
+
+
+def test_models_items_shortcuts_unique_per_target():
+    """A restricted folder cannot be targeted by two shortcuts."""
```
sampaccoud: I understand that you decided not to use the same object for the shortcut feature that will come next and that will probably need to allow several shortcuts per item?
It's probably a good decision that the restricted access target uses a specific item type but you should use another name that is more explicit and keep the name "shortcut" for the future shortcuts?

src/backend/core/models.py
```diff
@@ -71,6 +71,7 @@ class ItemTypeChoices(models.TextChoices):
 
     FOLDER = "folder", _("Folder")
     FILE = "file", _("File")
+    SHORTCUT = "shortcut", _("Shortcut")
```
sampaccoud: RESTRICTION?

src/backend/core/models.py
```diff
@@ -1020,6 +1021,13 @@ class Item(TreeModel, BaseModel):
     )
     mimetype = models.CharField(max_length=255, null=True, blank=True)
     is_restricted = models.BooleanField(default=False)
+    target = models.ForeignKey(
```
sampaccoud: If this is unique you should use a OneToOneField?

src/backend/core/permissions/backends/role.py
```diff
@@ -137,7 +137,19 @@ def can_convert(self) -> bool:
...
         )
 
-    @property
+    @cached_property
+    def can_restrict(self) -> bool:
+        """Return whether the user can toggle restriction on the folder."""
+        # A restricted folder lives at the tree root: deactivation must stay
+        # possible there, while activation requires a parent for the shortcut
+        return (
+            self.is_owner
```
sampaccoud, line 135: Start by the properties that don't trigger any db query?

src/backend/core/models.py
```diff
@@ -1568,6 +1592,83 @@ def move(self, target):
...
+    @transaction.atomic
+    def restrict(self, user):
+        """Restrict the folder by detaching it to the tree root behind a shortcut."""
+        item = self._meta.model.objects.select_for_update().get(pk=self.pk)
+
+        if item.type != ItemTypeChoices.FOLDER:
+            raise ValidationError(
+                {
+                    "is_restricted": ValidationError(
+                        _("Only folders can be restricted"),
```
sampaccoud, line 1626: this should be a db constraint of the type and is_resticted columns [second thought, is_resticted column should be removed]

src/backend/core/models.py
```diff
@@ -1568,6 +1592,83 @@ def move(self, target):
...
+                    "is_restricted": ValidationError(
+                        _("This folder is already restricted"),
+                        code="item_restrict_already_restricted",
+                    )
+                }
+            )
+        if item.depth == 1:
+            raise ValidationError(
+                {
+                    "is_restricted": ValidationError(
```
sampaccoud, line 1643: add a db constraint on this as well?

src/backend/core/models.py
```diff
@@ -1568,6 +1592,83 @@ def move(self, target):
...
+                        _("A root folder cannot be restricted"),
+                        code="item_restrict_root",
+                    )
+                }
+            )
+        if item.ancestors_deleted_at:
+            raise ValidationError(
+                {
+                    "is_restricted": ValidationError(
```
sampaccoud, line 1652: also db constraint on both deleted_at and ancestors_deleted_at fields paired with is_restricted field

src/backend/core/models.py
```diff
@@ -1568,6 +1592,83 @@ def move(self, target):
                 path=RawSQL("%s || subpath(path, nlevel(%s))", (str(self.path), str(old_path)))
             )
 
+    @transaction.atomic
```
sampaccoud, line 1617: add a test on atomicity by simulating a break in the middle of the method and checking that the folder is not broken?

src/backend/core/models.py
```diff
@@ -1669,6 +1669,43 @@ def restrict(self, user):
...
+        shortcut = self._meta.model.objects.select_for_update().filter(target=item).first()
+        parent = None
+        if shortcut:
+            if shortcut.ancestors_deleted_at is None:
+                parent = shortcut.parent()
+            self._meta.model.objects.filter(pk=shortcut.pk).delete()
+
+        # The flag must fall before the move: move() refuses restricted folders
+        item.is_restricted = False
+        item.save(update_fields=["is_restricted", "updated_at"])
```
sampaccoud: I'm thinking: this is_restricted field is a weak and breaks information unicity. 
An item is restricted when it is targeted by a restriction type item... so we shouldn't make it a db column.

src/backend/core/api/serializers.py
```diff
@@ -221,6 +222,43 @@ class Meta:
...
+    class Meta:
+        model = models.Item
+        fields = ["id", "title", "is_restricted", "deleted", "can_access"]
+        read_only_fields = ["id", "title", "is_restricted", "deleted", "can_access"]
+
+    def get_deleted(self, target) -> bool:
+        """Return whether the target is in the trash."""
+        return target.deleted_at is not None
+
+    def get_can_access(self, target) -> bool:
```
sampaccoud, line 245: can_access should be in abilities to respect the pattern used elsewhere?

src/backend/core/api/viewsets.py
```diff
@@ -768,6 +768,32 @@ def list(self, request, *args, **kwargs):
         )
         queryset = queryset.filter(path__in=root_paths)
 
+        # Hide restricted roots the user already reaches through a live
```
sampaccoud, line 774: the fact that you have to do this reinforces my earlier take on removing the is_restricted field... the only remaining case for it was that it would ease filtering on the root... But in fact filtering on the root is not automatic for all restricted folders...

src/backend/core/api/viewsets.py
```diff
@@ -1151,12 +1157,23 @@ def children(self, request, *args, **kwargs):
...
+                )
+
+            with transaction.atomic():
+                child_item = models.Item.objects.create_child(
+                    creator=request.user,
+                    parent=item,
+                    **serializer.validated_data,
+                    **self.get_create_extra_attributes(),
+                )
+                if is_restricted:
```
sampaccoud: I don't see the point of allowing this. It's seems simpler and more robust to consider restriction as a definite action taken on an already existing folder?

src/backend/core/tests/test_models_items.py
kernicPanel: As `soft_delete()` raises, the following code is skipped, so the assert didn't get executed.

src/backend/core/tests/test_models_items.py
kernicPanel: Yes it is, we could open a PR, yes.

src/backend/core/tests/test_models_items_shortcuts.py
kernicPanel: You're right, I renamed it to "restriction" in a separate commit for now. We'll see if fixups are preferable.

src/backend/core/models.py
kernicPanel: We need to keep the is_restricted column to match the spec: the initial shortcut may be deleted, but the targeted folder should be kept, and marked as restricted.
Please see rules 9 and 10 in  https://docs.numerique.gouv.fr/docs/8cfe8d2e-83b7-4750-8a13-8b08fa26d466/

src/backend/core/api/serializers.py
kernicPanel: I don't think we should regroup this with existing abilities as it would be confusing: the restriction already exposes its own abilities, and can_access is just a display hint on its target.
Also, real abilities there would cost one role query per row.

src/backend/core/models.py
sampaccoud: if the initial shortcut is deleted, the folder is not restricted anymore... it de-facto becomes a real root item and everything works. So rules 9 and 10 are respected.

src/backend/core/tests/test_models_items.py
sampaccoud: oh I see

src/backend/core/models.py
kernicPanel: Ok, I finally removed it. All rules seems followed.

src/backend/core/permissions/backends/role.py
NathanVss: I agree with that, we should implement tests in a way that we make sure that no DB requests are made in `can_*` methods so we make sure we enforce this logic long term.

kernicPanel: I do too, so I did it, and resolved the conversation.

src/backend/core/permissions/backends/role.py
```diff
@@ -0,0 +1,192 @@
...
+            "destroy": self.can_destroy(),
+            "download": self.can_get(),
+            "duplicate": self.can_duplicate(),
+            "export": self.can_export(),
+            "hard_delete": self.can_hard_delete(),
+            "favorite": self.can_favorite(),
+            "link_configuration": self.can_manage(),
+            "invite_owner": self.can_invite_owner(),
+            "link_select_options": self.link_select_options(),
+            "move": self.can_manage(),
```
NathanVss: It seems using the API we will be able to move restricted targets that are located on root

src/backend/core/models.py
```diff
@@ -1630,6 +1614,160 @@ def move(self, target):
...
+        if item.ancestors_deleted_at:
+            raise ValidationError(
+                {
+                    "is_restricted": ValidationError(
+                        _("A deleted folder cannot be restricted"),
+                        code="item_restrict_deleted",
+                    )
+                }
+            )
+        if item.get_role(user) != RoleChoices.OWNER:
```
NathanVss, line 1658: There is a logic dup, this part should rely on the `PermissionBackend` that already defines the restrict ability.

NathanVss (approved):

docker/auth/realm.json
```diff
@@ -64,6 +64,62 @@
...
+          {
+            "type": "password",
+            "value": "pass"
+          }
+        ],
+        "realmRoles": ["user"]
+      },
+      {
+        "username": "mark",
+        "email": "mark.down@plain.text",
```
lunika, line 111: I love this one ❤️

lunika (approved):

PR #764, opened 2026-07-07
✨(global) add storage gauge

NathanVss:
## Purpose

Add a storage gauge, indicating the user how much storage is remaining.

See https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=14155-57350&t=BDVBrHbIVUYgp1JN-0 

Capture d’écran 2026-07-07 à 15 23 22
Capture d’écran 2026-07-07 à 15 23 42


src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -458,9 +472,17 @@ export class StandardDriver extends Driver {
...
-    });
+    const response = await fetchAPI(
+      `items/${id}/duplicate/`,
+      {
+        method: "POST",
+      },
+      {
+        // A duplication can be rejected (e.g. quota gate): let the global
+        // handler toast the error instead of redirecting to the 403 page.
+        redirectOn40x: false,
```
PanchoutNathan, line 487: ditto

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -163,10 +163,18 @@ export class StandardDriver extends Driver {
...
+    await fetchAPI(
+      `items/${id}/move/`,
+      {
+        method: "POST",
+        body: JSON.stringify(payload),
+      },
+      {
+        // A move can be rejected (e.g. quota gate): let the caller toast the
+        // error instead of redirecting the whole app to the 403 page.
+        redirectOn40x: false,
```
PanchoutNathan, line 175: And what if it's a 401?

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
```diff
@@ -417,10 +425,16 @@ export class StandardDriver extends Driver {
...
-        signal: abortController.signal,
-      });
+      await fetchAPI(
+        `items/${item.id}/upload-ended/`,
+        {
+          method: "POST",
+          signal: abortController.signal,
+        },
+        {
+          redirectOn40x: false,
```
PanchoutNathan, line 439: ditto

src/frontend/apps/drive/src/features/explorer/components/ExplorerDndProvider.tsx
```diff
@@ -116,20 +116,24 @@ export const ExplorerDndProvider = ({ children }: ExplorerDndProviderProps) => {
...
+        {
+          onSuccess: () => {
+            addItemsMovedToast(ids.length);
+            // Reset the selected items after the move
+            setSelectedItems([]);
+          },
+        },
+      )
+      .catch(() => {
+        // The error feedback is already handled by the mutation's onError.
```
PanchoutNathan, line 135: Not a regression , but unfortunate: the tree is modified before the request, then the error is swallowed without a rollback. After a quota rejection, the document visually remains in the wrong folder until the page is reloaded.We modify the tree data right before making the request.

src/frontend/apps/drive/src/features/explorer/hooks/useRefreshItems.ts
```diff
@@ -14,14 +14,19 @@ export const useGetQueryKeyToRefresh = () => {
...
     }
-    // let queryKey = parentId ? ["items", parentId, "children"] : [];
-    // if (queryKeyForRoute.length > 0) {
-    //   queryKey = queryKeyForRoute;
-    // }
     return queryKeys;
   };
 };
 
+export const useRefreshEntitlementsQueryCache = () => {
```
PanchoutNathan, line 21: The `["entitlements"]` invalidation is currently wired only to uploads and hard deletes. A successful duplication and a move to the root also change the accounted storage, but their `onSuccess` handlers do not refresh this query. The gauge therefore remains stale after these actions. Could this helper also be called from `useMutationDuplicateItem` and from `useMoveItems` when ownership is affected?

src/frontend/apps/drive/src/features/layouts/components/explorer/ExplorerLayout.tsx
```diff
@@ -137,3 +144,192 @@ export const ExplorerPanelsLayout = ({
...
+
+type LeftPanelFooterProps = {
+  openSettingsModal: () => void;
+};
+
+const LeftPanelFooterMobile = (props: LeftPanelFooterProps) => {
+  return (
+    <div className="c__left-panel__footer__drive">
+      <UserProfile />
+      <Gaufre />
```
PanchoutNathan, line 181: For an anonymous user on tablet, `leftPanelContent` renders `LeftPanelMobile`, which already contains a `Gaufre`, while this mobile footer renders another one. This produces two app-grid triggers in the same panel. Could this `Gaufre` be conditional on an authenticated user, or could the one in `LeftPanelMobile` be removed?

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
PanchoutNathan, line 175: And what if it's a 401?

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
PanchoutNathan, line 439: ditto

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
PanchoutNathan, line 487: ditto

PanchoutNathan (approved):

src/backend/core/api/viewsets.py
```diff
@@ -2442,5 +2467,8 @@ def list(self, request):
                 method = getattr(entitlements_backend, method_name)
                 if callable(method):
                     entitlements[method_name] = method(request.user)
+        quota = entitlements_backend.get_quota(request.user)
+        if quota:
```
kernicPanel:
```suggestion
        if quota := entitlements_backend.get_quota(request.user)
```

src/backend/core/entitlements/backends/base.py
```diff
@@ -1,6 +1,38 @@
...
 
 from abc import ABC, abstractmethod
+from enum import StrEnum
+
+
+class QuotaState(StrEnum):
+    """State of a quota gauge returned by get_quota."""
+
+    DEFAULT = "default"
+    EXCEDEED_LOCKED = "excedeed_locked"
```
kernicPanel: This should be fixed in many places :/
```suggestion
    EXCEEDED_LOCKED = "exceeded_locked"
```

src/backend/core/signals.py
```diff
@@ -22,6 +26,20 @@ def file_post_save(sender, instance, **kwargs):  # pylint: disable=unused-argume
...
+def item_post_save_invalidate_storage_used(sender, instance, update_fields, **kwargs):  # pylint: disable=unused-argument
+    """
+    Invalidate the creator's cached storage usage when an item save may have
+    changed it. Bulk queryset updates bypass this signal and must invalidate
+    the cache explicitly.
+    """
+    if update_fields is not None and not STORAGE_USED_FIELDS.intersection(update_fields):
+        return
+    if instance.creator_id is None:
+        return
```
kernicPanel:
```suggestion
    if update_fields and STORAGE_USED_FIELDS.isdisjoint(update_fields):
        return
    if not instance.creator_id:
        return
```

kernicPanel (commented):
I'm not sure that we should use a cache for the quota, the added index should be fast enough for big files count. Also I don't like signals...
Did you benchmark reply times with and without cache ?
IMHO, we should keep thing as simple as possible.

src/frontend/apps/drive/src/features/drivers/implementations/StandardDriver.ts
NathanVss: Already guarded via abitilities

src/frontend/apps/drive/src/features/layouts/components/explorer/ExplorerLayout.tsx
NathanVss: Yeepp good catch !

kernicPanel (approved):

PR #766, opened 2026-07-07
🐛(frontend) refresh the Recent view after item mutations

maartendra:
## Bug

A mutation performed while viewing the **Recent** page does not refresh the list — a newly created / converted / renamed / deleted item only shows up after a manual page reload. Other views (My files, Shared, Favorites) refresh correctly.

## Cause

`useGetQueryKeyToRefresh` returns `["items", "infinite"]` (plus `["items", <parentId>, "children"]` when a parent is known). React Query invalidation is **prefix-based**, so `["items", "infinite"]` matches the general list views but **not** the Recent view, whose query key is `["items", "recent", "infinite"]` (`useInfiniteRecentItems`). As a result no mutation ever invalidates the Recent list.

## Fix

Also invalidate `["items", "recent", "infinite"]` in the refresh helper.

## Steps to reproduce

1. Open the **Recent** view.
2. Create a document, convert a legacy file, or delete/rename an item.
3. Observe the list does not update until a manual reload.

With this change, Recent refreshes like the other views.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(frontend) refresh the Recent view after item mutations
 - 🐛(frontend) keep uploaded items usable while malware analysis runs
 - 🐛(backend) stream export files from S3 without buffering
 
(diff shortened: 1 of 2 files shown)


NathanVss (approved):

NathanVss:
Thank you for contributing 🙏

PR #779, opened 2026-07-20
🐛(backend) delete malware detection record when purging an item

kernicPanel:
## Purpose

When an item is purged while its malware analysis is still pending or
processing, the detection record is left behind. The analysis keeps
being relaunched and crashes on the missing file: records stay stuck
in processing and hold analysis slots.

## Proposal

- [x] delete the malware detection record when purging an item
- [x] add an "item exists" column and filter to the malware detection
  admin, to help cleaning up orphan records already accumulated
- [x] show human readable item sizes in the admin

### DIFF
--- a/src/backend/core/admin.py
+++ b/src/backend/core/admin.py
@@ -5,10 +5,16 @@ from functools import partial
 from django.contrib import admin, messages
 from django.contrib.auth import admin as auth_admin
 from django.db import transaction
+from django.db.models import Exists, OuterRef, Subquery, UUIDField
+from django.db.models.fields.json import KeyTextTransform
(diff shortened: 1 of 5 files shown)


src/backend/core/admin.py
NathanVss, line 1: couldn't we make the tasks self destroy when they detect that the file is destroyed? To avoid 100% manual tasks?

NathanVss (approved):

src/backend/core/admin.py
kernicPanel: With the added code in this PR, It should not happen. If it still happens, I prefer to get a warning through pending tasks.

PR #784, opened 2026-07-22
✨ Batch share an item from an imported contacts file

NathanVss:
## Purpose

Sharing an item with a large group means inviting people one by one.
This lets administrators and owners import a CSV or XLSX contacts file
(email, role) from the share modal to batch share an item. The feature
is opt-in per environment through a new `ALLOW_SHARE_IMPORT_FILE`
setting.

https://github.com/user-attachments/assets/203ae5bf-df75-4c45-a69a-25abccb3f28e

## Proposal

- ✨ add the `ALLOW_SHARE_IMPORT_FILE` setting (default off), exposed by
  the config endpoint so the frontend shows or hides the import entry,
  and enforced server side by the new route
- ✨ add `POST /api/v1.0/items/<item_id>/batch-share/`: every row is
  validated before any database write (a user cannot grant a role higher
  than their own); emails are resolved server side — registered users
  get an access, unknown emails an invitation; already shared or invited
  contacts are skipped and reported in the response; notification emails
  behave like single sharing
- ♻️ extract the share helpers into `core/services/accesses.py`
- ✨ wire the ui-kit contacts import in the share modal: flag-gated menu
  entry, async import that closes the modal on success and renders the
  backend error inside the import modal on failure
- ✅ backend tests for the new route and config flag, plus e2e coverage
  (hidden entry, successful import, failed import)

## Notes

- Draft until the next `@gouvfr-lasuite/ui-kit` release ships the share
  import modal: `package.json` must be bumped from 0.26.0 before review
  (the branch was developed against a local ui-kit checkout).
- The e2e backend environment (`env.d/development/common.e2e`) enables
  the flag so the success scenario exercises the real endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to
 - ✨(backend) make the upload ACL configurable to support GCS based storages
 - ✨(frontend) show the messages widget button on the homepage
 - ✨(frontend) open the messages widget from the help menu
+- ✨(backend) add an item batch share endpoint gated by ALLOW_SHARE_IMPORT_FILE
 
 ### Changed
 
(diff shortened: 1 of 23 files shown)


kernicPanel (approved):
Don't forget src/frontend/apps/drive/package.json, but lgtm

PR #787, opened 2026-07-27
Move favorite API endpoint to `/items/favorites/`

jmaupetit:
## Purpose

Current favorite items listing endpoint path is `/items/favorite_list/`. To respect current API semantic (_e.g._ `/items/{id}/favorite/`), we think we can make a shorter version of the endpoint path without being less explicit 😜

## Proposal

- [x] 📝(backend) fix documented `trashbin` API endpoint
- [x] 🚚(global) move favorite items API endpoint to `/items/favorites/`


kernicPanel (commented):
IMHO, we should use `items/favorites`

jmaupetit:
> IMHO, we should use `items/favorites`

Sorry I missed this. This is now fixed.

kernicPanel (approved):

PR #788, opened 2026-07-27
✨(frontend) open the messages widget from the help menu

NathanVss:
## Purpose

Some instances use the messages widget for support instead of a plain mailto link and display the messages widget on homepage if configured.

### Connected
CleanShot 2026-07-27 at 16 35 21

### Anon
Capture d’écran 2026-07-27 à 17 14 56

- Add a `supportMessagesWidget` flag to the help menu config
- When enabled, the "Contact us" entry opens the messages widget, taking precedence over the support email
- Add messages widget on homepage for anon users
- Add e2e tests

### DIFF
--- a/src/frontend/apps/drive/src/features/feedback/useMessagesWidget.tsx
+++ b/src/frontend/apps/drive/src/features/feedback/useMessagesWidget.tsx
@@ -1,6 +1,7 @@
 import { useTranslation } from "react-i18next";
 import { useAuth } from "../auth/Auth";
 import { useConfig } from "../config/ConfigProvider";
+import { WidgetHelper } from "./widget-helper";
 
 /**
  * Hook that opens the feedback widget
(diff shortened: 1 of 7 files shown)


src/frontend/apps/drive/src/features/feedback/useMessagesWidget.tsx
```diff
@@ -14,59 +15,73 @@ export const useMessagesWidget = () => {
...
+      emailPlaceholder,
+      submitText,
+      successText,
+      successText2,
+    };
+  };
+
+  const canLoadWidget = () => {
+    return channel && apiUrl && widgetPath;
+  };
```
jbpenrath: Why it has to be a callback? 
Furthermore, you are using this method as deps of useEffect. Its naming let's think it is a boolean and nope. Furthermore as it a callback and it is not memoized, the func is recreated at each render, pointing to a new memory slot each time so I feel related useEffect will be called more than we expect.

jbpenrath (approved):
✅ LGTM

PR #792, opened 2026-07-30
✨(backend) exclude external service items from storage quota

NathanVss:
## Purpose

Allow external services consuming the resource server API to create items that do not
count against their creator's storage quota, and provide an operational path to enable
quotas on existing instances without blocking heavy users.

## Proposal

- **`quota_excluded` flag on `Item`**: items flagged are ignored by
  `CreatorStorageComputeBackend` and the flag participates in the storage-used cache
  invalidation. The covering index is replaced by `item_creator_size_quota_idx`, whose
  condition also filters out excluded rows so the aggregation stays index-only.
- **`EXTERNAL_API_AUD_ITEM_ATTRIBUTES` setting**: maps a resource server token audience
  to attributes applied to items at creation, e.g.
  `{"some_audience": {"quota_excluded": true}}`. Applied on both the root create and
  `children` creation paths of the external API only; the regular API is unaffected.
  Documented in `docs/resource_server.md`.
- **`grant_unlimited_storage` command**: sets `storage_limit_override=0` (unlimited) for
  users above a threshold (`--threshold-gb`, decimal GB), skipping users with an existing
  override. Supports `--dry-run`. The aggregation uses the same filter as the storage
  compute backend so it agrees with quota enforcement.
- Migration `0028` swaps the covering index with `CREATE`/`DROP INDEX CONCURRENTLY`
  (non-atomic): no write locks, but if it fails mid-run it can leave an INVALID index to
  drop manually before retrying.

### DIFF
--- a/src/backend/core/management/commands/grant_unlimited_storage.py
+++ b/src/backend/core/management/commands/grant_unlimited_storage.py
@@ -1,0 +1,64 @@
+"""Grant unlimited storage to users already using more than a threshold."""
+
+from django.core.management.base import BaseCommand
+from django.db.models import Q, Sum
+
+from core.models import User
+
+
+class Command(BaseCommand):
+    """
+    Set storage_limit_override to 0 (unlimited) for every user whose quota
+    counted storage exceeds the given threshold. Users with an existing
+    override are left untouched so explicit per-user decisions are never
+    clobbered.
+    """
+
+    help = "Set storage_limit_override=0 (unlimited) for users above a storage threshold"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--threshold-gb",
+            type=float,
+            required=True,
+            help="Storage threshold in GB, users strictly above it are granted unlimited storage",
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Only report the users that would be updated",
+        )
+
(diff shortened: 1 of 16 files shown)


src/backend/drive/settings.py
```diff
@@ -1296,6 +1296,15 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    # Extra attributes applied to items created through the external API,
+    # keyed by the token audience of the request,
+    # e.g. {"some_audience": {"quota_excluded": True}}
```
kernicPanel, line 1301: IMHO, we should update docs/env.md with this docstring.

src/backend/core/management/commands/grant_unlimited_storage.py
```diff
@@ -0,0 +1,64 @@
...
+                    ),
+                    default=0,
+                )
+            )
+            .filter(total_size__gt=threshold)
+        )
+
+        if options["dry_run"]:
+            for email in users.values_list("email", flat=True):
+                self.stdout.write(f"[dry-run] Would grant unlimited storage to {email}")
```
kernicPanel, line 54: I don't know if logging emails is useful, and rgpd friendly.
Maybe replace this with their id ?

kernicPanel (approved):

PR #794, opened 2026-08-03
✨(backend) make forward auth request uri header configurable

erincandescent:
## Purpose

In our deployment we're using Traefik, not nginx, as an ingress. Traefik uses `X-Forwarded-Url` instead of `X-Original-Url`. This adds a setting which lets users adapt Docs to their ingress proxy of choice

Port of https://github.com/suitenumerique/docs/pull/2241 (written by me) from Docs

## Proposal

Add a `MEDIA_AUTH_ORIGINAL_URL_HEADER` which lets you configure it, with `HTTP_X_ORIGINAL_URL` as default value

## AI usage

Porting from the Docs codebase to Drive was done by DeepSeek 4 Flash. The original code was written by me and @lunika , and it appears to have reproduced it faithfully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- ✨(backend) make the forward auth request uri header configurable
 - ✨(backend) make the upload ACL configurable to support GCS based storages
 - ✨(frontend) show the messages widget button on the homepage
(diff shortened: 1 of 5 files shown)


erincandescent:
I see there's a similar PR #743 but (a) honestly I don't see a good reason to allow multiple headers (in fact it seems like a security disaster) and (b) I don't see a reason to diverge from docs

PR #416, opened 2025-11-24
Implement wopi proof keys mechanism

lunika:
## Purpose

The wopi specs provide a documentation explaining how to verify that a wopi request provides from a trusted wopi client.

This documentation is available here https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/scenarios/proofkeys

We have first to save wopi proof keys if present in the discovery xml than use them to validate the signature provided in the request.

## Proposal

- [x] ✨(wopi) save woopi client proof key in the configuration
- [x] ✨(wopi) add utils to validate a WOPI request signature
- [x] ✨(wopi) verify wopi signature in Wopi viewset

PR #795, opened 2026-08-03
🔒️(backend) fix two WOPI security bypasses

kernicPanel:
## Purpose

Two WOPI endpoints let a standard user store a file that neither the
allowed extension list nor the antivirus ever checked.

`RENAME_FILE` used the requested name as decoded, without validating the
resulting extension. `PutFile` wrote content straight to object storage
without triggering malware detection, which only ran on the upload
endpoint.

## Proposal

- [x] Reject a rename whose target contains a path separator or lands on a
      disallowed extension, and compare extensions case insensitively
- [x] Run malware detection on the content written by PutFile

### DIFF
--- a/src/backend/wopi/viewsets.py
+++ b/src/backend/wopi/viewsets.py
@@ -35,6 +35,8 @@ X_WOPI_ITEMVERSION = "X-WOPI-ItemVersion"
 X_WOPI_LOCK = "X-WOPI-Lock"
 S3_VERSION_ID = "VersionId"
 
+ILLEGAL_FILENAME_CHARS = ("/", "\\")
+
 
 class WopiViewSet(viewsets.ViewSet):
     """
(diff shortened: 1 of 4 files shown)


src/backend/wopi/viewsets.py
```diff
@@ -332,25 +332,40 @@ def _rename_file(self, request, pk=None):
             return Response(status=401)
 
         new_filename = request.META.get("HTTP_X_WOPI_REQUESTEDNAME")
-
         if not new_filename:
+            invalid_filename_error = "No filename provided"
+        else:
+            # Convert it to utf-7 to avoid issues with special characters
+            new_filename = new_filename.encode("ascii").decode("utf-7")
```
NathanVss, line 346: Why utf-7, why not utf-8 as utf-7 is considered obsolete?

src/backend/wopi/viewsets.py
```diff
@@ -332,25 +332,40 @@ def _rename_file(self, request, pk=None):
...
+            new_filename_with_extension = f"{new_filename}{splitext(item.filename)[1]}"
+            _, target_extension = splitext(new_filename_with_extension)
+
+            invalid_filename_error = None
+            if "/" in new_filename or "\\" in new_filename:
```
NathanVss: WDYT about defining a list of ILLEGAL_CHARS instead?

src/backend/wopi/viewsets.py
```diff
@@ -185,8 +186,12 @@ def _put_file_content(self, request, pk=None):
         s3_client = default_storage.connection.meta.client
         default_storage.save(item.file_key, file)
         item.size = file.size
+        # Keep the item READY during re-analysis: non-creators cannot open
+        # non-READY files in WOPI.
         item.save(update_fields=["size", "updated_at"])
 
+        malware_detection.analyse_file(item.file_key, item_id=item.id)
```
NathanVss, line 195: This approach leave a window during which a dangerous file is in READY state. 

That being said, to avoid degrading live edition I understand we have no better choice for now. 
The fine grade solution is indeed only allowing wopi editors to access the wopi API via this PR: https://github.com/suitenumerique/drive/pull/416

NathanVss (approved):

src/backend/wopi/viewsets.py
kernicPanel: utf-7 is part of wopi specs…

lunika (approved):

PR #807, opened 2026-08-21
🐛(backend) fix malware re-analysis of a file already scanned

kernicPanel:
## Purpose

Saving a document through WOPI PutFile returns a 500 when the file
already has a malware detection record (pending, processing or failed):
the JCOP backend creates one record per file path and fails on the
unique constraint. The admin "trigger file analysis" action has the
same issue.

## Proposal

- [x] add a `reanalyse_file` helper that drops the previous detection
  record, when the backend supports it, before scheduling the analysis
- [x] use it in WOPI PutFile and in the admin reanalysis action

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- 🐛(backend) fix malware re-analysis of a file already scanned
+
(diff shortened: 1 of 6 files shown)


lunika (approved):

kernicPanel:
Superseded: the unique constraint was removed in django-lasuite 0.0.28, drive will bump the dependency instead.

Issue #285, opened 2025-07-25
Share a file with an ephemeral link


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

As a user I want to share an item with someone else not having a drive account.

**Describe the solution you'd like**

For this I want to have a feature allowing me to share a link, I can configure this link with these parameters:
- End validity date
- One time usage

Possible other parameters open to discussion:
 
- Password to access item

robin-lecomte:
Here’s my proposal for the advanced link settings (with the help of @PanchoutNathan)

[FIGMA HERE](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/DRIVE?node-id=11016-14581&t=NSgvHrrObiJFO9sK-1)

Summary of the proposal:
- We’re adding “connected” access, in addition to “private” and “public”
- We’re introducing an advanced settings button for the link (previously, there was only the role—reader, editor, etc.)
- A new advanced settings modal will include: role, expiration, password, and the option to create a one-time link

The idea of creating and managing multiple links was explored extensively but ultimately rejected due to significant complexity. The current proposal has the advantage of requiring only one additional modal and maintaining the current link-sharing approach that users are becoming quite familiar with, especially in docs.

<img width="3584" height="1604" alt="Image" src="https://github.com/user-attachments/assets/ded003e0-f459-4568-aa06-f43a0274dfe6" />

robin-lecomte:
[Here is an updated proposal](https://www.figma.com/design/XDjWHDLIiNY4mtpyj21EVK/Drive?node-id=11016-14581&t=JYIUipgSKTcnkPlZ-1)

I’ve reworked the link-sharing section of LaSuite’s sharing modal to introduce new options while maintaining the right level of information for the user, without adding too many steps that would hurt overall usability.

What’s new:  
- Two sections instead of one: a button for the link reach, and a button for link settings that allows changing the role and expiration date  
- Clearer, more explicit button labels to make actions immediately understandable at a glance  
- Unlike my previous proposal, the extra modal step is gone : no more opening a modal just to view or edit information, which improves overall ergonomics  
- The addition of more refined dropdown menus

<img width="3808" height="1522" alt="Image" src="https://github.com/user-attachments/assets/a84830a6-da79-413b-8c14-a356005b6daf" />

robin-lecomte:
[...]

PR #810, opened 2026-08-21
✨(backend) add expiration date and password to item share links

kernicPanel:
## Purpose

Let users share an item through a link that expires at a given date
and/or requires a password. Backend part of #285.

## Proposal

- [x] add `link_expires_at` and `link_password` fields on Item
- [x] deny link access once the link has expired, in abilities, inherited
      link definitions, tree and breadcrumb
- [x] deny link access until the password has been unlocked, the
      password of the closest ancestor link applies to descendants
- [x] `POST /items/{id}/unlock/` remembers the unlock, in session for
      anonymous users and on the link trace for authenticated users,
      rate limited by `API_ITEMS_UNLOCK_THROTTLE_RATE`
- [x] carry unlocked links in the WOPI access token
- [x] `link-configuration` accepts `link_expires_at` and `link_password`
- [x] expose `link_expires_at`, `has_link_password` and the
      `password_locked` ability
- [x] filter indexed search results on expired or locked links
- [x] unlocked items stay searchable while their link is live


kernicPanel:
@coderabbit review

kernicPanel:
@coderabbitai full review

src/backend/core/tests/items/test_api_items_link_configuration.py
```diff
@@ -538,3 +542,141 @@ def test_api_items_link_configuration_sync_link_reach_descendants():
...
+    client = APIClient()
+    client.force_login(user)
+
+    item = factories.ItemFactory(users=[(user, models.RoleChoices.OWNER)])
+    expires_at = timezone.now() + timedelta(days=1)
+
+    response = client.put(
+        f"/api/v1.0/items/{item.id!s}/link-configuration/",
+        {
+            "link_reach": models.LinkReachChoices.PUBLIC,
```
lunika, line 559: can you parametrize this test adding the `connected` link_reach ? It is allowed to set an expiration date on a connected link_reach I think

src/backend/core/models.py
```diff
@@ -1310,6 +1311,19 @@ def link_definition(self):
             return {"link_reach": LinkReachChoices.RESTRICTED, "link_role": None}
         return {"link_reach": self.link_reach, "link_role": self.link_role}
 
+    @property
+    def link_unlock_key(self):
+        """Return the key remembering an unlock of this item link, or None without password."""
+        return link_unlock_key(self.id, self.link_password) if self.link_password else None
```
lunika, line 1317: This function is used in this commit `✨(backend) hash and check item link passwords` but is not yet existing

src/backend/core/api/viewsets.py
```diff
@@ -1637,6 +1692,34 @@ def batch_share(self, request, *args, **kwargs):
...
+                user=request.user,
+                item=password_item,
+                defaults={"link_unlocked_at": timezone.now()},
+            )
+        else:
+            request.session[UNLOCKED_LINK_ITEMS_SESSION_KEY] = list(
+                request.user.unlocked_link_items
+            )
+
+        return drf.response.Response(self.get_serializer(item).data, status=drf.status.HTTP_200_OK)
```
lunika, line 1721: Is it needed to serialize the item or the front will redirect the user to the item and then will fetch again the item ?

src/backend/core/permissions/backends/role.py
```diff
@@ -172,6 +196,7 @@ def as_dict(self) -> dict[str, bool | dict[str, list[str]]]:
             "invite_owner": self.can_invite_owner(),
             "link_select_options": self.link_select_options(),
             "move": self.can_move(),
+            "password_locked": self.password_locked(),
```
lunika, line 199: Why not naming the ability `unlock` like the action in the viewset ?

src/backend/core/api/viewsets.py
```diff
@@ -1464,24 +1503,38 @@ def search(self, request, *args, **kwargs):
...
+        root_paths = utils.filter_root_paths(sorted(top_level_paths, key=str), skip_sorting=True)
 
         path_list = db.Q()
         for top_level_item in root_paths:
             path_list |= db.Q(path__descendants=top_level_item)
 
         queryset = queryset.filter(path_list)
+        # Link passwords are the only restriction the scope cannot express, keep the
+        # database pagination when the scope holds none
+        scope_has_link_password = queryset.filter(link_password__isnull=False).exists()
```
magopian, line 1537: This is only used in the "non indexed search" case, right? So maybe move it down so it's not running an extra query when not needed?

src/backend/core/models.py
```diff
@@ -1295,9 +1298,16 @@ def compute_ancestors_links_paths_mapping(self):
...
+    def is_link_expired(self):
+        """Return whether the link expiration date is past."""
+        return self.link_expires_at is not None and self.link_expires_at <= timezone.now()
+
     @property
     def link_definition(self):
-        """Returns link reach/role as a definition in dictionary format."""
+        """Return link reach/role as a definition in dictionary format."""
+        if self.is_link_expired:
+            return {"link_reach": LinkReachChoices.RESTRICTED, "link_role": None}
```
magopian, line 1310: Would it make sense to have a code comment here explaining that by "translating" the fact that the link is expired to `RESTRICTED`, everything else that already exist will just work as expected?

src/backend/core/tests/test_models_items.py
```diff
@@ -1479,3 +1479,47 @@ def test_models_items_restore_complex_bis():
...
+        link_role="reader",
+        link_expires_at=timezone.now() + timedelta(minutes=1),
+    )
+
+    assert item.computed_link_reach == LinkReachChoices.PUBLIC
+    assert item.get_abilities(AnonymousUser())["retrieve"] is True
+
+
+def test_models_items_link_expired_ancestor_not_inherited():
+    """An expired link on an ancestor should not be inherited by its descendants."""
```
magopian, line 1515: I had a hard time understanding the test, and I'm wondering if it's because of the naming and its docstring?

I now understand that it's testing that the **reach** is not inherited when the link is expired, but the fact that the link is expired is indeed blocking the child from being accessed (so the "inaccessibility of the item" is inherited).

How about renaming the test something like `test_models_items_link_expired_ancestor_blocks_descendants`, and/or changing the docstring to something like `"""An expired link REACH on an ancestor should not be inherited by its descendants."""`

src/backend/core/permissions/backends/role.py
```diff
@@ -26,11 +26,30 @@ def __init__(self, user: models.User | AnonymousUser, item: models.Item) -> None
         # The access role is based on accesses only, before any link boost
         self.access_role = item.get_role(user)
         self.is_deleted = bool(item.ancestors_deleted_at)
-        link_definition = item.computed_link_definition
+        link_reachable, link_locked = self._link_status()
+        if link_reachable and not link_locked:
+            # The highest of the access role and the link role, needed for a user
+            # with an access lower than the link role and for a user without access
```
magopian, line 32: Is this comment equivalent to the following? I had a hard time parsing it (maybe i'm just tired 😬 ):
```python
# The link boosts access when its role exceeds the user's explicit access, and never reduces it.
```

magopian (approved):
This was very nice to review, thanks to the multiple atomic commits, thanks and congrats, huge work! Very clean and thorough.

I've commented a couple of nits, up to you to decide if they need fixing, I didn't see anything blocking.

Well done!

PR #814, opened 2026-08-28
🐛(backend) block item delete by creator without access

NathanVss:
## Purpose

Fix a broken access control issue in item deletion.

A user could delete an item they had created inside a folder shared with them, even after the owner had revoked their access. Because the delete permission relied on the `creator` field alone (an immutable historical reference), revoking the share left the delete ability intact. The user could then call `DELETE /api/v1.0/items/<id>/` on a subfolder they once created and soft-delete the whole subtree, including files and folders added by the owner.

## Changes

- Gate the creator branch in `Item.get_abilities` on the user still holding editor access on the item. The ability is now lost as soon as access is revoked, consistent with the update permission.
- Owner/admin deletion of content created by other users is unchanged.
- Add API and model tests covering a creator whose access was revoked (denied) and a creator who keeps editor access (allowed).


kernicPanel:
I think we should wait for https://github.com/suitenumerique/drive/pull/756 , or rebase this PR in top of it.

kernicPanel (approved):

PR #815, opened 2026-08-28
Make drive interoperable for local development

jmaupetit:
## Purpose

We need a way to run every LaSuite services locally to work on services inter-operabilities, without hostname and port conflicts, nor duplicate services.

## Proposal

- [ ] update drive services configuration


compose.yaml
```diff
@@ -16,34 +22,34 @@ services:
...
+  #     - "9001:9001"
+  #   healthcheck:
+  #     test: ["CMD", "mc", "ready", "local"]
+  #     interval: 1s
+  #     timeout: 20s
+  #     retries: 300
+  #   entrypoint: ""
+  #   command: minio server --console-address :9001 /data
+  #   volumes
+  #     - ./data/media:/data
```
jmaupetit: fixme

PR #818, opened 2026-09-02
✨(backend) add file-scanner malware webhook endpoint

NathanVss:
## Purpose

Wire Drive to the `FileScannerBackend` landing in django-lasuite (suitenumerique/django-lasuite#79), so malware scanning can run on [suitenumerique/file-scanner](https://github.com/suitenumerique/file-scanner) instead of JCOP. The backend is selected purely through the existing `MALWARE_DETECTION_BACKEND` / `MALWARE_DETECTION_PARAMETERS` env vars — no settings code change.

file-scanner pushes its verdict: it downloads the file from a presigned URL and POSTs the report to a webhook. Drive only has to expose that webhook; verdict mapping, secret verification and callback dispatch stay in the library, so `malware_detection_callback` and the upload states are untouched.

## Proposal

- New route `POST /api/v1.0/webhooks/malware/<uuid:record_id>/` handled by `MalwareScanWebhookView`, which forwards the payload to the backend's `handle_webhook_result()`. It answers 404 when the active backend has no webhook flow (dummy, JCOP), and always 2xx otherwise — the scanner treats redirects as failed deliveries.
- New `MALWARE_DETECTION_RESCHEDULE_PROCESSING_ANALYSIS_DAYS` setting (float): with file-scanner a lost webhook is only recovered by the `reschedule_processing_analysis` cronjob, so deployments need a much lower delay than the 3-day default (e.g. `0.25`).
- Tests cover clean/malware/file-too-large verdicts, wrong secret, unknown record, duplicate delivery and malformed JSON. They import the backend directly, so `test-back` stays red until the django-lasuite pin is bumped to the release containing it.
- New `docs/malware_detection.md` page (upload states, env vars, file-scanner setup, maintenance cronjobs) and the missing `MALWARE_DETECTION_*` entries in `docs/env.md`.

Depends on suitenumerique/django-lasuite#79. Once it is released, the follow-up here is a one-line `django-lasuite` bump in `pyproject.toml` (plus lockfile), which will bring `malware_detection` migrations 0003–0005 and un-skip the tests.

Deployment prerequisites: register Drive's Ed25519 public key in the scanner's `JWT_ISSUER_KEYS`; allow the S3 endpoint host and Drive's internal origin in the scanner's SSRF config; presigned URL expiry (`AWS_QUERYSTRING_EXPIRE`, 3600s) must exceed the scanner's queue latency.

Verified end to end against a local file-scanner: clean upload goes ANALYZING → READY, EICAR goes SUSPICIOUS with the scanner breakdown stored in `malware_detection_info`, WOPI re-saves don't stack duplicate scans.


CHANGELOG.md
```diff
@@ -8,27 +8,15 @@ and this project adheres to
...
-### Fixed
-
-- 🐛(frontend) refresh the Recent view after item mutations
-
-## [v0.22.0] - 2026-09-09
-
 ### Added
 
+- ✨(backend) add file-scanner malware webhook endpoint
+- 🔧(backend) make malware reschedule delay configurable
```
kernicPanel, line 14: Should be added in the next commit.

kernicPanel (commented):
LGTM, but let's wait for the django-lasuite PR to be merged.

PR #820, opened 2026-09-04
Improve psql wrapper script

jmaupetit:
## Purpose

The `bin/psql` wrapper script can be used to run the `psql` command using a TTY or not depending on whether we need the REPL or to execute a SQL script. Also, getting environment variable from their definition file is not necessary as they are already defined in the running container.

## Proposal

- [x] 🔨(project) improve psql script
- [x] 🔨(project) remove unused DB_HOST and DB_PORT env vars


bin/psql
```diff
@@ -0,0 +1,26 @@
...
+declare cmd="psql -U \${POSTGRES_USER} -d \${POSTGRES_DB} -v ON_ERROR_STOP=1 $*"
+declare -a opts
+declare -i TTY=${TTY:-0}
+
+if [[ ${TTY} -eq 0 ]]; then 
+  opts+=("-T")
+fi
+
+# Run psql command using the postgres user, not host user
+_docker_compose exec "${opts[@]}" drive-postgresql sh -c "${cmd}"
```
jmaupetit: If we want to avoid depending on #815 

```suggestion
_docker_compose exec "${opts[@]}" postgresql sh -c "${cmd}"
```

Makefile
```diff
@@ -26,12 +26,7 @@
 BOLD := \033[1m
 RESET := \033[0m
 GREEN := \033[1;32m
-
-
-# -- Database
-
-DB_HOST                 = postgresql
-DB_PORT                 = 5432
+SHELL := /bin/env bash
```
NathanVss: On MacOS it breaks: 


```
make clear-db-e2e
./bin/psql < bin/clear_db_e2e.sql
make: /bin/env: No such file or directory
make: *** [clear-db-e2e] Error 1
```

Makefile
```diff
@@ -48,14 +43,14 @@ COMPOSE_RUN_APP_NO_DEPS = $(COMPOSE_RUN) --no-deps app-dev
 COMPOSE_RUN_CROWDIN     = $(COMPOSE_RUN) crowdin crowdin
 
 # -- Backend
-MANAGE              	= $(COMPOSE_RUN_APP) python manage.py
-MANAGE_EXEC         	= $(COMPOSE_EXEC_APP) python manage.py
-MAIL_YARN           	= $(COMPOSE_RUN) -w /app/src/mail node yarn
-PSQL_E2E 				= ./bin/postgres_e2e
+MANAGE                  = $(COMPOSE_RUN_APP) python manage.py
+MANAGE_EXEC             = $(COMPOSE_EXEC_APP) python manage.py
+MAIL_YARN               = $(COMPOSE_RUN) -w /app/src/mail drive-node yarn
```
NathanVss: `drive-node` is not a compose service 🤔

Makefile
```diff
@@ -140,7 +135,7 @@ bootstrap-e2e: \
 .PHONY: bootstrap-e2e
 
 clear-db-e2e: ## quickly clears the database for e2e tests, used in the e2e tests
```
NathanVss, line 137: With this new approach it clears whatever postgres instance is running, the previous one was only clearing the e2e one even if the regular dev one was running

Makefile
NathanVss: Just removing this line should work fine?

jmaupetit: Should be fixed now using `/usr/bin/env bash` instead!

Makefile
jmaupetit: Sorry, it has been cherry-picked from #815. This is now fixed!

Makefile
jmaupetit: Good catch! I've added support to enforce target database definition.

jmaupetit: The command will fail if the environment database name differs from the running database.

Makefile
```diff
@@ -140,7 +135,7 @@ bootstrap-e2e: \
...
-	$(PSQL_E2E) -c "$$(cat bin/clear_db_e2e.sql)"
+	POSTGRES_DB=drive_e2e $(PSQL) < bin/clear_db_e2e.sql
```
jmaupetit: Also: as the SQL script is generic, maybe it should be renamed as follow:

```suggestion
	POSTGRES_DB=drive_e2e $(PSQL) < bin/clear_records.sql
```

@NathanVss WDYT?

NathanVss: Agreed

NathanVss (approved):
Thank you for your contrib :) !!

PR #826, opened 2026-09-09
✨ allow to leave an item

### LINKED ISSUE #515: Allow users to "leave" a folder/file
## Feature Request

Some users are asking for a leave feature. They cannot leave a folder or item if someone invited them.

magopian:
Fixes #515

## Purpose

We want to allow users to leave an item where they have an access or they have visited creating a link_trace. All sub items should also be left at the same time.

## Proposal

This PR will mimic what's been done on the [docs](https://github.com/suitenumerique/docs) project in the following PRs:
- backend: https://github.com/suitenumerique/docs/pull/2365
- frontend: https://github.com/suitenumerique/docs/pull/2410

To know if the user can leave an item we have to check when computing the abilities if a record is existing in the LinkTrace table. This is a N+1 query situation. To avoid it, we added an annotation in the ItemQueryset like we already do to annotate the user role. There is one edge case where the annotation is made to soon, it is when the user is visiting an item for the first time, the `get_object` adds the annotation and in the permission, we compute the abilities. The `leave` property is False because the entry in the LinkTrace table is not made, when the serializer asks for the abilities again, it is still False.
So in the `retrieve` method in the viewset we force the `user_has_link_trace` to the correct value.

- [x] backend implementation
- [x] frontend implementation


kernicPanel:
Hello @magopian 
Can you please rebase on main ?

magopian:
> Hello @magopian Can you please rebase on main ?

Done!

kernicPanel (approved):
I reviewed the backend part which looks good to me.

PR #827, opened 2026-09-09
🐛(backend) gate item creation at the root on the upload entitlement

NathanVss:
`can_upload` was only enforced on `POST /items/{id}/children/`. Creating a file through `POST /items/` bypassed it, and the "+ New" button takes that path whenever the current folder is not writable — including on "My files".

Template based creation stores the file and marks the item READY itself, so `upload-ended`, the only other gate, never runs. An over quota user could create documents indefinitely from the root, while the same action inside a folder returned 403.

The check now lives in a helper shared by both create paths. The frontend also disables the document entries and shows the reason, instead of letting the user name a document only to have it refused.


src/backend/core/tests/items/test_api_items_children_from_template.py
kernicPanel, line 1: IMO, the tests docstrings should be clearer about their intent.

kernicPanel (approved):

PR #836, opened 2026-09-16
✨(backend) create items on behalf of another user via external API

kernicPanel:
## Purpose

Services integrating Drive through the external API can only create items
in the space of the user authenticated by the token. They need to drop
files into a given user's space, so that user owns the item and its size
counts in their storage quota.

## Proposal

Add an optional `owner_email` field on item creation through the external
API. The matching user becomes the creator and owner of the item, the
token user keeps an owner access to complete the upload.

- [x] `get_create_owner()` hook on `ItemViewSet`
- [x] `owner_email` on `POST /external_api/v1.0/items/`, with storage
  quota check on the target user
- [x] `EXTERNAL_API_AUD_CREATE_ON_BEHALF` setting listing the token
  audiences allowed to use it, others get a 403
- [x] unknown email: the item is created without creator and an owner
  invitation is sent, the existing first-login conversion assigns it
- [x] documentation and changelog


src/backend/core/external_api/viewsets.py
```diff
@@ -50,6 +54,74 @@ def get_create_extra_attributes(self):
...
+    owner_email = None
+
+    def get_create_owner(self):
+        """Resolve the user given by owner_email, or the token user when absent."""
+        owner_email = self.request.data.get("owner_email")
+        if not owner_email:
+            return self.request.user
+
+        audience = getattr(self.request, "resource_server_token_audience", None)
+        if audience not in settings.EXTERNAL_API_AUD_CREATE_ON_BEHALF:
```
jmaupetit, line 66: Maybe we should also log this for security/audit purpose?

src/backend/core/external_api/viewsets.py
```diff
@@ -50,6 +54,74 @@ def get_create_extra_attributes(self):
...
+                issuer=self.request.user,
+            )
+            transaction.on_commit(
+                lambda: item.send_invitation_email(
+                    invitation.email,
+                    invitation.role,
+                    self.request.user,
+                    self.request.user.language or settings.LANGUAGE_CODE,
+                )
+            )
```
jmaupetit, line 123: How do we deal with multiple invites for the same user?

If used in the migrator context, a user with no drive account will receive an invitation per-file!

jmaupetit (commented):
Awesome!

PR #839, opened 2026-09-16
✨(backend) reserve upload size against the storage quota

NathanVss:
## Purpose

Storage quota accounting is approximate. The entitlement decision is taken
before the size of an upload is known, and a pending upload is not counted
until it is finalized. Item creation is not rate limited either.

## Proposal

- [x] Require a `size` when creating a file item and reserve it on creation
- [x] Take the entitlement decision in the same transaction, rolling back if refused
- [x] Bind the upload authorization and the finalization check to that size
- [x] Rate limit the item creation endpoints, with configurable rates
- [x] Adapt the frontend, the load test scenarios and the resource server example


kernicPanel (approved):
The fist commit should be split, but everything seems ok. Nice work 👍

